PR #8, opened 2020-06-25
Add docker support and typescript

### LINKED ISSUE #2: Setup docker compose
Set up Docker compose file for running app in a node container

cailyncodes:
Should close #2


src/utils/is-development.ts
```diff
@@ -0,0 +1,4 @@
+export function isDevelopment() {
+	const environment = process.env.NODE_ENV;
+	return !!environment || environment === 'development';
```
jlhogan, line 3: not sure this conditional is correct. When `environment` is not falsy, then right side of conditional is true, and left side does not get evaluated. When `environment` is falsy, then the left side gets evaluated and is always false since `falsy === 'development'` is never true.
was the intention to have `isDevelopment` = true when `environment` is falsy?

cailyncodes: Oh good check. Yes this should be `!environment || environment === 'development'`

PR #11, opened 2020-06-26
Add identity server integration

### LINKED ISSUE #3: Integrate WingedKeys/IdentityServer
Allow for authenticated sign in with WingedKeys

cailyncodes:
Should close #3


client/src/contexts/AuthenticationContext/AuthenticationContext.tsx
```diff
@@ -0,0 +1,313 @@
...
+		fetchServiceConfiguration();
+	}, [openIdConnectUrl]);
+
+	// Make an authorization request whenever:
+	// 1) the authorizationHandler is set,
+	// 2) the configuration is set or changes, and
+	// 3) it is the login screen (isLogin or isCallback)
+	useEffect(() => {
+		if (configuration && authorizationHandler) {
+			if (isLogin) {
```
jlhogan, line 177: can this be cleaned up at all by making more 'pure functions' as we discussed in some other ticket in the other repo somewhere?

docker-compose.yaml
```diff
@@ -22,11 +22,34 @@ services:
...
+
+  winged-keys-db:
+    image: mcr.microsoft.com/mssql/server
+    environment:
+      SA_PASSWORD: TestPassword1
+      ACCEPT_EULA: Y
+  winged-keys:
+    build:
+        context: .
+        dockerfile: ./winged-keys/Dockerfile
```
jlhogan, line 42: can we add checkout of winged keys repo requirement to README?

client/src/contexts/AuthenticationContext/AuthenticationContext.tsx
cailyncodes: This is the same code from `Hedwig`. Open to discussions on purifying things but don't think this PR is the one for that.

jlhogan: my thinking is always that the most likely time a non essential change is gonna get made is _now_ hehe but that's fine

PR #16, opened 2020-06-29
Add file upload route

### LINKED ISSUE #5: Allow for saving of file form uploads to disk
Add a route handler for accepting the uploaded file and write it to disk

cailyncodes:
Should close #5

### DIFF
--- a/src/middleware/authenticate.ts
+++ b/src/middleware/authenticate.ts
@@ -18,7 +18,7 @@ const handleError = (
 	next: NextFunction
 ) => {
 	if (err.name === 'UnauthorizedError') {
-    res.status(401).json({
+		res.status(401).json({
 			status: 401,
 			message: 'Unauthorized'
 		});
--- a/package.json
+++ b/package.json
@@ -7,11 +7,15 @@
 	"dependencies": {
 		"@types/express": "^4.17.6",
 		"@types/http-proxy": "^1.17.4",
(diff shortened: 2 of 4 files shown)


src/data/users.ts
```diff
@@ -0,0 +1,14 @@
+import { User } from "../models/user";
+
+export const users: User[] = [
+	{
+		id: 1,
+		wingedKeysId: "2c0ec653-8829-4aa1-82ba-37c8832bbb88",
+		firstName: "Voldy",
```
mmazanec22, line 7: Lol, yes

src/middleware/authenticate.ts
```diff
@@ -0,0 +1,48 @@
...
+	requestProperty: 'token'
+});
+
+const handleError = (
+	err: Error,
+	_: Request,
+	res: Response,
+	next: NextFunction
+) => {
+	if (err.name === 'UnauthorizedError') {
```
jlhogan, line 20: nit: weird whitespace

src/types/custom.d.ts
```diff
@@ -0,0 +1,7 @@
+declare namespace Express {
```
jlhogan, line 1: why is this a custom type?

cailyncodes: B/c of Typescript, we need to define that we are adding `token` and `sub` to the request properties

Issue #17, opened 2020-07-01
As a provider, I can log in with my username and password.


julyytran:
[Mockup](https://app.abstract.com/share/5ea2033d-b4eb-4322-8a2a-1b601b77c8fc?collectionId=e2036f17-be42-40a0-9a0b-aa94c3efbc51&sha=721beadef8c072936fe5c4358852ea21dd9f32b5)

I can log in to staging using the username `voldemort` and password `thechosenone`. 
If I enter an invalid username or password, I seen an error that I have "Invalid username or password."

When I am logged in, I see "Hi, [firstName]" in the header in the top right hand corner.

I can log out by clicking "Log out" in the header's upper right hand corner.

When I log out, I am taken back to the sign in page.

cailyncodes:
Closed by #22

PR #22, opened 2020-07-02
Add form upload component and logout

### LINKED ISSUE #4: Create file upload form
Blocked by #3 

Create the frontend components necessary for a form upload

cailyncodes:
Should close #4 and #17

### DIFF
--- a/client/src/containers/Header/Header.tsx
+++ b/client/src/containers/Header/Header.tsx
@@ -1,0 +1,23 @@
+import React, { useContext } from 'react';
+import { Link } from 'react-router-dom';
+import UserContext from '../../contexts/UserContext/UserContext';
+
+const Header: React.FC = () => {
+	const { loading, user } = useContext(UserContext);
+
+	if (loading) {
+		return <></>;
+	}
+
+	return (
+		<header>
+			<h1>OEC Data Collection</h1>
+			<nav>
+				{user ? <span>Hi, {user.firstName}</span> : <Link to="/login">Sign in</Link>}
+				<p>{user && <Link to="/logout">Logout</Link>}</p>
+			</nav>
+		</header>
+	);
+}
+
+export default Header;
--- a/client/src/containers/Header/Header.tsx
+++ b/client/src/containers/Header/Header.tsx
@@ -11,7 +11,7 @@ const Header: React.FC = () => {
 
 	return (
 		<header>
-			<h1>OEC Data Collection</h1>
+			<p>OEC Data Collection</p>
 			<nav>
 				{user ? <span>Hi, {user.firstName}</span> : <Link to="/login">Sign in</Link>}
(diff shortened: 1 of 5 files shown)


client/src/containers/Header/Header.tsx
```diff
@@ -0,0 +1,23 @@
...
+const Header: React.FC = () => {
+	const { loading, user } = useContext(UserContext);
+
+	if (loading) {
+		return <></>;
+	}
+
+	return (
+		<header>
+			<h1>OEC Data Collection</h1>
```
mmazanec22, line 14: I know this is a common pattern, but according to Deque it's a better practice for navigability to have the `h1` tag for the page be at the start of the `main` content: https://dequeuniversity.com/assets/html/jquery-summit/html5/slides/headings.html

client/src/containers/Upload/Upload.tsx
```diff
@@ -0,0 +1,51 @@
...
+		const file = e.target.files[0];
+		formData.delete("file");
+		formData.set("file", file);
+		return false;
+	}
+	return (
+		<>
+			<p>{status && status.message}</p>
+			<form onSubmit={onSubmit}>
+				<label htmlFor="file"></label>
```
mmazanec22, line 43: htmlFor needs to refer to an ID-- so we should add `id="file"` on the input

client/src/containers/Upload/Upload.tsx
```diff
@@ -0,0 +1,51 @@
...
+			return;
+		}
+		const file = e.target.files[0];
+		formData.delete("file");
+		formData.set("file", file);
+		return false;
+	}
+	return (
+		<>
+			<p>{status && status.message}</p>
```
mmazanec22, line 41: If this is an error message, we should use `aria-describedby` to associate it with the input field-- so the `p` would have an id, and the input would have aira-describedby that id.  We should also add `aria-invalid="true"` to the input field if there's an error (... and I should go make sure we're doing that in ECE reporter!).

We also need to send the focus to the input field if there's an error.

client/src/containers/Upload/Upload.tsx
```diff
@@ -0,0 +1,51 @@
...
+			},
+			body: formData
+		})
+		.then(value => value.json())
+		.then(value => setStatus(value))
+		.catch(_ => setStatus({
+			error: true,
+			message: 'There was an error'
+		}));
+		e.persist();
```
jlhogan, line 25: what does this do ?

cailyncodes: Oh that was just a debugging thing

client/src/containers/Upload/Upload.tsx
cailyncodes: This is going to need to be rewritten once we get the component library in, so going to punt on these changes for now. But 100% will need to address them before releasing any version of this.

PR #63, opened 2020-07-17
Add FileInput component, status indicator, and check data link

cailyncodes:
Should close #20 

This one was a doozy


client/src/assets/styles/index.scss
```diff
@@ -0,0 +1,14 @@
+/*
```
cailyncodes, line 1: Oops can be deleted

client/src/containers/CheckData/CheckData.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+const CheckData: React.FC = () => {
+	return <>Coming soon...</>;
```
cailyncodes, line 4: The check data page in the mocks -- scaffolding here

client/src/containers/Upload/Upload.tsx
```diff
@@ -1,14 +1,29 @@
-import React, { useState, useContext } from 'react';
+import React, { useState, useContext, useEffect } from 'react';
 import AuthenticationContext from '../../contexts/AuthenticationContext/AuthenticationContext';
+import { FileInput, Form, FormSubmitButton, Button, Alert } from '@ctoec/component-library';
 
 const Upload: React.FC = () => {
+	// USWDS File Input is managed by JS (not exclusive CSS)
```
mmazanec22, line 6: This is a clever solution to a weird problem... is it worth using USWDS for this?

client/src/containers/Upload/Upload.tsx
```diff
@@ -17,33 +32,56 @@ const Upload: React.FC = () => {
...
+		.then(value => {
+			setStatus(value)
+		})
+		.catch(_ => {
+			setStatus({
+				error: 'There was an error'
+			})
+		})
+		.finally(() => {
+			setLoading(false)
```
JordanGuinn, line 44: +1 for the removal of potential infinite loading wheels 🎉

client/src/containers/Upload/Upload.tsx
JordanGuinn: What's the USWDS library doing here in particular?

cailyncodes: It creates additional DOM elements that are then subsequently styled by the CSS. Without the JS library, it looks like a browser-native file upload.

PR #67, opened 2020-07-20
Bare bones Fawkes swagger spec pulled from Hedwig

### LINKED ISSUE #56: Create openAPI spec from Hedwig auto generated spec
In dev, Hedwig app serves a generated openAPI yaml spec at `.../swagger/v1/swagger.json`

**AC**:
- use generate Hediwg spec as a starting point for an openAPI spec for data-collection app
- remove routes except /user/current
- add route for /reports PUT (shoudl this be called reports?)

JordanGuinn:
Closes #56.


swagger.json
```diff
@@ -0,0 +1,1204 @@
...
+          "responses": {
+            "200": {
+              "description": "The name of the file uploaded"
+            },
+            "400": {
+                "description": "Bad Request",
+                "content": {
+                  "text/plain": {
+                    "schema": {
+                      "$ref": "#/components/schemas/ValidationProblemDetails"
```
jlhogan, line 67: We should scrap `ValidationProblemDetails` since that's a .Net thing

swagger.json
```diff
@@ -0,0 +1,1204 @@
...
+              "format": "int32",
+              "nullable": true
+            },
+            "lastReportingPeriod": {
+              "$ref": "#/components/schemas/ReportingPeriod"
+            },
+            "validationErrors": {
+              "type": "array",
+              "items": {
+                "$ref": "#/components/schemas/ValidationError"
```
jlhogan, line 303: think we should also scrap `ValidationError` type; probably we will want to reimagine this to be a lil better in the more flexy express backend world

PR #68, opened 2020-07-20
Add TSOA, generate spec and client code

### LINKED ISSUE #57: create dockerized processes to generate server and client stubs
openApi publishes an image of their [generator CLI](https://github.com/OpenAPITools/openapi-generator#16---docker) which we have used successfully in Hedwig to generate typescript fetch client code. It also has option to generate `graphql-nodejs-express-server` and `nodejs-express-server (beta)` servers. 

there is also a [swagger+node specific codegen package](https://www.npmjs.com/package/swagger-node-codegen) that can be used as a CLI tool

Either way, we will need to generate types for the generated server code as there's not currently support for typescript server generation. Maybe this: https://github.com/horiuchi/dtsgenerator

NOTE: If generating typescript server is funky, we can always switch to server code first -> generated spec -> generated client stubs with https://www.npmjs.com/package/tsoa

**AC**
- Add service(s) to docker compose project that can
  - easily, reproducibly, repeatedly generate client code
  - easily, reproducibly, repeatedly generate server stubs

cailyncodes:
Should close #57 

Exciting times!


src/controllers/UserController.ts
```diff
@@ -0,0 +1,21 @@
...
+  Get,
+	Path,
+	Request,
+  Route,
+} from "tsoa";
+import { User } from "../models/User";
+import { UsersService } from "../services/user/UserService";
+
+@Route("users")
+export class UsersController extends Controller {
```
jlhogan, line 12: so tsoa generates all routes, and hooks up controllers to routes based on these decorators?

src/controllers/UserController.ts
```diff
@@ -0,0 +1,21 @@
...
+
+	@Get("current")
+	public async getCurrentUser(
+		@Request() req: Express.Request
+	): Promise<User> {
```
jlhogan, line 17: and presumably tsoa decides which models to expose in the spec based on what is returned by the linked up controller functions

src/controllers/UserController.ts
cailyncodes: Yee

src/controllers/UserController.ts
cailyncodes: Tis my guess

src/services/user/UserService.ts
```diff
@@ -0,0 +1,8 @@
+import { User } from "../../models/User";
+import { users } from "../../data/users";
+
+export class UsersService {
```
JordanGuinn, line 4: extreme nit: class name `UsersService` vs file name `UserService.ts`

src/controllers/UserController.ts
JordanGuinn, line 12: extreme nit #2: file name `UserController.ts` vs class name `UsersController`

src/controllers/UserController.ts
```diff
@@ -0,0 +1,21 @@
...
+		const wingedKeysId = req.sub;
```
JordanGuinn, line 18: Out of curiosity, is there an annotation that can be used to mark this particular request param?

src/controllers/UserController.ts
```diff
@@ -0,0 +1,21 @@
...
+	Request,
+  Route,
+} from "tsoa";
+import { User } from "../models/User";
+import { UsersService } from "../services/user/UserService";
+
+@Route("users")
+export class UsersController extends Controller {
+
+	@Get("current")
```
JordanGuinn, line 14: Love how intuitive these annotations are 👏🏽👏🏽

src/services/user/UserService.ts
cailyncodes: Ooh thanks! Good catch

src/controllers/UserController.ts
cailyncodes: I don't think I follow this question?

JordanGuinn: I'm not sure where `sub` comes from or how it's supposed to be provided by the client here, so I'm wondering if that needs to captured in our swagger spec at all?  And assuming it does, I'm wondering if there are any annotations we can use (similar to `@Request` or `@Route`) to make that happen?

cailyncodes: Check out the `custom.d.ts` file -- we are extending the included `Express.Request` type. (After integrating with OpenAPI auth middleware, this has become `req.User` with our defined `User` object (also documented in `custom.d.ts`)

JordanGuinn: Okay I think I see now - the user is retrieved with the ID made available after decoding the jwt, and is then subsequently attached to the request as part of our middleware run.  Def no need to document that! 🙌🏽🙌🏽

src/index.ts
```diff
@@ -16,7 +18,14 @@ const app = express();
 app.use(json());
 
 /* Register the API middleware */
-app.use('/api', api);
+// Require authentication for all API requests
```
jlhogan, line 21: nit clean up commented out code

src/router/api.ts
```diff
@@ -2,22 +2,17 @@ import express from 'express';
 import multer from 'multer';
 import path from 'path';
 import { handleError } from '../middleware/error';
-import { authenticate } from "../middleware/authenticate";
-import { getUserByWingedKeysId } from '../data/users';
+// import { authenticate } from "../middleware/authenticate";
```
jlhogan, line 5: nit clean up

src/router/api.ts
```diff
@@ -2,22 +2,17 @@ import express from 'express';
...
+
+// TODO: Combine file upload with TSOA
```
jlhogan, line 7: meaning  create decorated controller for file upload route?

cailyncodes: This todo is for another ticket -- this entire file would be scrapped and file upload needs to be added to a controller that is actually used

jlhogan: like.... this ticket https://github.com/ctoec/data-collection/issues/73 :)

cailyncodes: Oops lol realizing I did not really answer your question but yes 73 is great

PR #69, opened 2020-07-20
21: Adds data header validation to CheckData

### LINKED ISSUE #21: Validate upload
As I provider, if I try to upload a CSV or Excel file that is not formatted according to the template, or if I try to upload a file that is not a CSV or Excel file, I get an error that I must upload a CSV or Excel file formatted according to the template.

First pass - check ~file type and~ headers

jlhogan:
should close #21 

this is v1; eventually to be replaced with editable data table

### DIFF
--- a/client/src/containers/CheckData/CheckData.tsx
+++ b/client/src/containers/CheckData/CheckData.tsx
@@ -1,7 +1,90 @@
-import React from 'react';
+import React, { useState, useContext, useEffect } from 'react';
+import AuthenticationContext from '../../contexts/AuthenticationContext/AuthenticationContext';
+import { useParams } from 'react-router-dom';
+import { Alert } from '@ctoec/component-library';
+
+const EXPECTED_HEADERS = [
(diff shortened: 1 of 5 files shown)


client/src/containers/CheckData/CheckData.tsx
```diff
@@ -1,7 +1,90 @@
...
+		if (loading) return;
+		if (error) return;
+		if (data) return;
+
+		setLoading(true);
+		fetch(`/api/reports/${fileUploadId}`, {
+			headers: {
+				authorization: `Bearer ${accessToken}`
+			}
+		})
```
cailyncodes, line 34: If this lands after #68, let's update it to use the API generated code

client/src/containers/CheckData/CheckData.tsx
```diff
@@ -1,7 +1,90 @@
...
+		.catch((_) => {
+			console.error(_);
+			setError("There was an error");
+		})
+		.finally(() => setLoading(false));
+		
+	}, [accessToken, loading, error, data, fileUploadId]);
+
+	console.log("data", data);
```
cailyncodes, line 48: Nit: cleanup

client/src/containers/CheckData/CheckData.tsx
jlhogan: was planning to wait so i could play w that!

cailyncodes: Don't worry about this -- I need base of this PR for #73. So I can take care of this.

jlhogan: not sure i understand ^

jlhogan:
think ii'm going to close this in favor of eventual table PR landing

Issue #72, opened 2020-07-21
integrate DB client 


jlhogan:
we can either use a full ORM, or a lighter-weight driver/driver + query builder plus some migration tool

Driver:
- https://node-postgres.com/

Query builder:
- http://knexjs.org/

ORMs:
- https://github.com/typeorm/typeorm,  https://github.com/typeorm/typescript-express-example
- https://sequelize.org/master/manual/typescript.html

migration tools:
- https://www.npmjs.com/package/db-migrate

PR #74, opened 2020-07-21
Adds postgres backend

### LINKED ISSUE #71: Add psql server to docker-compose project

jlhogan:
should close #71 

and rider: removes ProblemDetails model, which is a holdover from .Net

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -30,6 +30,14 @@ services:
     environment:
       HOST: '0.0.0.0'
       NODE_ENV: development
+      DB_PASSWORD: TestPassword1
+    depends_on:
+      - db
+
+  db:
+    image: postgres:13
+    environment:
+      POSTGRES_PASSWORD: TestPassword1
 
   winged-keys-db:
     image: mcr.microsoft.com/mssql/server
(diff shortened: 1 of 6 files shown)


src/models/problemDetails.ts
```diff
@@ -1,8 +0,0 @@
-export interface ProblemDetails {
```
JordanGuinn, line 1: 👏🏽👏🏽👏🏽

package.json
```diff
@@ -18,6 +18,7 @@
 		"http-proxy": "^1.18.1",
 		"jwks-rsa": "^1.8.1",
 		"multer": "^1.4.2",
+		"node-postgres": "^0.6.2",
```
JordanGuinn, line 21: Is `node-postgres` here just for this initial POC, or is this to say we'll be using drivers + raw SQL over an ORM (captured in #72)?

docker-compose.yaml
```diff
@@ -30,6 +30,14 @@ services:
     environment:
       HOST: '0.0.0.0'
       NODE_ENV: development
+      DB_PASSWORD: TestPassword1
+    depends_on:
+      - db
+
+  db:
+    image: postgres:13
```
JordanGuinn, line 38: Well that was easy 🎉

package.json
jlhogan: just here for now to :heavy_check_mark: that we're indeed connected to a working postgres serverr

PR #76, opened 2020-07-22
Add & run prettier, and add git hooks

jlhogan:


README.md
```diff
@@ -22,17 +26,7 @@ yarn run tsoa spec
...
-#### Local
-1. Install (if you haven't already) Visual Studio, [Node 12](https://nodejs.org/en/download/) and [Yarn](https://yarnpkg.com/lang/en/docs/install/).
-
-2. Install all corresponding yarn dependencies, based on the static versions specified in `yarn.lock`:
-    ```.sh
-    yarn install --frozen-lockfile
-    ```
-
-3. Follow steps 2 and 3 again from within the `client` directory, located in project root.
-
```
cailyncodes, line 34: I think @JordanGuinn added this section. Including him on disco about remove it

jlhogan: i started on adding instructions for setting up local postgres DB, and it was too annoying,and at that point i figured ppl just arent gna spin up a local postgres so they just have to do docker dev BUT that's not necessarily true :upside_down_face:

jlhogan: merging this to get my other PR into a reasonable state, but open to adding this back to readme if we deem yes!!

JordanGuinn: Honestly, local dev is _already_ difficult as it is on OSX with LocalDB not being supported for `winged-keys` (though I don't know what that pain is like on any other OS).  So I'm fine with removing the bit about local development, so long as we add something indicating that we _really_ only support Docker dev right now.

PR #87, opened 2020-07-29
Remove models in favor of typeorm entities, and rebaselines migrations

jlhogan:


src/entity/Report.ts
```diff
@@ -1,61 +0,0 @@
-import {
```
JordanGuinn, line 1: Are these being removed cuz we don't actually need these report entities after all?  Or at least at this stage?

src/entity/embeddedColumns/UpdateMetaData.ts
```diff
@@ -1,10 +1,10 @@
-import { Column, ManyToOne } from 'typeorm';
+import { ManyToOne, UpdateDateColumn } from 'typeorm';
 import { User } from '../User';
 
 export class UpdateMetaData {
-  @Column()
+  @UpdateDateColumn()
```
JordanGuinn, line 5: Wow, seems like TypeORM has everything covered 👏🏽

src/entity/Report.ts
jlhogan: yeah that was my thinking -- no need to make this already more complicated than it needs to be app even more more complicated than it needs to be. And we have it in version control when/if we want it

src/entity/embeddedColumns/UpdateMetaData.ts
jlhogan: :)

PR #97, opened 2020-08-03
83 check data

### LINKED ISSUE #83: Don't serve file on check data, report EnrollmentReport
When displaying check data, we fetch the present data. Currently, we return the file blob. Change this to return the parsed data in a structured model.

Connect check data to use API data

Blocked by #82

### LINKED ISSUE #84: Connect services to data store
Connect services to a real database instead of in-memory solutions.

jlhogan:
closes #83 and closes #84


jlhogan:
fixing up client routing

src/services/enrollmentReport/EnrollmentReportService.ts
```diff
@@ -0,0 +1,83 @@
...
+          column.propertyName !== 'reportId'
+      )
+      .map((column) => column.propertyName);
+
+    const parsedSheet = utils.sheet_to_json(sheet, {
+      range: this.getStartingRow(sheet),
+      header: expectedHeaders,
+    });
+
+    console.log('parsed sheet', parsedSheet);
```
cailyncodes, line 40: Console log

src/services/enrollmentReport/EnrollmentReportService.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * Gets the boolean value for a raw string boolean.
+   *
+   * Empty or missing string, 'N', and 'No' return false.
+   * Other values return true.
+   * @param value
+   */
+  private getBoolean(value: string): boolean {
+    if (['', 'N', 'NO', undefined].includes(value.toUpperCase())) return false;
```
JordanGuinn, line 107: Super ridiculous nitpick: this will break if `undefined` is supplied because of that `toUpperCase()` bit

jlhogan: it shouldn't actually ever be undefined if i understand the parsing bit correctly BUT better type safe than type sorry :)

PR #119, opened 2020-08-04
Fawkes Staging Deployment

JordanGuinn:
### Description

This PR includes all the Fawkes changes necessary to allow for staging deployments, both through CodePipeline and manually from the CLI.  There will need to be some follow-up work to distinguish staging deploys from prod deploys (and devsecure deploys later on), but that can be addressed at a later time.

### Sibling PRs

- https://github.com/skylight-hq/ctoec-devops/pull/59
- https://github.com/ctoec/winged-keys/pull/52


package.json
```diff
@@ -46,9 +46,6 @@
     "watch-test": "npm run test -- --watchAll",
     "watch-ts": "tsc -w"
   },
-  "engines": {
```
JordanGuinn, line 49: This was unfortunately necessary because Elastic Beanstalk wants an exact version to be deployed, which apparently isn't covered w/Node ^12 😕

client/.env.production
```diff
@@ -0,0 +1 @@
+REACT_APP_WINGED_KEYS_HOST="https://staging.ece-wingedkeys.ctoecskylight.com/"
```
JordanGuinn, line 1: Eventually we're gonna have to be able to distinguish between prod and staging, but that's more work that we don't need to think about for now

README.md
```diff
@@ -49,3 +49,42 @@ To generate the client code, run the included Bash file: `./generate-client-code
...
 
     ```
+
+## Deploy
+
+Deployments can either be triggered directly through the corresponding Azure pipeline for the applicable stage, or directly from the CLI.  Additionally, we also don't utilize Docker in any of our deployment environments, but rather Linux/Node for the sake of simplicity.
+
+### Pipeline
+
+### Command Line
```
cailyncodes, line 59: Nit: I think we want another `#` here?

README.md
```diff
@@ -49,3 +49,42 @@ To generate the client code, run the included Bash file: `./generate-client-code
...
+NOTE: This option will require you to have the Elastic Beanstalk CLI installed on your machine.  If you're on OSX, [Homebrew](https://formulae.brew.sh/formula/aws-elasticbeanstalk) is your best bet - otherwise, following the setup scripts outlined in the [AWS docs](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/eb-cli3-install.html) works just fine.
+
+1. Install the dependencies for the React application.
+```bash
+cd <path-to-project-root>/client && yarn install
```
cailyncodes, line 64: Nit: Should this have `--frozen-lockfile`?

client/src/contexts/AuthenticationContext/AuthenticationContext.tsx
```diff
@@ -126,7 +126,7 @@ const AuthenticationProvider: React.FC<AuthenticationProviderPropsType> = ({
       if (!!defaultOpenIdConnectUrl) {
         setOpenIdConnectUrl(defaultOpenIdConnectUrl);
       } else if (!openIdConnectUrl) {
-        const wingedKeysUri = await getConfig('WingedKeysUri');
+        const wingedKeysUri = process.env.REACT_APP_WINGED_KEYS_HOST;
```
cailyncodes, line 129: What's the reason for this change?

src/middleware/authenticate.ts
```diff
@@ -8,8 +8,7 @@ import jwks from 'jwks-rsa';
 import { UserService } from '../services/user/UserService';
 
 const jwksClient = jwks({
-  // @TODO Use ENV var for URI host
-  jwksUri: 'https://winged-keys:5050/.well-known/openid-configuration/jwks',
+  jwksUri: `${(process.env.WINGED_KEYS_HOST || 'https://localhost:5050/')}.well-known/openid-configuration/jwks`,
```
cailyncodes, line 11: This might conflict with Julia's open PR just fyi

jlhogan: easy to reconcile and i am already ready to take on that job as the person introducing SWEEPING changes during active dev :) :) :)

jlhogan:
i'm seeing a switch from json-based config stuff to env-var based config stuff. that's generally fine but to your point of
> Eventually we're gonna have to be able to distinguish between prod and staging

leveraging azure deploy pipelines, we can really easily inject values into json files, on a per-environment basis

README.md
JordanGuinn: Ahhh, yes - good call!  Will update.

client/src/contexts/AuthenticationContext/AuthenticationContext.tsx
JordanGuinn: I just wanted to get this working for now, and specifying a dev env and a prod env config was the simplest way to get staging up without actively breaking local dev.  This is eventually gonna need to be adjusted in order to distinguish prod from staging (since those'll both use a production build), but that seemed kinda cart-before-horsey to include with this PR.

JordanGuinn:
I'm sure AWS' CodePipeline has something equivalent for that kinda thing, but I honestly was gonna wait until we had another stage to get set up before digging in and finding it.  I was hoping we could just use the equivalent var in the Elastic Beanstalk environment that Express is utilizing, but that's unfortunately a no-go 😕

Whatever we do though, I really want to make sure that it's captured in source control (which is just not a thing w/Azure release pipelines)

README.md
JordanGuinn: Whoops, I was actually meaning to put something below "Pipeline" 🤦🏽‍♂️ Updating

Issue #104, opened 2020-08-04
Expandable Roster:  I can quickly see all data for each enrollment record organized by section.


RyanHansz:
![Separate page child record](https://user-images.githubusercontent.com/63253539/90180530-18815100-dd7d-11ea-9561-b47969dd947a.gif)
[Designs](https://share.goabstract.com/4128fcab-2a0b-4cea-9889-a5a0ec51a846)

1. Reduce columns shown in the table to: Name, Birthdate, Age group, Funding type, Contract space, Site and Enrollment date. 
2. Turn child name into a link
3. Clicking a child's name takes you to their record - with tabs for each section (Child Info, Family info, income, enrollment/funding and C4K). Child info is opened by default.
4. Each section will use form elements from ECE reporter
5. Withdraw will function the same as ECE reporter
6. Delete will open a confirmation modal, and confirming will delete the record and return the user to the roster.

RyanHansz:
- putting in page with tab navs
- ticket for each section and related form from ECE reporter

RyanHansz:
- Withdraw

PR #132, opened 2020-08-06
24 data accuracy

### LINKED ISSUE #24: As a provider who has uploaded my data, I can view the data I've upload to confirm accuracy.
[Mockup](https://app.abstract.com/share/5ea2033d-b4eb-4322-8a2a-1b601b77c8fc?collectionId=e2036f17-be42-40a0-9a0b-aa94c3efbc51&collectionLayerId=d1e55b27-ba88-4edb-adfe-761469fcb9c8&present=true&preview=false&sha=c48a65163e8532e6de4b7594d18763f321e83306)

After I have uploaded my data, I can click "Check your data" to see a view of the data I have uploaded.

I see how many records I have uploaded.
I see a table that shows the data I've uploaded.

If this information looks incorrect, I can go "Back to upload" to re-upload a file.

mmazanec22:
Closes #24 .

@RyanHansz some styling questions:

1. The table library puts a scroll bar at the bottom of the container.  It's annoying to have to scroll down that far to be able to see the other columns.  How should horizontal scroll work?
2. We don't want table pagination, correct?  It had it by default and was easy to turn off, and would be as easy to turn back on whatever the decision is.
3. The mockup shows the table being the same width as the header, and the button box thing being that same width too.  Is this what we're after?  Left to my own devices I would have assumed full width-- again, flexible either way.
4. How do I get specific box shadow settings out of abstract?  I tried to create it faithfully for the bottom button thing but haven't quite gotten there.  Does that exist somewhere or do I just need to keep tweaking it?


RyanHansz:
Hi Melanie, great questions! 

1. Horizontal scroll: It'd be great if the scroll bar is fixed to the bottom of the window instead of the table container. If that's difficult, then we can leave it until #104 removes the need for a scrollable table.
2. No pagination is preferred!
3. Full-width is preferred
4. You can see it in the inspect section of abstract (you may have to make an account first)
Screen Shot 2020-08-06 at 5 23 24 PM

client/src/containers/CheckData/EditableTable.tsx
```diff
@@ -0,0 +1,55 @@
...
+type EditableTableProps<T extends object> = {
+  title: string;
+  data: T[];
+  columns: Column<T>[];
+  onRowAdd?: any;
+  onRowUpdate?: any;
+  onRowDelete?: any;
+};
+
+// TODO: move this to component library
```
JordanGuinn, line 15: Is this the final component remaining that needs to be moved to `component-library?  Is there a GitHub Issue somewhere that reflects that work, if not?

Just wanna make sure we don't lose sight of that TODO 🙂

jlhogan: https://github.com/ctoec/component-library/issues/24!

PR #135, opened 2020-08-11
No more tsoa

jlhogan:
Huzzah! So i personally think the end result here is some much more "vanilla" backend code that feels more accessible and for me has been much easier to tweak, dig into, and understand. I imagine the diff here is gonna be hard to look at, but hopefully looking at the new code on its own shouldn't be too bad -- I tried to add lots of informative comments. 

Down for this to be rejected if people are really loving tsoa, but wanted to at least put our eyes on an alternative.


client/src/containers/DataDefinitions/DataDefinitions.tsx
```diff
@@ -14,9 +14,9 @@ const DataDefinitions: React.FC = () => {
   );
 
   useEffect(() => {
-    getApi(accessToken)
-      .getDataDefinitions()
-      .then((definitions) => setDataDefinitions(definitions));
+    apiGet('data-definitions').then((definitions) =>
```
cailyncodes, line 17: Is there a way we can do this without the magic string?

src/entity/decorators/dataDefinition.ts
```diff
@@ -0,0 +1,43 @@
+import 'reflect-metadata';
```
cailyncodes, line 1: What does this file do?

src/middleware/authenticate.ts
```diff
@@ -1,73 +1,50 @@
...
+ * Authentication middleware to decode auth JWT (JSON web token)
+ * with appropriate key from JWKS (JSON web key set).
+ * Leverages express-jwt, a middleware for validating JWT (https://github.com/auth0/express-jwt#readme)
+ *
+ * Adds the resulting claims to the "claims" property on the request.
+ */
+const decodeClaim = jwt({
+  secret: jwks.expressJwtSecret({
+    cache: true,
+    jwksUri: 'https://winged-keys:5050/.well-known/openid-configuration/jwks',
```
cailyncodes, line 19: We could add the todos (or create tickets) for the env host and ssl support

src/middleware/authenticate.ts
```diff
@@ -1,73 +1,50 @@
...
+ * which represents the WingedKeysId for a user, to lookup the authed user.
+ *
+ * Adds the resulting User object to the "user" property on the request.
+ */
+const addUser = passAsyncError(
+  async (req: Request, res: Response, next: NextFunction) => {
+    if (req.claims.sub) {
+      const user = await getManager().findOne(User, {
+        where: { wingedKeysId: req.claims.sub },
+      }); // const user = await getManager().findOne(User, { where: { wingedKeysId: req.claims.sub } })
```
cailyncodes, line 37: Extra comment

src/middleware/error/passAsyncError.ts
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * To avoid having to write the same try/catch logic
+ * in every single async middleware function,
+ * wrap them in this to handle passing promise rejections
+ * and throw exceptions along to the error handling middleware
+ *
+ * See `routes/enrollmentReport.ts` for an example.
+ * @param func
+ */
+export function passAsyncError(func: RequestHandler) {
```
cailyncodes, line 12: This is cool!

src/utils/parseUploadedTemplate.ts
```diff
@@ -0,0 +1,102 @@
...
+  ).map((column) => column.propertyName);
+
+  /**
+   * Gets the row at which data starts for each template type.
+   * In excel template, there are 3 rows of header data (section
+   * headers, column headers, and column descriptions). In the
+   * csv template there is one row of header data (column headers).
+   * @param sheet
+   */
+  function getStartingRow(sheet: WorkSheet): number {
```
cailyncodes, line 30: I'm not a huge fan of function declarations within a function body. While this isn't wrong, thoughts on instead moving these functions to be non exported functions in this file?

src/middleware/error/passAsyncError.ts
jlhogan: thank you internet !

src/entity/decorators/dataDefinition.ts
jlhogan: this already exists i just moved it from `src/decorators` to `src/entity/decorators`

this is a custom decorator that augments the FlattenedEnrollment model with data defn metadata, currently used to populate the  data definitions table view in the FE. I preferred this approach because I believe keeping the metadata with the object itself will encourage us to keep everything up to do date -- one source of truth

client/src/containers/DataDefinitions/DataDefinitions.tsx
jlhogan: yes was thinking about pulling route constants into the shared folder and using them in both the FE & BE, but didn't know if it was overkill.

if you're also having this reaction, then i'm gonna go ahead and do it.

jlhogan:
>What is the motivation to also get rid of the services?

typeorm pretty much provides "services" via the [entity manager](https://github.com/typeorm/typeorm/blob/master/docs/entity-manager-api.md) so they felt like overkill -- trying to reduce the amount files / boilerplate as that was IMO a big advantage of leaving the dotnet web service world

client/src/containers/DataDefinitions/DataDefinitions.tsx
cailyncodes: Is there a way we can do it without even those constants? Like can we draw inspiration from what we did with `useApi` where we pass an anonymous function?

src/entity/decorators/dataDefinition.ts
cailyncodes: Gotcha, sounds good!!

cailyncodes:
> > What is the motivation to also get rid of the services?
> 
> typeorm pretty much provides "services" via the [entity manager](https://github.com/typeorm/typeorm/blob/master/docs/entity-manager-api.md) so they felt like overkill -- trying to reduce the amount files / boilerplate as that was IMO a big advantage of leaving the dotnet web world and makes understanding this codebase that much easier
> 
> felt reasonable enough to me to encapsulate file parsing in a util for now. Should we end up in a place where much more complex service-level logic is needed, I'm not opposed to bringing them back but overall the goal here was as simple as possible

I can get behind it!

src/utils/parseUploadedTemplate.ts
jlhogan: so the reason this is like this is because i felt overall the thing was more understandable + cleaner with the derived consts pulled out and defined at the top. However, that has to happen inside the function to ensure the call to `getConnection()` does not happen on app start up, before the connection is initialized but rather when the util is called.  then i didn't want to have to be passing those consts around into all the helper functions, so ended up just definition the helper funcs within this one.

down to change it, as i have no strongly held opinion in either direction

client/src/containers/DataDefinitions/DataDefinitions.tsx
jlhogan: need a little more to understand what you have in mind, let's chat ?

Something I want to avoid is over engineering this to the point that it is less understandable to the general js/ts web dev. fetch API is generally well understood, or at least googled about. This getApi is a very (at least IMO) lightweight and low-complexity wrapper around that. What does passing anonymous functions look like?

src/utils/parseUploadedTemplate.ts
cailyncodes: Hmm I see what you are saying.

First thought is we could have an init method that returns an array of those needed constants and then have those helper functions take in another param. Maybe it makes it more difficult tho?

PR #141, opened 2020-08-12
21: validate template matches expected columns

### LINKED ISSUE #21: Validate upload

jlhogan:
closes #21


src/entity/FlattenedEnrollment.ts
```diff
@@ -173,7 +173,7 @@ export class FlattenedEnrollment implements FlattenedEnrollmentInterface {
```
jlhogan, line 173: changes here are to match the checked in template until (or if ever) template is generated from the model

src/entity/FlattenedEnrollment.ts
```diff
@@ -356,7 +356,7 @@ export class FlattenedEnrollment implements FlattenedEnrollmentInterface {
     formattedName: 'Provider',
     required: REQUIRED,
     definition: 'The provider from which the child is receiving services.',
-    reason: 'Used to link child information to provider data.',
+    reason: 'Used to link child information to provi data.',
```
jlhogan, line 359: typo

src/entity/decorators/dataDefinition.ts
```diff
@@ -22,7 +23,18 @@ export const UTILIZATION_REPORTING_REASON =
...
+  required: string;
+  definition: string;
+  reason: string;
+  format: string;
+  example: string;
+  section: string;
+  propertyName: string;
+}
+
+type DataDefinitionInfoInput = {
```
jlhogan, line 37: allow the user to set data defn info without providing propertyName, which should come from the model 
see `getDataDefinition` below

todo: add comment

src/entity/FlattenedEnrollment.ts
```diff
@@ -464,7 +464,7 @@ export class FlattenedEnrollment implements FlattenedEnrollmentInterface {
 
   @Column({ nullable: true })
   @DataDefinition({
-    formattedName: 'First funding reporting period',
+    formattedName: 'First funding period',
```
cailyncodes, line 467: Looks like every other column is every word caps, but this is only sentence case. Just confirming this is intended

src/entity/FlattenedEnrollment.ts
```diff
@@ -477,7 +477,7 @@ export class FlattenedEnrollment implements FlattenedEnrollmentInterface {
 
   @Column({ nullable: true })
   @DataDefinition({
-    formattedName: 'Last funding reporting period',
+    formattedName: 'Last funding period',
```
cailyncodes, line 480: Same here

src/routes/enrollmentReport.ts
```diff
@@ -37,12 +37,29 @@ router.post(
...
-      })
-    );
-    res.send(report);
+    try {
+      const { EXPECTED_HEADERS, headers, enrollments } = parseUploadedTemplate(
+        req.file
+      );
+
+      EXPECTED_HEADERS.forEach((header, idx) => {
+        if (header !== headers[idx]) console.log(header, headers[idx]);
```
cailyncodes, line 46: Do we want this console.log ?

src/utils/parseUploadedTemplate.ts
```diff
@@ -1,9 +1,15 @@
 import { readFile, utils, WorkSheet } from 'xlsx';
 import { getConnection, getManager } from 'typeorm';
 import { FlattenedEnrollment } from '../entity';
+import { getDataDefinition } from '../entity/decorators/dataDefinition';
 
 /**
- * Parses the uploaded file into an array of FlattenedEnrollments
+ * Parses the uploaded file into: 
+ * 	- an array of column header strings from the uploaded template
```
cailyncodes, line 8: Nit: spacing

src/entity/FlattenedEnrollment.ts
jlhogan: yeah no nothing about the way these titles do (rather do not) match is intentional but fixing the templates is annoying so i am just leaving them as is until we update the BE to serve generated templates from this metadata. At that point we should fix these to be consistently sentence or title case :)

src/routes/enrollmentReport.ts
jlhogan: nope!

PR #143, opened 2020-08-13
Update README for set up

jlhogan:


README.md
```diff
@@ -13,10 +13,21 @@ A web application to collect information from child care providers in the State
...
+
+
+    - Via the containers:
     ```sh
-    docker-compose exec client yarn install --frozen-lockfile
-    docker-compose exec server yarn install --frozen-lockfile
+    docker-compose run --rm --entry-point "yarn install --frozen-lockfile" client
+    docker-compose run --rm --entry-point "yarn install --frozen-lockfile" server
+    ```
+    **NOTE**: This `docker-compose run` command is only necessary on initial set up, since the containers cannot start without installed dependencies. To execute yarn commands once the containers have started, you can use `docker-compose exec`.
```
bamader, line 24: Since these are run commands, do we need to start the container services before we can run installations within them? I.e. `docker-compose start client` and `docker-compose start server`?

bamader: Also, should we reference the `docker-compose up --no-start` command somewhere in here? When I was trying my local install I encountered a build problem where client and server were both skipped because they were supposed to build from an image, but the --no-start command tells docker-compose that it needs to build those images so that the client and server can be started. This might be captured already in the run entry-point commands you wrote, just want to check.

jlhogan: no you don't need to start them, run brings up an instance of the container and with `--rm` flag also removes it when the command finishes

jlhogan: im sure there are other ways to accomplish this set of tasks via docker-compose, but this is what my brain did with the commands i knw :)

bamader: Totally get it! I did a trial run and it seemed to work so I think everything's good to go.

PR #144, opened 2020-08-13
Alters income fields in DB to be decimal, not integer

jlhogan:
Also adds some changes to DB that were made to model for missing ENUM
types

### DIFF
--- a/src/migration/1597328000967-UpdateIncomeFieldsMakeDecimal.ts
+++ b/src/migration/1597328000967-UpdateIncomeFieldsMakeDecimal.ts
@@ -1,0 +1,60 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class UpdateIncomeFieldsMakeDecimal1597328000967
+  implements MigrationInterface {
+  name = 'UpdateIncomeFieldsMakeDecimal1597328000967';
+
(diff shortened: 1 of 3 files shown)


src/migration/1597328000967-UpdateIncomeFieldsMakeDecimal.ts
```diff
@@ -0,0 +1,60 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class UpdateIncomeFieldsMakeDecimal1597328000967
+  implements MigrationInterface {
+  name = 'UpdateIncomeFieldsMakeDecimal1597328000967';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "site" DROP COLUMN "region"`);
+    await queryRunner.query(
+      `CREATE TYPE "site_region_enum" AS ENUM('East', 'NorthCentral', 'NorthWest', 'SouthCentral', 'SouthWest')`
```
cailyncodes, line 10: Looks like there are some other things happening in the migration -- just want to make sure that's expected

jlhogan: yes see comment in PR :) made those changes a while ago to resolve errors and never ran the migration generate stuff whoops

Issue #91, opened 2020-07-30
I can export my data to get it back out


RyanHansz:
As a person enters data into Fawkes, they can export a .csv of their current data view.

Once the data is submitted to OEC, the person entering data can download a .csv copy of their enrollment data.

PR #156, opened 2020-08-14
Add backend API for downloading enrollment data

cailyncodes:
Partial towards #91 

Adds the backend api for downloading generated CSV file for enrollment data.

Infrastructure note: We'll want to cron job cleaning up files in `/tmp/downloads/` in order to prevent the file system from being exhausted. Alternatively, we can have the app do it's own clean up but there might be some difficulty in managing the async aspect of HTTP requests.


src/routes/enrollmentReportDownload.ts
```diff
@@ -0,0 +1,40 @@
...
+  passAsyncError(async (req, res, next) => {
+    const id = parseInt(req.params['reportId']) || 0;
+    const report = await getManager().findOne(EnrollmentReport, id);
+    if (!report) throw new NotFoundError();
+
+    const enrollments = report.enrollments;
+    const stream = format();
+
+    const randomString = uuid();
+    const filename = `/tmp/downloads/${randomString}.csv`;
```
JordanGuinn, line 28: Not saying we necessarily would want to do this _now_, but we _could_ send these files directly to S3 with a temporary expiration set, as opposed to writing them directly to the EC2 filesystem that Fawkes runs on.

jlhogan: should be able to pipe the stream directly to the res, no ? without writing to disk or s3?

JordanGuinn: Ah right, this is _download_, not upload.  Yeah I'd think so.

src/middleware/error/errors.ts
```diff
@@ -1,7 +1,7 @@
 import { UnauthorizedError } from 'express-jwt';
```
jlhogan, line 1: curious that code wasn't formatted on my machine....

src/middleware/error/errors.ts
```diff
@@ -10,7 +10,7 @@ export class ApiError extends Error {
  */
 export class InvalidSubClaimError extends ApiError {
   constructor() {
-		super("No valid user found for decided sub claim");
+    super('No valid user found for decided sub claim');
```
jlhogan, line 13: note to self typo: `decided` -> `decoded`

src/routes/index.ts
```diff
@@ -13,3 +14,8 @@ router.use('/data-definitions', dataDefinitionRouter);
 /* AUTHENTICATED ROUTES */
 router.use('/users', authenticate, userRouter);
 router.use('/enrollment-reports', authenticate, enrollmentReportRouter);
+router.use(
+  '/enrollment-reports/download',
```
jlhogan, line 18: nit: should we register sub-routes in the router to stick to a convention and keep this file super simple. Maybe make an `enrollmentReports` dir in `routes` if putting the stuff from `src/routes/enrollmentReportDownload.ts` into `src/routes/enrollmentReport.ts` feels too cluttered?

src/routes/enrollmentReportDownload.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * /enrollment-reports/:reportId GET
+ *
+ * Returns the given EnrollmentReport as a CSV
+ */
+router.get(
+  '/:reportId',
+  passAsyncError(async (req, res, next) => {
+    const id = parseInt(req.params['reportId']) || 0;
```
jlhogan, line 20: since the reportId is the identifier, should it be a path param instead of query param ?

src/middleware/error/errors.ts
```diff
@@ -21,7 +21,7 @@ export class InvalidSubClaimError extends ApiError {
  */
 export class NotFoundError extends ApiError {
   constructor() {
-		super("Resource not found");
+    super('Resource not found');
```
JordanGuinn, line 24: Just out of curiosity - any particular reason for the shift from double quotes to single quotes here?

src/routes/enrollmentReportDownload.ts
cailyncodes: I tried piping straight to the res object but was unable to get the headers and encoding to work successfully. I can take another stab at it though, or if anyone has suggestions/pointers on a way to do it.

src/middleware/error/errors.ts
cailyncodes: I think the formatter did this?

PR #157, opened 2020-08-15
Some changes to models

jlhogan:
- enums are only defined in shared/models/enums, not also in server
- renamed Unspecified to NotSpecified so that all gender enum options
start with a different letter, which makes parsing the
FlattenedEnrollment into child a little bit easier and why not
- server now transpiles shared node_module because it was all of a
sudden getting mad and unable to find the module without this
- so, updates references to built `dist/` resources to be at `dist/src`
- adds special ed fields to child
- condenses addressLine1 and addressLine2 into streetAddress on family
- makes all server entity models in `src/entity` extend the appropriate interfaces from
`shared/models`

### DIFF
--- a/client/tsconfig.json
+++ b/client/tsconfig.json
@@ -15,5 +15,5 @@
     "noEmit": true,
     "jsx": "react"
   },
-  "include": ["src", "../shared/models"]
+  "include": ["src"]
 }
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "data-collection",
   "version": "0.0.1",
-  "main": "dist/index.js",
+  "main": "dist/src/index.js",
   "repository": "https://github.com/ctoec/data-collection.git",
   "license": "MIT",
   "dependencies": {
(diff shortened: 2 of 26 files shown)


tsconfig.json
```diff
@@ -21,5 +21,5 @@
 
     "forceConsistentCasingInFileNames": true
   },
-  "include": ["src/**/*"]
+  "include": ["./src", "./node_modules/shared/"]
```
JordanGuinn, line 24: What was the exact error that made this necessary?  And why is `shared` being referenced from `node_modules` and not just from `./shared`?

I'd think that because the contents of `shared/models` are entirely TS interfaces, it wouldn't make a difference whether it was included in compilation or not (since they're entirely removed during transpilation).

client/tsconfig.json
```diff
@@ -15,5 +15,5 @@
     "noEmit": true,
     "jsx": "react"
   },
-  "include": ["src", "../shared/models"]
+  "include": ["src"]
```
jlhogan, line 18: this doesn't seem to be doing anything, i have NO CLUE why I added it, as far as i can tell the FE build succeeds without it.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "data-collection",
   "version": "0.0.1",
-  "main": "dist/index.js",
+  "main": "dist/src/index.js",
```
jlhogan, line 4: relative imports of `../shard` in the src code mean dist ends up containing `src` and `shared` dirs at top level

tsconfig.json
```diff
@@ -21,5 +21,5 @@
 
     "forceConsistentCasingInFileNames": true
   },
-  "include": ["src/**/*"]
+  "include": ["src"]
```
jlhogan, line 24: for consistency with `client/tsconfig.json`, and doesn't seem to make a diff???

jlhogan:
the mystery of how this works for the FE remains a mystery to me....

tsconfig.json
JordanGuinn: Really?  That's... surprising, but works for me!

jlhogan: yup appears that `src` === `src/**/*` ¯\_(ツ)_/¯

PR #167, opened 2020-08-17
142 readiness

### LINKED ISSUE #142: I can enter data for school readiness funded enrollments
**Funding types:**
CSR (Competitive School Readiness)
PSR (Priority School Readiness)

**Space types:**
Full Day	
School Day
Part Day
Extended Day
		

**Other data needed:**
On the backend, we will need to associate each school readiness funded program with a school readiness community (and liaison eventually)

bamader:
Closes #142 
Created a barebones community model in the shared directory, then implemented its interface as an Entity (followed the paradigm set by Organization here). I know Julia is working on a PR to eliminate the enums from the backend (and just rely on the shared dir), but since this wasn't committed yet I just followed what was there.

### DIFF
--- a/shared/models/Community.ts
+++ b/shared/models/Community.ts
@@ -1,0 +1,7 @@
+import { Organization } from '.';
+
+export interface Community {
+  id: number;
+  name: string;
+  organizations?: Array<Organization>;
+}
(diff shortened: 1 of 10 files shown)


jlhogan:
the PR with changes to shared models + entities landed, so would be awesome if you could update this accordingly:
- update the enums in `shared/models`
- make `Community` entity implement the `shared/model` interface (see other entities for example of how to do that)

bamader:
Yep I'll make those changes!

bamader:
Looks like I'd already updated the enums in `shared/models`, so I just had to update the interface implementation

PR #169, opened 2020-08-18
Update index.html ref in non-dev environments

JordanGuinn:


JordanGuinn:
Confirmed this works both locally and in AWS.

PR #170, opened 2020-08-18
Added new funding types to enums

### LINKED ISSUE #121: I can enter data for smart start funded enrollments.
**Funding type:**
SS - (Smart Start)

**Space type:**
School-day/school-year

### LINKED ISSUE #122: I can enter data for state head start funded enrollments.
**Funding type:**
SHS (State Headstart)

**Space types:**
Extended Day
Extended Year

bamader:
Closes #121 and Closes #122 
It looks like there isn't any standardization between the funding times of various funding sources. For example, different funding sources reference "full day" and "school day," which sound like the same thing but are currently treated separately. Same thing with "expanded day." I'm guessing this is an artifact of the agencies / funding sources rather than something by design on our end, but should we be handling this in any specific way? As in, is it feasible for us to consolidate these different names for the same things down into one 'type' (i.e. full day and school day both now just refer to 'full day')?

### DIFF
--- a/shared/models/FundingSource.ts
+++ b/shared/models/FundingSource.ts
@@ -1,6 +1,5 @@
 export enum FundingSource {
   CDC = 'CDC',
-  SchoolReadiness = 'SchoolReadiness',
   CSR = 'CSR',
   PSR = 'PSR',
   SHS = 'SHS',
--- a/shared/models/FundingSource.ts
+++ b/shared/models/FundingSource.ts
@@ -3,4 +3,6 @@ export enum FundingSource {
   SchoolReadiness = 'SchoolReadiness',
   CSR = 'CSR',
   PSR = 'PSR',
+  SHS = 'SHS',
+  SS = 'SS',
 }
--- a/src/entity/enums/FundingSource.ts
+++ b/src/entity/enums/FundingSource.ts
@@ -1,8 +1,0 @@
-export enum FundingSource {
-  CDC = 'CDC',
-  SchoolReadiness = 'SchoolReadiness',
-  CSR = 'CSR',
-  PSR = 'PSR',
-  SHS = 'SHS',
-  SS = 'SS',
-}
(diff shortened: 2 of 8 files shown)


shared/models/FundingSource.ts
```diff
@@ -3,4 +3,6 @@ export enum FundingSource {
   SchoolReadiness = 'SchoolReadiness',
```
jlhogan, line 3: can we make this `SR = 'SR'` to match all the other values?

src/entity/enums/FundingSource.ts
```diff
@@ -3,4 +3,6 @@ export enum FundingSource {
   SchoolReadiness = 'SchoolReadiness',
```
jlhogan, line 3: also sorry I guess i didn't fully clean up in my other PR but these `src/entity/enums` should exist, can you just delete the whole dir?

shared/models/FundingSource.ts
jlhogan, line 3: can we make this `SR = 'SR'` to match all the other abbreviated values while you're in there?

src/entity/enums/FundingSource.ts
bamader: Yeah I'll kill the whole enums folder inside `src/entity`

shared/models/FundingSource.ts
bamader: Yep, will enforce an abbreviation convention

bamader: Wait, based on the issue Ryan opened about adding School readiness funding types, SR can occur as either PSR or CSR, both of which are already implemented. Should we just remove SchoolReadiness altogether then @jlhogan ?

jlhogan: yes

jlhogan:
>As in, is it feasible for us to consolidate these different names for the same things down into one 'type' (i.e. full day and school day both now just refer to 'full day')?

I think the desired end state here is funding-source-specific funding-times, so there's a set of valid funding times for CDC funding, a set of valid funding times for SR fundings, etc etc etc. Not something that needs to be implemented right now, but something to think about.

PR #175, opened 2020-08-18
88 flattened enrollment to enrollment

### LINKED ISSUE #88: Translate upload to database: I can see how Fawkes has translated my spreadsheet into its database.
**AC**
discussed in eng sync:
1) ~save uploaded file~ DONE
2) ~read uploaded file~ DONE
3) ~parse into flattened enrollment~ DONE
4) ~save to flattened enrollment table~ DONE
5) convert to real models
6) send real models back to UI

After uploading data to Fawkes, the user's data is translated into our database to show any errors or changes in formatting. For example, 2019/03/02 gets changed to 03/02/2019. The result is displayed in the Check Data table.

jlhogan:
closes #88 

TODO: Add seed data for non-prod app. I manually added Organizations, Sites, FundingSpaces, and ReportingPeriods to my local while testing this. Make that happen for everyone on non-prod start up.


src/utils/mapFlattenedEnrollment.ts
```diff
@@ -0,0 +1,309 @@
...
+  } catch (err) {
+    console.error('Unable to map row: ', err);
+    return;
+  }
+};
+
+/**
+ * Get organization from our system that matches FlattenedEnrollment
+ * source provider name.
+ * TODO: How do we want to implement fuzzy matching here?
```
bamader, line 68: As far as fuzzy matching goes, I'm assuming you're talking about syntax/usage patterns in the actual text people will type? If that's not the case, then this will be useless. But if it is the case, there's a number of approaches we could take:
- Generic edit distance should account for most spelling errors (incl. additions, subtractions, and letter switches)
- Heuristic association rules can map variant forms to a single entity (i.e. '&', '+', and 'and' all map to 'and' in our database, and that's how we store it)
- Hardline stance of rejecting things our db doesn't recognize (though this might create some confusion; I suppose we could also make a page where each provider/organization is told how to list themselves, but getting people to use that might be more trouble than it's worth)

src/utils/mapFlattenedEnrollment.ts
```diff
@@ -0,0 +1,309 @@
...
+ *
+ *
+ * TODO: Implement some Org/Site access authorization layer.
+ * When/where/how do we layer in internal app authorization?
+ *
+ * TODO: We ran into problems before implementing a REST api but
+ * treating objects like arbitrary GraphQL-style graphs. What shape
+ * should all these data objects be returned in? Nested (implemented
+ * now, mirrors how we did it in ECE Reporter)? Array of dicts containing
+ * objects for every row a la:
```
bamader, line 34: I think an array of dicts return shape makes a lot of sense. If GraphQL-style entities gave you trouble before, then I think an array of super basic object notation can still deliver all the functionality we want in a predictable way without too much complexity. I guess I also lean array of dicts because that's how I mentally conceive of the table. It's not a list of kids who own the properties of having a certain family, income, funding source, etc. Each row is its own record, and that record pertains to a child, family, funding source, etc. So the dict array matches my internal organization. Curious what others think!

src/utils/mapFlattenedEnrollment.ts
```diff
@@ -0,0 +1,309 @@
...
+  return getManager().findOneOrFail(Organization, {
+    where: { name: source.provider },
+  });
+};
+
+/**
+ * Get Site from our system that matches FlattenedEnrollment source
+ * site name.
+ * TODO: Confirm that site belongs to org from same row
+ * TODO: How do we want to implement fuzzy matching here?
```
bamader, line 83: Same things above apply in my mind. I suppose since multiple organizations can belong to the same site (or maybe I got that backward) we could also take things like edit distance a step further and check if any of the edit permutations match a known site, and then ask the user "did you mean XXXXX"

src/utils/mapFlattenedEnrollment.ts
```diff
@@ -0,0 +1,309 @@
...
+  const nameParts = (source.name || '').split(' ');
+  if (nameParts.length < 2) {
+    throw new Error(
+      'Name is required, and must include at least first and last name separated by space'
+    );
+  }
+
+  // Name
+  let firstName, middleName, lastName, suffix;
+  firstName = nameParts[0];
```
bamader, line 108: Do you think we need to code some kind of failsafe or robustness check for if people accidentally enter data in the form "last, first"? Or is that not a real risk?

src/utils/mapFlattenedEnrollment.ts
```diff
@@ -0,0 +1,309 @@
...
+};
+
+/**
+ * Create Child object from FlattenedEnrollment source.
+ * TODO: How do we handle blocking data errors in a single row?
+ * @param source
+ */
+const mapChild = (source: FlattenedEnrollment) => {
+  const SUFFIXES = ['sr', 'jr', 'ii', 'iii'];
+  const nameParts = (source.name || '').split(' ');
```
bamader, line 99: How defensively do we need to code here? Should we do things like:
- strip leading and trailing whitespaces (which may be hard to visually spot in front of a first name)
- use a regex to turn multiple consecutive spaces into a single space (so that the split array doesn't have an artificially empty slot)

src/utils/mapFlattenedEnrollment.ts
```diff
@@ -0,0 +1,309 @@
...
+    }
+  }
+
+  // If we could not find a fundingSpace, we cannot create a funding
+  // so NOTHING is returned.
+  if (
+    source.fundingType ||
+    source.spaceType ||
+    source.firstFundingPeriod ||
+    source.lastFundingPeriod
```
bamader, line 300: Should this whole big or statement have an ! in front? Since it's if we didn't have the info to make a funding?

jlhogan: so this checks if there is user supplied funding info, but we couldn't come up with any funding THEN we error out. If the user didn't supply any funding info, that is a valid state (the system _can_ accept data about children who are not state funded, it just is unlikely that providers will enter it) and no error

bamader: Ooooh I gotcha, my b

client/src/containers/App/App.tsx
```diff
@@ -37,6 +37,7 @@ const App: React.FC = () => {
                 {props.text}
               </a>
             ),
+            path: '///',
```
JordanGuinn, line 40: What's this additional path for?

client/src/containers/App/App.tsx
```diff
@@ -47,6 +48,7 @@ const App: React.FC = () => {
             id: 'feedback-nav',
             title: 'Feedback',
             renderer: () => <MailToLink text="Feedback" />,
+            path: '',
```
JordanGuinn, line 51: And this one?

src/entity/Enrollment.ts
```diff
@@ -21,12 +21,18 @@ export class Enrollment implements EnrollmentInterface {
   @PrimaryGeneratedColumn()
   id: number;
 
-  @ManyToOne((type) => Child, { nullable: false })
+  @ManyToOne(() => Child, { nullable: false })
   child: Child;
 
-  @ManyToOne((type) => Site, { nullable: false })
+  @Column({ type: 'uuid' })
```
JordanGuinn, line 27: Interesting - does the `@ManyToOne` annotation above `child` and `site` not auto-generate the corresponding `childId` and `siteId` columns?

src/data/initialize.ts
```diff
@@ -0,0 +1,127 @@
...
+} from '../entity';
+import {
+  Region,
+  AgeGroup,
+  FundingSource,
+  FundingTime,
+} from '../../shared/models';
+import { reportingPeriods } from './reportingPeriods';
+
+export const initialize = async () => {
```
JordanGuinn, line 17: 👏🏽👏🏽👏🏽👏🏽👏🏽👏🏽

src/index.ts
```diff
@@ -3,24 +3,19 @@ import path from 'path';
...
-        id: 1,
-        wingedKeysId: '2c0ec653-8829-4aa1-82ba-37c8832bbb88',
-        firstName: 'Voldy',
-        lastName: 'Mort',
-      });
-      getManager().save(user);
-    }
+    // Staging != 'development', and we don't currently have a way of distinguishing
+    // env in deployed app BUT we don't have prod env so just always init seed data!
+    // if (isDevelopment()) {
```
JordanGuinn, line 16: Truuuuuuuuue.  We'll prolly also wanna only run these seeds in AWS when the RDS instance is deleted/restored too, so we don't just generate redundant data every time the app is restarted

jlhogan: the way the seed queries are written, they only create data if none exists. I added unique constraints to tables that are legit, from a real life data model stand point and also ensure the insert query will fail if the object exists. I used `orIgnore` in the bootstrapping operations so no stack trace is logged by the orm when these queries fail since thats an ok event

src/entity/Enrollment.ts
jlhogan: it does, but to be able to access them directly (so the relationship can be persisted to db) without making the whole deep object reference (which causes cycle issues with json parsing when writing back out the response) you need to explicitly define the id column in the entity defn

client/src/containers/App/App.tsx
jlhogan: oh that was me trying to get rid of a deprecation warning about `javascript:void` being passed as the path to the anchor object. didn't mean to commit

jlhogan: (it didnt work)

PR #192, opened 2020-08-20
Skeleton tabnav to put forms in

bamader:
Assuming the TabNav PR in the component library gets approved, here's the blank field form navigation from the edit roster page. UI is a work in progress but this at least lets us start getting forms created. Also, as we make them, we can refactor the currently hardcode TabNavProps objects in the middle of CheckData to be their own module entities.

Screen Shot 2020-08-20 at 12 21 03 PM

### DIFF
--- a/client/src/containers/CheckData/CheckData.tsx
+++ b/client/src/containers/CheckData/CheckData.tsx
@@ -13,6 +13,9 @@ import {
   Button,
   Table,
   Column,
+  TextInput,
+  ExpandRow,
+  TabNav,
 } from '@ctoec/component-library';
 import { ReactComponent as Arrow } from '@ctoec/component-library/dist/assets/images/arrowRight.svg';
 import { ProcessStep } from '@ctoec/component-library/dist/components/ProcessList/ProcessStep';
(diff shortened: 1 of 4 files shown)


client/src/containers/CheckData/CheckData.tsx
```diff
@@ -106,28 +107,58 @@ const CheckData: React.FC = () => {
...
-                        type="input"
-                        onChange={() => {}}
-                      />
-                      <Button
-                        className="margin-top-2"
-                        text="A button can go here"
-                      />
-                    </div>
-                    <div className="grid-col-1">
+                    <div >
```
jlhogan, line 113: we went back and forth on this, but actually want to update this to be a separate page, not a row expansion. So the name becomes a button that links to a new route, which shows a page containing the tab nav and some header content (child's name etc)

bamader: Ah gotcha. Will update.

bamader: How's this?

Screen Shot 2020-08-20 at 1 56 03 PM

client/src/containers/CheckData/CheckData.tsx
```diff
@@ -15,8 +15,10 @@ import {
   Column,
   TextInput,
   ExpandRow,
+  TabNav,
 } from '@ctoec/component-library';
 import { ReactComponent as Arrow } from '@ctoec/component-library/dist/assets/images/arrowRight.svg';
+import { ProcessStep } from '@ctoec/component-library/dist/components/ProcessList/ProcessStep';
```
jlhogan, line 21: nit: unused import

PR #193, opened 2020-08-20
Template generation

### LINKED ISSUE #125: generate template from data model
The templates are generated from the FlattenedEnrollment model & served via the backend

JordanGuinn:
Closes #125.


client/src/routes.ts
```diff
@@ -40,8 +40,8 @@ export const routes: RouteConfig[] = [
     unauthorized: false,
   },
   {
-    path: '/data-definitions',
-    component: DataDefinitions,
+    path: '/data-requirements',
```
JordanGuinn, line 43: I wanted to leave the actual URL as something a little more human-friendly than `ColumnMetadata`, even though it's not quite as accurate.

src/entity/decorators/columnMetadata.ts
```diff
@@ -23,38 +23,19 @@ export const UTILIZATION_REPORTING_REASON =
...
-type DataDefinitionInfoInput = {
-  formattedName: string;
-  required: string;
-  definition: string;
-  reason: string;
-  format: string;
-  example: string;
-  section: string;
-};
+type ColumnMetadataInput = Omit<ECEColumnMetadata, "propertyName">
```
jlhogan, line 32: :) ty

src/routes/columnMetadata.ts
```diff
@@ -0,0 +1,129 @@
...
+
+columnMetadataRouter.get('/csv', (req: Request, res: Response) => {
+  streamTemplate(res, 'csv');
+});
+
+columnMetadataRouter.get('/xlsx', (req: Request, res: Response) => {
+  streamTemplate(res, 'xlsx');
+});
+
+function streamTemplate(response: Response, bookType: 'csv' | 'xlsx') {
```
jlhogan, line 31: do you feel ok with these functions pulled out in the router file? If we're decomposing things, should we go ahead and define controller classes? not blocker just conversation starter

src/routes/columnMetadata.ts
```diff
@@ -0,0 +1,129 @@
...
+  const workbook = utils.book_new();
+  utils.book_append_sheet(workbook, sheet);
+
+  return workbook;
+}
+
+/**
+ * Creates template data as array of arrays, derived from FlattenedEnrollment entity.
+ * Template data includes column headers and optionally descriptions, which are stored
+ * along with the entity as DB column comments.
```
jlhogan, line 118: this comment is no longer accurate ("stored along with entity as DB column comments"); they're stored as ColumnMetadata :)

src/routes/columnMetadata.ts
JordanGuinn: Agreed!  This stuff should _absolutely_ go into a controller class - I had a similar thought with some of the utility methods I was writing, but also just wanted to get this PR up.  I'll follow up and split things out, tho

PR #200, opened 2020-08-21
Fix CodePipeline

JordanGuinn:
Looks like CodeBuild is failing in the CodePipeline for Fawkes staging.


client/src/containers/EditRecord/EditRecord.tsx
```diff
@@ -4,8 +4,8 @@ import { useLocation } from 'react-router-dom';
 import { TabNav } from '@ctoec/component-library';
 
 const EditRecord: React.FC = () => {
-    const record = useLocation().state['childName'];
-    const reportId = useLocation().state['reportId'];
+    const record = (useLocation().state as any)['childName'];
```
jlhogan, line 7: this went away in a subsequent change i merged so shouldn't need to merge

PR #201, opened 2020-08-24
Persists parsed flattened enrollment data to DB

jlhogan:
Adds a DB save to the mapping step, and returns all data
as a nested child object, as opposed to a flat map of all object

Also adds an endpoint to GET a single child, and uses that for the
"Edit-Record" view

Also fixes client routing from upload to check data

### DIFF
--- a/src/routes/children.ts
+++ b/src/routes/children.ts
@@ -1,0 +1,33 @@
+import express, { Response, Request } from 'express';
+import { getManager } from 'typeorm';
+import { Child } from '../entity';
+import { passAsyncError } from '../middleware/error/passAsyncError';
+import { NotFoundError } from '..//middleware/error/errors';
+
+export const childrenRouter = express.Router();
+
+/**
+ * /children/:childId GET
+ *
+ * Returns the given child, with all nested data desired by the client
+ * including family, incomeDeterminations, enrollments, and fundings
+ */
+childrenRouter.get(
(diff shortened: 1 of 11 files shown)


src/routes/children.ts
```diff
@@ -0,0 +1,33 @@
...
+    const childId = req.params['childId'];
+    const child = await getManager().findOne(Child, {
+      where: { id: childId },
+      relations: [
+        'family',
+        'family.incomeDeterminations',
+        'enrollments',
+        'enrollments.fundings',
+      ],
+    });
```
bamader, line 27: Code itself looks good! Only question I have (and it's a very general one so this might not even be the appropriate place to put it) is what's the rationale for having the child object own the relations with family and enrollment? I had thought we would maintain a table of just "record" objects with various fields (dictionary style), but the nested data structure here has you accessing family through child.family, which seems more indirect. Can you explain a little just so I understand the intuition / motivation?

jlhogan: from slack convo: 
```
 it is not sustainable to continue to use the reportId/rowId from the flattened enrollment object to get/update data as we move into a world where we have to reconcile existing data with the newly input data. So then what id identifies the single record you want to be updating? childId. 
...
over time, providers will upload reports that need to be ingested into the pool of enrollment/child data they have already put in there
```

PR #202, opened 2020-08-24
162 child info

mmazanec22:
This is a WIP because it was waiting on some other work to be finished (i.e. it modifies the FlattenedEnrollment and that should be changed, and also the component library needs to be updated after the tab nav merge), but wanted to put it up in case there are discrepancies between how I approached this and how we're doing other forms


client/src/contexts/AlertContext/AlertContext.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { createContext, useState, useEffect, useCallback } from 'react';
+import { useLocation } from 'react-router-dom';
+import { AlertProps } from '@ctoec/component-library';
+
+export type AlertContextType = {
```
jlhogan, line 5: so I know alerts were causing us some problems in ECE Reporter world. Do we went to bring this over as is? Do we want to strip it out so we're encouraged to take a moment to evaluate what if anything needs to change?

mmazanec22: I don't remember what was problematic about this but we can ditch it for now if that makes sense!

client/src/containers/EditRecord/ChildInfo/Fields/common.ts
```diff
@@ -0,0 +1,6 @@
+// Common helper type for supplying additional props to field components
+export type ChildInfoFormFieldProps = {
```
mmazanec22, line 2: Is this worth keeping/porting?

client/src/containers/EditRecord/ChildInfo/Fields/BirthCertificateFieldSet.tsx
```diff
@@ -0,0 +1,45 @@
...
+ */
+export const BirthCertificateFormFieldSet: React.FC<ChildInfoFormFieldProps> = ({
+  blockErrorDisplay = false,
+}) => {
+  return (
+    <FormFieldSet<Child>
+      id="birth-certificate-fields"
+      legend="Birth certificate"
+      className="display-inline-block"
+      // status={(enrollment) =>
```
jlhogan, line 20: nit: delete unused code, it exists in hedwig if we need to ref!!

client/src/containers/EditRecord/ChildInfo/Fields/Gender.tsx
```diff
@@ -0,0 +1,55 @@
...
+      preprocessForDisplay={(data) => data}
+      parseOnChangeEvent={(e) => e.target.value}
+      inputComponent={Select}
+      id="gender-select"
+      label="Gender"
+      hint="As identified by family"
+      options={[
+        {
+          value: Gender.Female,
+          text: 'Female',
```
jlhogan, line 25: enums are formatted strings in this world, you should be able to just use the same `Gender.Female` for the text

client/src/containers/EditRecord/ChildInfo/Fields/Sasid.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { FormField, TextInputProps, TextInput } from '@ctoec/component-library';
+import { Child } from '../../../../shared/models';
+/**
+ * Component for entering the SASID of a child in an enrollment.
+ */
+export const SasidField: React.FC = () => {
+  return (
+    <FormField<Child, TextInputProps, number | string | null>
```
jlhogan, line 9: do we need to support number? is a string in the Child model now

client/src/containers/EditRecord/ChildInfo/Fields/common.ts
jlhogan: since we're not using any of these concepts yet, i would get rid of it for now

client/src/containers/EditRecord/EditRecord.tsx
```diff
@@ -30,9 +30,9 @@ const EditRecord: React.FC = () => {
   return rowData ? (
     <div className="grid-container">
       <div className="margin-top-4">
-        <h2>
+        <h1>
```
jlhogan, line 33: lol just went thru and re did all the headers my components use to make them properly nested from this h2, shall update again to account for h1. didnt notice there wasn't one on the page already!

client/src/containers/EditRecord/EditRecord.tsx
```diff
@@ -42,7 +42,7 @@ const EditRecord: React.FC = () => {
           {
             id: 'child-tab',
             text: 'Child Info',
-            content: <span>This is where the child info form goes</span>,
+            content: <ChildInfo child={rowData} passData={handleChange} />,
```
jlhogan, line 45: i am in support of switching to refetch paradigm over client-side object state patching, see #226

client/src/containers/SubmitSuccess/SubmitSuccess.tsx
```diff
@@ -1,50 +1,53 @@
...
+        </Link>
+      </div>
+      <div className="tablet:grid-col-auto margin-top-2 margin-bottom-2">
+        <img src={LogoWithCheckSrc} alt="" />
+      </div>
       <div>
         <h1>You've submitted your data to OEC!</h1>
-        <h5 className="margin-top-1">Date submitted: {new Date().toLocaleDateString()}</h5>
+        <p className="margin-top-1 text-bold">
+          Date submitted: {new Date().toLocaleDateString()}
```
jlhogan, line 32: we're using moment everywhere else, is it better to use that here as well for consistency (and so this can be formatted however we want?)

jlhogan: oops this is just a style diff not your actual diff, so feel free to ignore

client/src/contexts/AlertContext/AlertContext.tsx
jlhogan: when/if we have a need, we can port it back over. For now i think we're fine letting containers instantiate their own alerts since we dont yet have a need to cause an alert to show up after a page transition

client/src/hooks/useCatchAllErrorAlert.ts
```diff
@@ -0,0 +1,37 @@
+import { useState, useEffect, useContext } from 'react';
```
jlhogan, line 1: also this, don't need it now and there's v possibly going to be a simpler way we do errors w/out `useApi`

client/src/hooks/useFocusFirstError.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, DependencyList, useState } from 'react';
```
jlhogan, line 1: this i think makes sense to keep esp if it works for error/warning `Alerts` -- TODO make sure those have some class that we check for here

src/entity/transformers/momentTransformer.ts
```diff
@@ -3,6 +3,12 @@ import moment, { Moment } from 'moment';
 export const momentTransformer = {
   from: (dbDate?: string) => (dbDate ? moment.utc(dbDate) : undefined),
   to: (entityDate?: Moment) => {
-    return entityDate ? entityDate.format('YYYY-MM-DD') : undefined;
+    if (!entityDate) return undefined;
```
jlhogan, line 6: i didn't have a need for this -- what value is the entityDate when it is not a moment?

src/routes/child.ts
```diff
@@ -0,0 +1,58 @@
+import express, { Response, Request, json } from 'express';
```
jlhogan, line 1: do you think we should align all routers/controllers around singular resource name? even tho paths are gonna be plural?

src/routes/child.ts
```diff
@@ -0,0 +1,58 @@
...
+    try {
+      const id = req.params['childId'];
+      const manager = getManager();
+      const _child = req.body;
+      // Delete relations from this object so we don't attempt to update them
+      delete _child.family;
+      delete _child.enrollments;
+      await manager.update(Child, { id }, _child);
+      // Send back the original request so that family and enrollments are still there
+      // Q: is this too sketchy or ok?
```
jlhogan, line 51: no need for this if we have PUT request return 200 OK with empty response body, and then have the client do the fetch again!!!

src/routes/child.ts
```diff
@@ -0,0 +1,58 @@
...
+router.put(
+  '/:childId',
+  json(),
+  passAsyncError(async (req, res, next) => {
+    try {
+      const id = req.params['childId'];
+      const manager = getManager();
+      const _child = req.body;
+      // Delete relations from this object so we don't attempt to update them
+      delete _child.family;
```
jlhogan, line 47: did you confirm that this is necessary? I'm pretty sure the orm requires us to configure relationships as "CASCADING" for this to happen

src/routes/child.ts
```diff
@@ -0,0 +1,58 @@
...
+      delete _child.enrollments;
+      await manager.update(Child, { id }, _child);
```
jlhogan, line 49: a ha! cc @bamader -- this is how we want to do updates :) 

also @mmazanec22  lookign at the docs seems you should be able to just pass `id` straight as the second param, it is an alternative func signature

src/utils/mapFlattenedEnrollment.ts
```diff
@@ -59,11 +60,16 @@ export const mapFlattenedEnrollment = async (source: FlattenedEnrollment) => {
...
-const mapOrganization = (source: FlattenedEnrollment) => {
+const mapOrganization = async (source: FlattenedEnrollment) => {
   if (!source.provider) return Promise.reject('Provider is required');
-  return getManager().findOneOrFail(Organization, {
+  let organization = await getManager().findOne(Organization, {
     where: { name: source.provider },
   });
+  if (!organization) {
+    // Added this for development purposes-- see todo
+    organization = getManager().create(Organization, { name: source.provider });
```
jlhogan, line 70: i dont prefer this to alignment around test seed data

src/routes/child.ts
jlhogan: from source code comment in typeorm:
```
 /**
     * Updates entity partially. Entity can be found by a given condition(s).
     * Unlike save method executes a primitive operation without cascades, relations and other operations included.
     * Executes fast and efficient UPDATE query.
     * Does not check if entity exist in the database.
     * Condition(s) cannot be empty.
     */
``` leads me to believe deleting relations is not necessary

File: src/routes/child.ts [hunk 17]
```diff
@@ -0,0 +1,58 @@
...
+      // Send back the original request so that family and enrollments are still there
+      // Q: is this too sketchy or ok?
+      res.send(req.body);
+    } catch (err) {
+      console.log('Error saving changes to enrollment: ', err);
```
jlhogan, line 54: nit: should be child not enrollment

src/routes/child.ts
mmazanec22: @jlhogan either way, I just changed this to make it consistent-- I'm not attached to plural or singular

src/routes/child.ts
mmazanec22: @jlhogan with the relations on there, it gave me an error, that's why I added the deletes-- is there a way to not do this and also not get an error?  I'm not v familiar with typeorm

client/src/hooks/useFocusFirstError.ts
mmazanec22: @jlhogan I *think* we just need to move focus to the first input element that has an error, and that using an h1 or h2 for the alert itself handles our accessibility needs there-- so no alert class needed: https://www.w3.org/WAI/tutorials/forms/notifications/

src/entity/transformers/momentTransformer.ts
mmazanec22: @jlhogan when I upload a file and console log entity date here it's the format that I had in the spreadsheet-- so in this case a string "05-22-2019"

jlhogan: ah ok so that is not a format that the excel software you are using recognized as a date i guess. I think the fix i'd prefer is more robust date parsing, b/c otherwise there will also be places in dealing with the entity in memory where it is expected to be a moment, but is just a string. I'll open a ticket for this; you can leave for now!

src/routes/child.ts
jlhogan: hmmm that's strange. I haven't played around with it, but will check out your branch and see whats happening. the errors from typeORM are weird, but the sourcecode is p readable and the stackoverflow/gh issues support is p robust. i'm sure we'll figureit out, but also doing the deletes is a totally fine solution :)

client/src/containers/CheckData/CheckData.scss
```diff
@@ -1,31 +1,31 @@
 .CheckData__content {
-  margin-bottom: 5rem;
```
jlhogan, line 2: I feel like we keep going back and forth on formatting of some of these files. Let's take a moment to compare at some point? and ensure we have all the config set up that we need to keep it consistent

client/src/containers/EditRecord/ChildInfo/Fields/Gender.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { FormField, SelectProps, Select } from '@ctoec/component-library';
+import { Child, Gender } from '../../../../shared/models';
+
+/**
+ * Component for entering the gender of a child in an enrollment.
+ */
+export const GenderField: React.FC = () => {
+  return (
+    <FormField<Child, SelectProps, string>
```
jlhogan, line 10: the type on this could maybe be Gender? But doesn't really matter

src/entity/transformers/momentTransformer.ts
```diff
@@ -3,6 +3,13 @@ import moment, { Moment } from 'moment';
...
+    let _entityDate = entityDate;
+    if (!moment.isMoment(entityDate)) {
+      // Need to do this so upload works
+      console.log('ENTITY DATE ', entityDate)
```
jlhogan, line 10: console log

PR #203, opened 2020-08-24
Adds user permissions to authenticated user object on request, and al…

jlhogan:
…so adds communityPermissions all around


JordanGuinn:
I also disabled the GitHub check against CircleCI for now, since that of course needs to be merged in first before it can provide any value 😅

PR #206, opened 2020-08-24
Care4Kids Info Form

bamader:
Not ready to merge, but like Melanie, I thought it would be good to get this up where it can be seen. I know we're blocked on the ticket because of the API, but if we're happy with this way of storing change back in the object I'll move forward with that (the idea is that while the individual Forms in the TabNav can save fields in the Child object representing a row, the EditRecord page will have a button below all the forms for Save & Return to CheckData that will handle the API PUT request).

### DIFF
--- a/client/src/containers/EditRecord/Forms/CareForKidsForm.tsx
+++ b/client/src/containers/EditRecord/Forms/CareForKidsForm.tsx
@@ -1,0 +1,82 @@
+import React, { useState } from 'react';
+import {
+    Form, 
+    FormSubmitButton,
+    FormField,
+    RadioButtonGroupProps,
+    RadioButtonGroup,
+    RadioButton } from '@ctoec/component-library';
+
+/*
+* Basic functional component designed to allow user to edit
+* the Care For Kids field of a data record.
+* Data is not processed using a FlattenedEnrollment.
+* Rather, information is inherited via props from the parent
+* object (the EditRecord page), which will bundle all 
+* edits together and make a single unified record update
+* when the user is done editing. This allows users to make
+* changes to the state of individual fields without needing
+* to make repeated calls to the databse via API methods.
+*/
+export const CareForKidsForm: React.FC = (props) => {
+
+    const [currentState, setcurrentState] = useState(props.initState);
+
+    // Updates only the changed information by using the object spread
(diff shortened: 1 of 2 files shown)


client/src/containers/EditRecord/Forms/CareForKidsForm.tsx
```diff
@@ -0,0 +1,82 @@
...
+        console.log(currentState);
+        props.passData(currentState);
+        alert('Data saved successfully!');
+    }
+
+    return(
+        <div className='grid-container margin-top-2'>
+            <h2 className='grid-row'>Receiving Care For Kids?</h2>
+            <div>
+                <Form<object>
```
jlhogan, line 49: should be `Child` type param for the form

client/src/containers/EditRecord/Forms/CareForKidsForm.tsx
```diff
@@ -0,0 +1,82 @@
...
+* the Care For Kids field of a data record.
+* Data is not processed using a FlattenedEnrollment.
+* Rather, information is inherited via props from the parent
+* object (the EditRecord page), which will bundle all 
+* edits together and make a single unified record update
+* when the user is done editing. This allows users to make
+* changes to the state of individual fields without needing
+* to make repeated calls to the databse via API methods.
+*/
+export const CareForKidsForm: React.FC = (props) => {
```
jlhogan, line 21: spread props to avoid needing to use `props.[thing]` to access:
```
type CareForKidsFormProps = {
initState: Child;
passData: (_: Child) => void;
}

export const CareForKidsForm: React.FC<CareForKidsFormProps> = ({
  initState,
  passData
}) => {
...
}
```

client/src/containers/EditRecord/Forms/CareForKidsForm.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    const [currentState, setcurrentState] = useState(props.initState);
```
jlhogan, line 23: Don't think this copy of state is necessary. The form creates an internal state representation of the object, which is passed as an argument into the `onSubmit` function so you can get the user updated version of the child object in there

PR #218, opened 2020-08-25
Add CircleCI to Fawkes (in the form of a PR build step)

JordanGuinn:
This sets up a build workflow through [CircleCI](https://app.circleci.com/pipelines/github/ctoec) that's triggered for all Fawkes pull requests.  The workflow (called `build-and-test`) has three separate jobs defined within:

- install all dependencies
- build both client and server
- run all tests

**All three jobs need to pass** in order for a pull request to be mergeable.


.circleci/config.yml
```diff
@@ -0,0 +1,110 @@
...
+      - attach_workspace:
+          at: .
+      - run:
+          name: Install deployment dependencies
+          command: |
+            sudo apt-get -y -qq update
+            sudo apt-get install python3-pip python3-dev build-essential
+            sudo pip3 install awsebcli
+      - run: yarn run bundle
+      - run: ls -la && eb deploy ece-fawkes-staging-staging --staged
```
jlhogan, line 62: `ls -la` is maybe left over debug?

.circleci/config.yml
```diff
@@ -0,0 +1,110 @@
...
+          requires:
+            - build
+      - deploy-staging:
+          requires:
+            - test
+          filters:
+            branches:
+              only: base
+
+  build-and-deploy-qa:
```
jlhogan, line 98: is `build-and-deploy-qa` something we can trigger manually ? :heart_eyes_cat:  and then `build-and-test` occurs on PR event?

.circleci/config.yml
JordanGuinn: Ahhhh yes - deleting now

.circleci/config.yml
JordanGuinn: Yep!  The `build-and-test` workflow is triggered by all PR events, but it ALSO has a secret `deploy-staging` job that's implicitly triggered on all events against our default branch (`base`), effectively doing what our staging pipeline currently is, just without all the AWS infrastructure in Terraform. 🎉

CircleCI's option for manual triggers is to set up a workflow that has an explicit approval step - so `build-and-deploy-qa` is actually triggered regularly, it just requires somebody to push a button in order to complete the deploy.  Which feels... _weird_, but overall fine for the number of envs we have rn.

JordanGuinn: I'm also gonna throw together a teeny shell script for direct manual deploys, JUST so we can have a more convenient option to do so direct from the CLI if we need to.

PR #221, opened 2020-08-26
Adds /sites GET route

jlhogan:
Also:
- adds QS parsing util
- adds some explicit Id props to entities to enable easier checking
without unnecessary object loading
- makes org -> community have an FK
- pulls children logic into controller file
- trims down the built resources with change to tsconfig

### DIFF
--- a/src/routes/sites.ts
+++ b/src/routes/sites.ts
@@ -1,0 +1,44 @@
+import express, { Response, Request } from 'express';
+import { passAsyncError } from '../middleware/error/passAsyncError';
+import { parseQS } from '../utils/parseQS';
+import * as controller from '../controllers/sites';
+
+export const sitesRouter = express.Router();
+
+/**
+ * /sites GET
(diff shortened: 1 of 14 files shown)


src/routes/sites.ts
```diff
@@ -0,0 +1,44 @@
...
+    const organizationIds = parseQS(req, 'organizationId', {
+      post: parseInt,
+      forceArray: true,
+    }) as number[];
+    const communityIds = parseQS(req, 'communityId', {
+      post: parseInt,
+      forceArray: true,
+    }) as number[];
+
+    console.log(siteIds);
```
mmazanec22, line 33: Console log

src/utils/parseQS.ts
```diff
@@ -0,0 +1,25 @@
+import { Request } from 'express';
+
+export const parseQS = <T = string>(
```
mmazanec22, line 3: This function name would be clearer if it was spelled out

PR #226, opened 2020-08-26
Family form with backend support

### LINKED ISSUE #166: add family info form
**AC**:
- adds forms to edit family info to the family info tab
- forms use "Family" as root form data type
- on submit, form PUTs to a new families/:familyId PUT endpoint, which updates the family object in the database

Screen Shot 2020-08-17 at 3 33 50 PM

bamader:
Closes #166


client/src/containers/EditRecord/Forms/CareForKidsForm.tsx
```diff
@@ -7,7 +7,7 @@ import {
   RadioButtonGroup,
   RadioButton,
 } from '@ctoec/component-library';
-import { Child } from '../../../shared/models';
+import { Child } from 'shared/models';
```
jlhogan, line 10: this is wrong, and you prob just havent cleaned out the `client/node_modules/shared` which is no longer a dep defined in package.json

client/src/containers/EditRecord/Forms/FormFields/AddressFieldset.tsx
```diff
@@ -0,0 +1,131 @@
...
+  TextInputProps,
+  TextInput,
+  SelectProps,
+  Select,
+} from '@ctoec/component-library';
+import { Family } from 'shared/models';
+
+// Listing of all possible US states to serve as options
+// in the address selector
+const possibleStates = [
```
jlhogan, line 14: it actually only needs to be 4 states: CT, NY, MA, RI
students must live in the those states to be eligible to attend state funded CT ECE programs

src/routes/families.ts
```diff
@@ -0,0 +1,41 @@
...
+export const familyRouter = express.Router();
+
+/*
+ * /families/:familyId PUT
+ */
+familyRouter.put(
+  '/:familyId',
+  passAsyncError(async (req: Request, res: Response) => {
+    try {
+      const recordId = req.body['recordId'];
```
jlhogan, line 16: you shouldn't need any concept of the child in here. This should be able to be a vanilla PUT where you just pass in the new object state, merge it with the family you pull from the DB (by family id), and then save that back to the db. There's prob a typeORM function for this... perhaps `preload` ? https://github.com/typeorm/typeorm/blob/master/docs/entity-manager-api.md

client/src/containers/EditRecord/Forms/FormFields/AddressFieldset.tsx
bamader: Great, that makes it way cleaner

jlhogan: also nit: I like ALL_CAPs snake case names for consts :)

src/routes/families.ts
bamader: Are family ID's unique in the database? My understanding was that more than one child could potentially be associated to the same family. I was worried this might generate edge cases, such as if there are two kids associated to the same family, one currently receiving funding and the other not, and the kids are listed with different addresses (maybe one of them stopped being enrolled while the family lived at a different address). Am I just being way too paranoid?

client/src/containers/EditRecord/Forms/CareForKidsForm.tsx
bamader: Got it, yep it was not cleaning out the `client/node_modules` folder

client/src/containers/EditRecord/Forms/FormFields/AddressFieldset.tsx
bamader: I'm a big snake_case guy anyway; for some reason camelCase just feels more java-y? idk.

bamader:
Re: discussion on different styles. I'm wondering if what's happening here might actually make more sense in the long run. Right now, EditRecord is the page that calls for a fetch on the child record being edited. This happens when the user clicks on a name the first time (which makes a lot of sense to me). But then, because the forms on the individual TabNavs get the record passed to them as an initial state, they don't have to do any fetching with the API. To me, it at least makes a lot of intuitive sense to not make them do that after an update when it seems simple to have them just sort of patchwork update their pieces. At least, that's how I intuitively process what's happening. Thoughts?

src/routes/families.ts
bamader: Yeah okay I think I was going overboard. I swapped out with the `preload` and `save` you recommended and everything's working perfectly fine. Now I pass back a 200 with the familyId too, and I use that to update the local state.

jlhogan:
I dont see patched partial object updates as simpler than re-triggering that initial child fetch in the parent component. while it seems simple here, more complex updates could happen. instead of requiring the client to correctly update local state, just grab the thing from the server again and it will always be exactly what is stored in the db! we don't need to over optimize here to avoid the network request -- the payload is small potatoes

bamader:
Got it. I do have a tendency to over-optimize, and if that's not a big deal then simplicity for the win! I'll wait to check out your PR  and then modify as needed!

src/routes/families.ts
jlhogan: I realized that we probably should not use preload here actually, because this endpoint should only update an existing resource NOT create a new one if it did not exists. So i think we actually want to: 
1. look up the entity by id, and return 404 if it does not exist
2. update the entity (probably using merge b/c i think it's fine n reasonable to support partial updates regardless of PATCH w/e)
3. save updated entity to DB

bamader: Got it. Was just reading over the merge docs and thought that a single save call would handle things for us. Will go ahead with this!

bamader: Actually wait the docs say that `save` will still insert an entity if it isn't already in the DB

bamader: I think `update` is actually the command we want? It's just a strict partial update that _won't_ create new entities in the db if the thing doesn't already exist.

jlhogan: oh sorry i meant to say `update` in there ha

client/src/containers/EditRecord/Forms/FamilyInfoForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+} from '@ctoec/component-library';
+import { Family } from '../../../shared/models';
+import { AddressFieldset } from './FormFields/AddressFieldset';
+import AuthenticationContext from '../../../contexts/AuthenticationContext/AuthenticationContext';
+import { apiPut } from '../../../utils/api';
+
+/*
+Simple props type to hold the state passed in from the edit
+record page and the state change function used to modify
+the root variable. Note that we need to use the id of the
```
jlhogan, line 17: nit: outdate comment

client/src/containers/EditRecord/Forms/FamilyInfoForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+parent child object because it is unique in the backend
+database, so this allows updating an existing record rather
+than creating a new one or searching unsuccessfully.
+*/
+export type FamilyFormProps = {
+  initState: Family;
```
jlhogan, line 23: would be fine with calling this `family` ¯\_(ツ)_/¯ i feel like `initState` doesn't really confer a lot of important meaning and is potentially confusing

client/src/containers/EditRecord/Forms/FamilyInfoForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+}) => {
+  const { accessToken } = useContext(AuthenticationContext);
+  const [saving, setSaving] = useState(false);
+
+  // Simple wrapper method that can be invoked via arrow function
+  // in the callback series of .thens while handling the API
+  // request. If the PUT update comes back with the family's
+  // correct ID number, that means we updated the DB successfully.
+  function responseWrapper(code: number) {
+    if (code == initState.id) {
```
jlhogan, line 46: I don't think this is necessary -- any errors around the child fetch can be dealt with by the code where that api call is defined.

also i'm having PUT return 200 OK with no body (`res.sendStatus(200)`); thoughts on what we'd prefer? Since we're relying on the GET to give us object state anyway I dont see the need for any return data

client/src/containers/EditRecord/Forms/FormFields/AddressFieldset.tsx
```diff
@@ -0,0 +1,80 @@
...
+      className="display-inline-block"
+    >
+      <div className="grid-col-12 margin-top-2">
+        <FormField<Family, TextInputProps, string>
+          getValue={(data) => data.at('streetAddress')}
+          type="input"
+          inputComponent={TextInput}
+          id="street-address"
+          label="Street address"
+          parseOnChangeEvent={(e) => {
```
jlhogan, line 31: my preference is the oneliner: `(e) => e.target.value`

jlhogan: for all `parseOnChangeEvent` in this file :)

src/routes/families.ts
```diff
@@ -0,0 +1,26 @@
...
+  '/:familyId',
+  passAsyncError(async (req: Request, res: Response) => {
+    const famId = req.params['familyId'];
+    try {
+      const newFam = req.body;
+      await getManager().update(Family, { id: famId }, newFam);
+      res.status(200).json(newFam.id);
+    } catch (err) {
+      console.log('Error saving changes to family: ', err);
+      res.status(400).json('Could not update family address information');
```
jlhogan, line 22: don't need to do this here... see https://github.com/ctoec/data-collection/blob/base/src/middleware/error/handleError.ts 
`handleError` middleware will send the error response based on the thrown error (BadRequestError extends ApiError, which has a status prop that will be used as the status. The message will be returned as "error" prop in response json)

src/routes/families.ts
```diff
@@ -0,0 +1,26 @@
...
+ * /families/:familyId PUT
+ */
+familyRouter.put(
+  '/:familyId',
+  passAsyncError(async (req: Request, res: Response) => {
+    const famId = req.params['familyId'];
+    try {
+      const newFam = req.body;
+      await getManager().update(Family, { id: famId }, newFam);
+      res.status(200).json(newFam.id);
```
jlhogan, line 19: my pref is `res.sendStatus(200)`, no response body

client/src/containers/EditRecord/Forms/FamilyInfoForm.tsx
bamader: If we're doing error checking and handling in the api call then I'm totally fine axing this. Actually, now that I think about it, it was almost totally an artifact of when I was trying to do local state patching. Because since we trigger a refetch every edit now anyway, we're always getting the current copy of the record.

Also I'm fine with put returning a 200 okay with no message. Again, the consistent refetch really takes care of any need to send any other data.

client/src/containers/EditRecord/Forms/FamilyInfoForm.tsx
bamader: Ope yeah that's def a holdover from my initial attempt at using state passing. Totally unnecessary now that we're doing refetch.

PR #227, opened 2020-08-26
Add backend service to change enrollment for a child

jlhogan:


client/src/shared/payloads/ChangeEnrollment.ts
```diff
@@ -0,0 +1,18 @@
+import { Moment } from 'moment';
```
bamader, line 1: General architecture question: what kinds of things will be putting in the payloads directory?

src/controllers/children.ts
```diff
@@ -0,0 +1,105 @@
...
+export const changeEnrollment = async (
+  id: string,
+  changeEnrollmentData: ChangeEnrollment
+) => {
+  const child = await getChildById(id);
+
+  if (!child) return;
+
+  // Get current enrollment
+  return getManager().transaction(async (tManager) => {
```
bamader, line 29: Does the same logic of updating the family information apply here? As in, could we use `preload` and then `save` to retrieve the object and then merge the updated version back into the db?

jlhogan: see https://github.com/ctoec/data-collection/pull/226#discussion_r478579540 i think the easiest way is to just use `getManager().update()` -- the transaction stuff is necessary here because a bunch of separate SQL commands need to all be executed for this to be successful. If any one fails I want to rollback anything else that has happened in the process. Updating family esp using `update` will be an atomic DB command so no need for transaction management

client/src/shared/payloads/ChangeEnrollment.ts
jlhogan: things like this -- interfaces to define expected PUT/POST request payloads that are not db entities but are created by client and ingested by server

jlhogan: another example of what could end up here: if we made a specific `enrollment/:enrollmentId/withdraw` endpoint, to enable us to enforce on the server side that `exit` and `exitReason` are required to do this instead of having to put that logic in the controller then a payload here would be something like:
```
{
  exit: Moment;
  exitReason: ExitReason
}
```

PR #229, opened 2020-08-27
does enrollment-funding stuff

jlhogan:


client/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@ctoec/component-library": "https://github.com/ctoec/component-library#base",
+    "@ctoec/component-library": "https://github.com/ctoec/component-library#update-day-picker",
```
mmazanec22, line 6: This is merged and can be reverted back to base :)

client/src/containers/EditRecord/EditRecord.tsx
```diff
@@ -6,17 +6,23 @@ import AuthenticationContext from '../../contexts/AuthenticationContext/Authenti
 import { apiGet } from '../../utils/api';
 import { Child } from '../../shared/models';
 import { CareForKidsForm } from './Forms/CareForKidsForm';
+import { EnrollmentFundingForm } from './EnrollmentFunding/Form';
 
 const EditRecord: React.FC = () => {
   const { childId } = useParams();
   const { accessToken } = useContext(AuthenticationContext);
   const [rowData, setRowData] = useState<Child>();
+  const [refetch, setRefetch] = useState<number>(0);
```
mmazanec22, line 15: Shouldn't setting the row data make this component re-render anyway?  Could we just use a useEffect with rowData as a dep to fire the get?

jlhogan: see convo on https://github.com/ctoec/data-collection/pull/226; i personally think it's cleaner to avoid having the forms do patch updates to the rowData state var. This isn't to force a re-render, but to force a refetch of the child object from the backend, so that we don't have to do any client-side object tree maintenance

client/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@ctoec/component-library": "https://github.com/ctoec/component-library#base",
+    "@ctoec/component-library": "https://github.com/ctoec/component-library#date-input-default-value",
```
jlhogan, line 6: to update before merge once component-lib is updated

src/controllers/enrollmentReports/parse.ts
```diff
@@ -101,6 +101,7 @@ function parseSheet(sheet: WorkSheet, objectProperties: string[]) {
   });
 
   const sheetType = getSheetType(sheet);
+  console.log('sheet type', sheetType);
```
mmazanec22, line 104: Console logs

PR #247, opened 2020-08-31
232, 217: Improves excel parsing for dates and zipcodes

### LINKED ISSUE #217: Zip codes starting with '0' are parsed as dates
In uploaded excel spreadsheets, zip codes beginning with a 0 (as many state of CT zip codes do) are treated as four digit numbers when excel turns them into strings (it throws away the 0). This then gets parsed as a year by the Moment check in parseUploadedTemplate, and so a UTC timestamp is stored in place of the zip code in the family object.

### LINKED ISSUE #232: Not all dates are parsed as Dates
**problem**:
sheetJS automatically parses excel date cells into javascript Dates. However not all valid date formats are recognized by all excel editors as dates; additionally csvs will not have date type cells. As a result some properties that should be Moments are not parsed correctly and remain strings

**AC**
- update `parseFlattenedEnrollment()` in `controllers/enrollmentReports/parse` to more robustly parse dates
- instead of just checking if a value is an instance of Date, the function should accept an array of properties that are defined as Dates in the entity model (as it does for Boolean fields) and use that to determine which fields should be moment-ified

jlhogan:
closes #232 closes #217

### DIFF
--- a/src/entity/FlattenedEnrollment.ts
+++ b/src/entity/FlattenedEnrollment.ts
@@ -1,6 +1,13 @@
 import { Entity, Column, PrimaryGeneratedColumn, ManyToOne } from 'typeorm';
 
-import { FlattenedEnrollment as FlattenedEnrollmentInterface } from '../../client/src/shared/models';
+import {
+  FlattenedEnrollment as FlattenedEnrollmentInterface,
+  Gender,
+  SpecialEducationServicesType,
+  AgeGroup,
+  FundingSource,
+  FundingTime,
+} from '../../client/src/shared/models';
 
 import { EnrollmentReport } from './EnrollmentReport';
 import {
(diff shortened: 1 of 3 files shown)


jlhogan:
@mmazanec22 this should remove the need for the diff in https://github.com/ctoec/data-collection/pull/202/files#diff-7ea4ac5cb52c2bdd0cc14d09456d723d

src/entity/FlattenedEnrollment.ts
```diff
@@ -195,7 +202,7 @@ export class FlattenedEnrollment implements FlattenedEnrollmentInterface {
     definition:
       "The child's gender, as identified by the family (not as it appears on the birth certificate).",
     reason: 'Allows linking to SASID-backed data',
-    format: 'Male, Female, Nonbinary, Unspecified',
+    format: Object.values(Gender).join(', '),
```
mmazanec22, line 205: Nice

PR #253, opened 2020-08-31
Add back button

### LINKED ISSUE #246: No back button on the child record page
Screen Shot 2020-08-31 at 11 04 59 AM
Back button should go back to the roster, right now it is moving through all the child data sections.

mmazanec22:
Closes #246

### DIFF
--- a/client/src/components/BackButton.tsx
+++ b/client/src/components/BackButton.tsx
@@ -1,0 +1,27 @@
+import React from 'react';
+import { Button, TextWithIcon } from '@ctoec/component-library';
+import { ReactComponent as Arrow } from '@ctoec/component-library/dist/assets/images/arrowRight.svg';
+import { useHistory } from 'react-router-dom';
+
+type BackButtonParams = {
+  text?: string;
+}
+
+export default ({ text = 'Back' }: BackButtonParams) => {
+
+  const { goBack } = useHistory();
+
+  return (<Button
+    className="margin-bottom-2 text-bold"
+    appearance="unstyled"
+    text={
+      <TextWithIcon
+        text={text}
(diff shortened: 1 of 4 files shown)


client/src/components/BackButton.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { Button, TextWithIcon } from '@ctoec/component-library';
+import { ReactComponent as Arrow } from '@ctoec/component-library/dist/assets/images/arrowRight.svg';
+import { useHistory } from 'react-router-dom';
+
+type BackButtonParams = {
+  text?: string;
+};
+
+export default ({ text = 'Back' }: BackButtonParams) => {
+  const { goBack } = useHistory();
```
jlhogan, line 11: :clap:

client/src/components/BackButton.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { Button, TextWithIcon } from '@ctoec/component-library';
+import { ReactComponent as Arrow } from '@ctoec/component-library/dist/assets/images/arrowRight.svg';
+import { useHistory } from 'react-router-dom';
+
+type BackButtonParams = {
+  text?: string;
+};
+
+export default ({ text = 'Back' }: BackButtonParams) => {
```
jlhogan, line 10: for some reason I always do named exports. Is there a reason to export default here? should we pick something and stick to it?

mmazanec22: I kinda prefer named exports too because it's easier to gather them together into an index file later.  This was just laziness basically.  If we want to officially go with named exports, I'm in.

PR #254, opened 2020-08-31
222 verify site

### LINKED ISSUE #222: Site is not displayed on verify table

mmazanec22:
Closes #222

### DIFF
--- a/client/src/containers/CheckData/TableColumns.tsx
+++ b/client/src/containers/CheckData/TableColumns.tsx
@@ -53,9 +53,7 @@ export const tableColumns: Column<Child>[] = [
     className: tableColumnClassName,
     name: 'Contract space',
     cell: ({ row }) => (
-      <td>
-        {idx(row, (_) => _.enrollments[0].fundings[0].fundingSpace.time)}
-      </td>
+      <td>{idx(row, (_) => _.enrollments[0].fundings[0].fundingSpace.time)}</td>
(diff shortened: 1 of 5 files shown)


client/src/containers/CheckData/TableColumns.tsx
```diff
@@ -5,71 +5,67 @@ import { Column, Button } from '@ctoec/component-library';
...
-        <td>
-          {idx(row, (_) => _.enrollments[0].fundings[0].fundingSpace.source)}
-        </td>
-      ),
-    },
-    {
-      className: tableColumnClassName,
-      name: 'Contract space',
-      cell: ({ row }) => (
+export const tableColumns: Column<Child>[] = [
```
mmazanec22, line 8: I couldn't see a reason why this would be a function-- will change back if needed

jlhogan: oh no that's a hold over from when it used to take in "reportId"! this should be all good

src/controllers/enrollmentReports/map.ts
```diff
@@ -214,6 +214,7 @@ const mapEnrollment = (
 
   const enrollment = getManager().create(Enrollment, {
     siteId: site.id,
+    site,
```
jlhogan, line 217: nit: should be able to get rid of `siteId` with this

PR #260, opened 2020-09-02
Income determination

### LINKED ISSUE #163: Add care 4 kids form
Separate this from enrollment/funding

Have one form element with Y/N

### LINKED ISSUE #165: add family income determination form

bamader:
Closes #165 
The commits are funky cuz I had to rebase to get the updated copy of the dateInput and TabNav. The yarn.lock file is modified here to reflect fetching the updated component library. Otherwise, this works at persisting changes and redeterminations to the DB.

UPDATE: Also added in the API call for Care4Kids using Melanie's backend route.
Closes #163


client/src/containers/EditRecord/EditRecord.tsx
```diff
@@ -72,7 +73,17 @@ const EditRecord: React.FC = () => {
           {
             id: TAB_IDS.INCOME,
             text: 'Family Income',
-            content: <span>This is where the family income form goes</span>,
+            content: (
+              <UpdateForm
+                familyId={rowData.family.id}
+                determinations={
```
mmazanec22, line 79: For consistency with the other forms, should we name this form something like "FamilyIncome" and have the values passed to it be refetchChild and the family object (rather than family id or determinations)?

client/src/containers/EditRecord/Forms/FamilyIncome/EditDeterminationForm.tsx
```diff
@@ -0,0 +1,92 @@
...
+  isNew,
+  refetchChild,
+}) => {
+  // Set up form state
+  const { accessToken } = useContext(AuthenticationContext);
+  const [closeCard, setCloseCard] = useState(false);
+  const [loading, setLoading] = useState(false);
+
+  useEffect(() => {
+    if (closeCard) setCloseCard(false);
```
mmazanec22, line 40: Would you mind leaving a note here about what this is for?  It took me a min to follow why we needed this

client/src/containers/EditRecord/Forms/FamilyIncome/UpdateForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * The main form rendered in the EditRecord TabNav that allows a user
+ * to update the income determination for a given Child record
+ * object. Updates are performed on individual income determinations
+ * before being re-persisted to the database, but these updates are
+ * passed off to the accessible forms (Edit and Redetermine) reached
+ * through this page.
+ */
+export const UpdateForm: React.FC<IncomeFormProps> = ({
```
mmazanec22, line 16: Same comment here as on the edit component

client/src/containers/EditRecord/EditRecord.tsx
bamader: I can (and will) def rename the form! As far as what we pass in, I think we actually wanted to just use the income determinations array owned by the family object as the root data source, rather than the family object itself. The idea (if I remember right) was to give each form the lowest "level" of object possible in the hierarchy that still contained all of the information it needed. For most of our other forms, this is a child or a family. But the determinations form doesn't need to know about things like a street address.

client/src/containers/EditRecord/Forms/FamilyIncome/EditDeterminationForm.tsx
bamader: Yes I'll put a comment in the file!

client/src/containers/EditRecord/EditRecord.tsx
mmazanec22: Since we're also passing in the familyId and the relevant endpoints are nested under families, I'd argue that passing around family for consistency's sake makes sense.  But then I guess that begs the question: do we need to be passing around familyId at all, and should the redeterminations be nested under family?

bamader: That's a fair point. I'm not sure what the best way to handle the object separation is. We do need the familyId unless we totally separate determinations from the family (since right now the ID is how we index into the DB in the getManager().findOne call), but I think that's what the second part is of what you're saying? @jlhogan Thoughts? (Since you wrote the AC)

jlhogan: I like `income-determinations` nested under `families` because it makes sense to me that creating a new income determination would be done at `families/:familyId/income-determinations`. Pretty sure the familyId is not actually necessary for doing any DB lookup, since each incomeDetermination has a distinct unique id.  

whether you pass in the family or just the id and determinations i have no opinion on

client/src/containers/EditRecord/EditRecord.tsx
```diff
@@ -72,7 +73,17 @@ const EditRecord: React.FC = () => {
           {
             id: TAB_IDS.INCOME,
             text: 'Family Income',
-            content: <span>This is where the family income form goes</span>,
+            content: (
+              <FamilyIncomeForm
+                familyId={rowData.family.id}
+                determinations={
+                  rowData.family.incomeDeterminations
```
jlhogan, line 80: `(rowData.family.incomeDeterminations || [])`

src/routes/families.ts
```diff
@@ -27,7 +27,62 @@ familyRouter.put(
...
+      throw new BadRequestError('Family not saved');
+    }
+  })
+);
+
+/**
+ * /families/:familyId/incomeDetermination/:determinationId PUT
+ */
+familyRouter.put(
+  '/:familyId/incomeDetermination/:determinationId',
```
jlhogan, line 39: `income-determinations`

src/routes/families.ts
```diff
@@ -27,7 +27,62 @@ familyRouter.put(
...
+    } catch (err) {
+      if (err instanceof ApiError) throw err;
+      console.log('Error saving changes to income determination: ', err);
+      throw new BadRequestError('Edited determination not saved');
+    }
+  })
+);
+
+familyRouter.post(
+  '/:familyId/incomeDetermination',
```
jlhogan, line 68: `income-determinations`

src/routes/families.ts
```diff
@@ -27,7 +27,62 @@ familyRouter.put(
...
+  passAsyncError(async (req: Request, res: Response) => {
+    try {
+      const famId = parseInt(req.params['familyId']);
+      const rootFam = await getManager().findOne(Family, { id: famId });
+      const determination = getManager().create(IncomeDetermination, {
+        id: req.body.id,
```
jlhogan, line 74: creating a new resource should not use a user-supplied id

src/routes/families.ts
```diff
@@ -27,7 +27,62 @@ familyRouter.put(
...
+  })
+);
+
+familyRouter.post(
+  '/:familyId/incomeDetermination',
+  passAsyncError(async (req: Request, res: Response) => {
+    try {
+      const famId = parseInt(req.params['familyId']);
+      const rootFam = await getManager().findOne(Family, { id: famId });
+      const determination = getManager().create(IncomeDetermination, {
```
jlhogan, line 73: this can just be
```
 const determination = getManager().create(IncomeDetermination, {...req.body, family })
```

client/src/containers/EditRecord/Forms/FamilyIncome/Fields/DeterminationCard.tsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
```
jlhogan, line 1: don't think this is really a "Field"

client/src/containers/EditRecord/Forms/FamilyIncome/Fields/Common.ts
```diff
@@ -0,0 +1,12 @@
+import { IncomeDetermination } from '../../../../../shared/models';
+
+/**
+ * Generic type to hold the props of the high-level income forms,
+ * specifically to allow indexing into the desire determination as
+ * well as persisting back to the DB.
+ */
+export type IncomeFormProps = {
```
jlhogan, line 8: this type is only used in one component, i think it should be defined in that file not exported as a type from it's own place

jlhogan (approved):
LGTM! one note: I don't think we need comments explaining how api interactions work. let's try to reserve lengthy comments for things that are unexpected / otuside of common patterns we're repeating. No real harm but I think by keeping inline comments to only those things that really require explanation, we make the code easier to understand and learn about

src/routes/families.ts
bamader: You're absolutely right. This was a holdover from before I realized that I was working with an out-of-date base so the dates were messing with the DB. Fixing!

client/src/containers/EditRecord/Forms/FamilyIncome/Fields/Common.ts
bamader: Good catch--another holdover from before I refactored. Will move accordingly!

PR #263, opened 2020-09-03
164 funding

### LINKED ISSUE #164: add enrollment/funding form
**AC**
- forms use "Enrollment" and "Funding" as root form data type
 - on enrollment edit submit, form PUTs to new `enrollments/:enrollmentId` endpoint to update the resource
- on enrollment creation submit, form POSTS to new `enrollment` endpoint to create the resource
- on funding edit submit, form PUTs to new `enrollments/:enrollmentId/fundings/:fundingId` PUT endpoint to update the resource
- on funding create submit, form POSTS to new `enrollments/:enrollmentId/fundings` endpoint to create the resource

** OPEN QUESTIONS** 
- When creating a new Enrollment with a new funding, does that happen in a single POST to `enrollments`, or as a follow up action to the enrollment create that is a seprate funding creation?
- Should these routes actually be nested under `child/:childId` ? Does it matter? 

Screen Shot 2020-08-17 at 3 30 23 PM

jlhogan:
closes #164 adding all funding functionality 

a chunk of this diff is not actually of interest, just moved the childInfo stuff to align with other sections 

TODO: Add colors for funding tags!!!!!


client/src/containers/EditRecord/Forms/EnrollmentFunding/Fields/Funding/ContractSpace/common.ts
```diff
@@ -0,0 +1,24 @@
+import {
+  FundingSpace,
+  FundingTime,
+  FundingTimeSplit,
+} from '../../../../../../../shared/models';
+
+export const getContractSpaceString = (fundingSpace: FundingSpace) => {
```
mmazanec22, line 7: Should this file maybe be utils rather than common?  I feel like utils is where I would look for something like this, whereas common is less of a thing?  (Or maybe it 's just a convention I'm not familiar with!)

client/src/containers/EditRecord/Forms/EnrollmentFunding/Fields/Funding/NewFunding.tsx
```diff
@@ -0,0 +1,112 @@
...
+  }, [enrollment, fundingSpaces]);
+
+  return (
+    <RadioButtonGroup
+      id="funding-source"
+      name="funding-source"
+      legend="Funding source options"
+      options={[
+        {
+          render: (props: RadioOptionRenderProps) => (
```
mmazanec22, line 80: Not for this PR but: what would you think about simplifying this so we don't have to tell the radio button group that it renders radio buttons?  I can't think of a use case where it would be a good idea to render anything but radio buttons here

client/src/utils/api.ts
```diff
@@ -63,7 +63,10 @@ async function api(
     headers['Authorization'] = `Bearer ${opts.accessToken}`;
   }
 
+  console.log('opts', opts);
```
mmazanec22, line 66: Console logs

mmazanec22 (approved):

client/src/containers/EditRecord/Forms/EnrollmentFunding/Fields/Funding/ContractSpace/common.ts
jlhogan: it should im just le lazy. moving!

PR #266, opened 2020-09-03
131 withdraw

### LINKED ISSUE #131: Withdraw: I can withdraw one record in my roster.
Withdraw

Functionality is the same as ECE reporter. After submitting the withdrawal, a confirmation alert should appear on the verify data table with the text of "[Child name] has been withdrawn."

jlhogan:
closes #131


client/src/containers/EditRecord/Forms/FamilyIncome/Fields/IncomeDeterminationFieldSet.tsx
```diff
@@ -14,8 +14,7 @@ import { IncomeDetermination } from '../../../../../shared/models';
  * copy of the ID of the determination being edited.
  */
 type IncomeDeterminationFieldSetProps = {
-  type: 'new' | 'redetermine' | 'edit';
```
bamader, line 17: I'm guessing we delete 'new' here because Melanie is handling the generation of a new added enrollment in the create flow?

client/src/containers/EditRecord/Forms/FamilyIncome/EditDeterminationForm.tsx
```diff
@@ -78,10 +78,7 @@ export const EditDeterminationForm: React.FC<EditDeterminationFormProps> = ({
           onSubmit={(data) => onFormSubmit(data)}
           className="usa-form"
         >
-          <IncomeDeterminationFieldSet
-            type={'edit'}
-            determinationId={determination.id}
```
bamader, line 83: Why are we removing the determinationId? Even if it doesn't help with the way the form works, doesn't it provide accessibility information for screen readers in an elementId below?

client/src/containers/EditRecord/Forms/FamilyIncome/Fields/IncomeDeterminationFieldSet.tsx
```diff
@@ -37,16 +35,11 @@ export const IncomeDeterminationFieldSet: React.FC<IncomeDeterminationFieldSetPr
       break;
 
     case 'edit':
-      elementId = `family-income-determination-edit-${determinationId}`;
```
bamader, line 40: This is the line I meant about the determinationId improving accessibility

client/src/containers/EditRecord/Forms/FamilyIncome/RedeterminationForm.tsx
```diff
@@ -31,8 +31,6 @@ export const RedeterminationForm: React.FC<RedeterminationFormProps> = ({
   const [closeCard, setCloseCard] = useState(false);
   const [loading, setLoading] = useState(false);
 
-  var newDet = { id: 0, numberOfPeople: 0, income: 0 } as IncomeDetermination;
```
bamader, line 34: Ahhh now I see what you were saying about not wanting to provide artificial values to ID numbers when we could just be passing in blank objects typed as the data the form expects

bamader (approved):
LGTM! General question: what's the motivation for using the controllers module when working with some of the API stuff, as opposed to just sticking it in the respective router method (this is my ignorance of good backend dev practices showing)?

client/src/containers/EditRecord/Forms/FamilyIncome/Fields/IncomeDeterminationFieldSet.tsx
jlhogan: if melanie is going to want to use this during the enroll flow, can leave it. just seemed like unused kruft  and i always prefer to delete unused things

client/src/containers/EditRecord/Forms/FamilyIncome/EditDeterminationForm.tsx
jlhogan: not sure that an arbitrary id that doesn't mean anything to the user actually improves accessibility. It used tobe needed for the form fields to access the correct  income determination in the array on the big nested enrollment object we were using for updates, but again was just tidying up. maybe overly so

client/src/containers/EditRecord/Forms/FamilyIncome/Fields/IncomeDeterminationFieldSet.tsx
jlhogan: i dont think random int is more accessible; probably the name here should use the determination date if we're actually thinking about what makes sense to a user of the app

jlhogan:
> API stuff, as opposed to just sticking it in the respective router method (this is my ignorance of good backend dev practices showing)?

we started with just the routers and defined simple logic inline in those, but then have added controllers as logic becomes more complex. Generally everything should be happening in the controllers now, but IMO not worth explicitly rewriting all what alraedy exists -- new things should go into controllers, and if while you're there you're compelled to move other logic into controllers as well i'd say that's welcome rider

client/src/containers/EditRecord/Forms/FamilyIncome/EditDeterminationForm.tsx
mmazanec22: A unique ID is helpful for accessibility purposes when we use it to programmatically associate two elements-- like a label and an input, or a dropdown and the button that opens it.  So long as the fieldset itself handles that stuff somehow, it's ok to not pass in an ID here.

client/src/containers/EditRecord/Forms/FamilyIncome/Fields/IncomeDeterminationFieldSet.tsx
jlhogan: @mmazanec22 are you using `new` type in enroll code? i will remove this part of the diff if so!!

client/src/containers/EditRecord/Forms/FamilyIncome/Fields/IncomeDeterminationFieldSet.tsx
mmazanec22: @bamader I think so long as the ID is unique on the page, we're ok

bamader: Gotcha, that makes sense. Then no need for this.

mmazanec22 (approved):

client/src/containers/EditRecord/Forms/FamilyIncome/Fields/IncomeDeterminationFieldSet.tsx
mmazanec22: @jlhogan I don't think so but I can add it back if needed!

PR #285, opened 2020-09-04
Fixes family determination POST, plus a little styling trim

jlhogan:


src/routes/families.ts
```diff
@@ -69,11 +69,13 @@ familyRouter.post(
   passAsyncError(async (req: Request, res: Response) => {
     try {
       const famId = parseInt(req.params['familyId']);
-      const rootFam = await getManager().findOne(Family, { id: famId });
+      const family = await getManager().findOne(Family, { id: famId });
+
       const determination = getManager().create(IncomeDetermination, {
         ...req.body,
-        rootFam,
+        family,
```
jlhogan, line 76: this is the thing that was actually _broken_... in javascript if you add a prop to an object without an explicit name, it will get the name of the variable so 
```
const foobar = "foobar"
{ foobar }  --> { foobar:  "foobar"}
```
vs
```
const foobar = "foobar";
{ family: foobar } -> { family: "foobar"}
```

bamader: Ooooooh that's totally my bad. Did not know that. That's kinda sneaky.

bamader (approved):
LGTM!

PR #293, opened 2020-09-08
204 add enrollment

mmazanec22:


client/src/containers/AddChild/AddChild.tsx
```diff
@@ -0,0 +1,184 @@
...
+  useEffect(() => {
+    // On initial load, create child
+    if (child || childId || !accessToken) return;
+
+    if (!organization) {
+      throw Error('Cannot create child without organization');
+    }
+    apiPost(
+      'children',
+      { firstName: '', lastName: '', organization },
```
jlhogan, line 130: this seems weird to me, is it necessary? we need to create the child before the user has supplied any info?

client/src/containers/CheckData/CheckData.tsx
```diff
@@ -28,6 +28,10 @@ const CheckData: React.FC = () => {
     }
   }, [reportId, accessToken]);
 
+  const organization =
```
jlhogan, line 31: I think we'd want to get all the orgs from the authenticated user, and if more than one the user has to pick, otherwise just that org the user has access to  ?

client/src/containers/EditRecord/EditRecord.tsx
```diff
@@ -60,14 +61,8 @@ const EditRecord: React.FC = () => {
...
   // Get reporting periods (needed to update enrollments with fundings)
-  const [reportingPeriods, setReportingPeriods] = useState<ReportingPeriod[]>(
-    []
-  );
-  useEffect(() => {
-    apiGet('reporting-periods', { accessToken }).then((_reportingPeriods) =>
-      setReportingPeriods(_reportingPeriods || [])
-    );
-  }, [accessToken]);
+  // TODO: we should probably use context rather than making lots of network requests
```
jlhogan, line 64: well like -- we should have a data cache

client/src/containers/EditRecord/Forms/ChildInfo/Form.tsx
```diff
@@ -94,3 +91,5 @@ export const ChildInfoForm: React.FC<ChildInfoFormProps> = ({
     </Form>
   );
 };
+
+export default ChildInfoForm;
```
jlhogan, line 95: why?

client/src/containers/EditRecord/Forms/EnrollmentFunding/Fields/Funding/NewFunding.tsx
```diff
@@ -73,6 +75,7 @@ export const FundingField = <T extends ChangeFunding | ChangeEnrollment>({
       id="funding-source"
       name="funding-source"
       legend="Funding source options"
+      showLegend
```
jlhogan, line 78: this should always be true?

client/src/containers/EditRecord/Forms/FamilyIncome/Fields/DeterminationCard.tsx
```diff
@@ -8,8 +8,8 @@ import {
   ExpandCard,
   Pencil,
 } from '@ctoec/component-library';
-import { currencyFormatter } from '../../../../../utils/formatters';
 import { IncomeDetermination } from '../../../../../shared/models';
+import currencyFormatter from '../../../../../utils/formatters/currency';
```
jlhogan, line 12: why this change?

client/src/containers/EditRecord/Forms/FamilyIncome/Form.tsx
```diff
@@ -24,14 +13,18 @@ export type IncomeFormProps = {
...
+export const FamilyIncomeForm: React.FC<EditFormProps> = ({
+  child,
+  onSuccess,
 }) => {
   const [showRedeterminationForm, setShowRedeterminationForm] = useState(false);
   const [currentIsNew, setCurrentIsNew] = useState(false);
 
+  if (!child) return <></>;
+
+  const determinations = child.family.incomeDeterminations;
```
jlhogan, line 25: should this be `child.family?.incomeDeterminations` ? same with below.
or will family always exist at this point?

jlhogan (approved):

client/src/containers/AddChild/AddChild.tsx
mmazanec22: This seemed like the easiest way to reuse forms, so that child info doesn't need to know whether it is creating a child on save.  If we prefer to stick this in child info and give it a create new child param or something that works too, also open to other suggestions

jlhogan: ahhh i see, that makes sense. my only worry is that we create the data and then network error/user gets a phone call/etc and they abandon the data in this state, then it's very hard to display to the user in a way that makes sense -- it'd be a  totally empty row. but we can cross that bridge when/if that happens, and can always update to move PUT/POST decisioning in childinfo form if necessary.

client/src/containers/CheckData/CheckData.tsx
mmazanec22: @jlhogan since this is going to require some design decisions, ok to punt to a new ticket in backlog?

client/src/containers/EditRecord/Forms/FamilyIncome/Fields/DeterminationCard.tsx
mmazanec22: Got an error because it was being exported by default but needs to be named-- changing it to be a named export

client/src/containers/EditRecord/Forms/EnrollmentFunding/Fields/Funding/NewFunding.tsx
mmazanec22: I think because of our header/legend issues we are usually not showing it.  Resolving that should be combined with the other header fixes probably

client/src/containers/EditRecord/Forms/FamilyIncome/Form.tsx
mmazanec22: I think family is required for a child to be created?  And we're compensating for the potentially undefined state of dets below this so I'll just move that up here for clarity

PR #295, opened 2020-09-08
Use main element and skip nav

mmazanec22:
Added the main element at the app level and a skip nav

### DIFF
--- a/client/src/containers/App/App.tsx
+++ b/client/src/containers/App/App.tsx
@@ -13,6 +13,7 @@ const App: React.FC = () => {
 
   return (
     <div className="App">
+      <a className="usa-skipnav" href="#main-content">Skip to main content</a>
       <Header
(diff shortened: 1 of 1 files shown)


client/src/containers/App/App.tsx
```diff
@@ -13,6 +13,7 @@ const App: React.FC = () => {
 
   return (
     <div className="App">
+      <a className="usa-skipnav" href="#main-content">Skip to main content</a>
```
jlhogan, line 16: is this visible & should it be?

mmazanec22: USDS handles the visibility/not-- when it receives focus via the keyboard, it slides into view in the top left corner, and is invisible the rest of the time

jlhogan: cooL!!!

jlhogan (approved):

Issue #283, opened 2020-09-04
Family homelessness needs a section header


RyanHansz:
<img width="438" alt="Screen Shot 2020-09-04 at 12 16 59 PM" src="https://user-images.githubusercontent.com/63253539/92262056-89adb300-eea8-11ea-8db3-348188990cc6.png">

PR #296, opened 2020-09-08
Delete individual components (income dets, fundings, etc.)

### LINKED ISSUE #250: Ability to delete income det, enroll, funding

bamader:
Closes #250 #283 #280 #279 #284

### DIFF
--- a/client/src/containers/EditRecord/EditRecord.tsx
+++ b/client/src/containers/EditRecord/EditRecord.tsx
@@ -252,8 +252,18 @@ const EditRecord: React.FC = () => {
           },
           {
             id: TAB_IDS.FAMILY,
+<<<<<<< HEAD
             text: 'Family Info',
             content: <FamilyInfoForm {...commonFormProps} />,
+=======
+            text: 'Family Address',
+            content: (
+              <FamilyInfoForm
+                family={rowData.family}
+                refetchChild={refetchChild}
+              />
+            ),
+>>>>>>> Incorporate some style fixes
           },
           {
             id: TAB_IDS.INCOME,
--- a/client/src/containers/EditRecord/EditRecord.tsx
+++ b/client/src/containers/EditRecord/EditRecord.tsx
@@ -3,7 +3,16 @@ import { useParams, useLocation, useHistory } from 'react-router-dom';
 import { TabNav, Button } from '@ctoec/component-library';
 import AuthenticationContext from '../../contexts/AuthenticationContext/AuthenticationContext';
 import { apiGet, apiDelete } from '../../utils/api';
+<<<<<<< HEAD
 import { Child } from '../../shared/models';
+=======
+import {
+  Child,
+  Funding,
+  Enrollment,
+  ReportingPeriod,
+} from '../../shared/models';
+>>>>>>> Cascading entity deletions on records, components that ACTUALLY WORK
 import { BackButton } from '../../components/BackButton';
 import {
   FamilyIncomeForm,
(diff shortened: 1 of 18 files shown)


client/src/containers/EditRecord/EditRecord.tsx
```diff
@@ -124,59 +127,57 @@ const EditRecord: React.FC = () => {
...
-              </Modal>
             </>
           )}
+          <Button
+            appearance="unstyled"
+            onClick={toggleDeleteModal}
+            text="Delete record"
+            className="margin-right-0"
+          />
+          <Modal
```
jlhogan, line 138: I would opt to pull this modal out as a separate delete form, like withdraw just to keep this file smaller

client/src/containers/EditRecord/Forms/EnrollmentFunding/Form/ChangeFundingForm.tsx
```diff
@@ -111,7 +109,7 @@ export const ChangeFundingForm: React.FC<ChangeFundingFormProps> = ({
                 onClick={() => setVisibleForm(undefined)}
               />
               <FormSubmitButton
-                text={loading ? 'Changing funding...' : 'Change funding'}
+                text={loading ? 'Ending funding...' : 'End current funding'}
```
jlhogan, line 112: this should probably still be `Change funding`, right? @RyanHansz (copy on "save" button for change or end current funding

src/entity/Enrollment.ts
```diff
@@ -47,7 +47,9 @@ export class Enrollment implements EnrollmentInterface {
   @Column({ nullable: true })
   exitReason?: string;
 
-  @OneToMany(() => Funding, (funding) => funding.enrollment)
+  @OneToMany(() => Funding, (funding) => funding.enrollment, {
+    onDelete: 'CASCADE',
```
jlhogan, line 51: do these need to be defined on both sides of the relationship? If not, i think it would be more clear to only define them on the owning entity (as is happening here for enrollment -> fundings) as opposed to on the child entity (as is happening above for enrollment -> child)

if it _is_ nec on both sides, then ignore this :) !

jlhogan (approved):

client/src/containers/EditRecord/EditRecord.tsx
bamader: Good call, can do!

client/src/containers/EditRecord/Forms/EnrollmentFunding/Form/ChangeFundingForm.tsx
bamader: That change is made from one of the linked issues up top (280 I think)--I think the rationale is that a user has to hit the End Funding button to get to this form, so a button that confirms that choice by saying Change Funding might seem off?

src/entity/Enrollment.ts
bamader: So I went back and forth on this for a while. Some GitHub issues page _seem_ to suggest it only needs to be on the side of the relationship with the ManyToOne holding, but there are tons of posters out there who say it only works for them when you put the onDelete cascade on both sides of the relationship. I ended up going with both sides just because it seemed safer and was a very minor addition.

client/src/containers/EditRecord/Forms/EnrollmentFunding/Form/ChangeFundingForm.tsx
jlhogan: well the button will either say "change funding" or "end funding" -- if the goal is parity with the button the user clicked to get here, then this needs to be dependent on the value of `visibleForm`, same as the header text

bamader: Oh that's true. I'll leave this for now then and just make whichever change is needed after Ryan weighs in

PR #297, opened 2020-09-08
First pass of basic FE test coverage

jlhogan:
adds snapshot and axe accessibility checks to all containers


client/package.json
```diff
@@ -3,10 +3,11 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@ctoec/component-library": "https://github.com/ctoec/component-library#base",
+    "@ctoec/component-library": "https://github.com/ctoec/component-library#aria-required",
```
jlhogan, line 6: TODO: Update to `base` when feature branch lands

client/package.json
```diff
@@ -42,7 +43,7 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "react-scripts test",
+    "test": "CI=true react-scripts test --env=jest-environment-jsdom-sixteen",
```
jlhogan, line 46: necessary to get some of the `dom` utils to work

client/package.json
```diff
@@ -59,5 +60,12 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "devDependencies": {
```
jlhogan, line 64: should these just be normal deps?

client/package.json
```diff
@@ -59,5 +60,12 @@
...
+    "@types/jest-axe": "^3.5.0",
+    "jest-axe": "^4.0.0",
+    "jest-canvas-mock": "^2.2.0",
```
jlhogan, line 67: needed to get rid of a warning from some dep

client/src/containers/CheckData/CheckData.tsx
```diff
@@ -17,7 +17,7 @@ const CheckData: React.FC = () => {
   const [reportData, setReportData] = useState<Child[]>([]);
 
   useEffect(() => {
-    if (reportId && accessToken) {
```
jlhogan, line 20: the `api` util does this check; no need to do it 2x

client/src/containers/App/App.test.tsx
```diff
@@ -2,13 +2,15 @@ import React from 'react';
...
-  );
-  expect(asFragment()).toMatchSnapshot();
+const routerWrapped = (
+  <BrowserRouter>
+    <App />
+  </BrowserRouter>
+);
+
+describe('App', () => {
+  snapshotTestHelper(routerWrapped);
```
mmazanec22, line 14: Nice!

client/src/containers/CheckData/CheckData.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+import {
+  waitFor,
+  waitForElementToBeRemoved,
+  render,
+} from '@testing-library/react';
+import { Child, FundingSource } from '../../shared/models';
+import CheckData from './CheckData';
+import { accessibilityTestHelper, snapshotTestHelper } from '../../testHelpers';
+
+const routerWrapped = (
```
mmazanec22, line 12: Should we add a "wrap in router" test helper too maybe?

client/src/containers/CheckData/CheckData.tsx
mmazanec22: Would keeping access token here help prevent unnecessary network requests/is that worth trying to do?

mmazanec22 (approved):

client/src/containers/CheckData/CheckData.tsx
jlhogan: api util does the check -- if accessToken is provided, but has no value (aka is null as opposed to undefined) then the api helper doesn't actually make the network request https://github.com/ctoec/data-collection/blob/base/client/src/utils/api.ts#L67

PR #298, opened 2020-09-08
Add alerts hook

### LINKED ISSUE #252: Scroll to top of page and show success alert when saving info on a child's record.
Record Updated
Your changes to [Firstname Lastname's] record have been saved.
alert

mmazanec22:
Closes #252


client/src/containers/App/App.tsx
```diff
@@ -50,14 +50,14 @@ const App: React.FC = () => {
         ]}
         userFirstName={user?.firstName}
       />
-      <Switch>
-        <main id="main-content">
+      <main id="main-content">
```
mmazanec22, line 55: This was causing some weird behavior from the router-- turns out the components need to be direct descendants of switch

client/src/hooks/useAlerts.tsx
```diff
@@ -0,0 +1,25 @@
...
+type LocationType = Location & {
+  state: {
+    alerts: AlertProps[];
+  };
+};
+
+// Push array of alerts to history state on page before
+// E.g. history.push('new-path', { alerts: [{ ...alertProps }] });
+// Optional additional alerts can be passed so that this hook handles all alert creation
+export default (additionalAlerts: AlertProps[] = []) => {
```
jlhogan, line 14: named exports ? :)

client/src/hooks/useAlerts.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const location = useLocation() as LocationType;
+  const alerts = (location.state ? location.state.alerts : []).concat(
+    additionalAlerts
+  );
+  if (alerts.length) {
```
jlhogan, line 19: niiiiiiiiiiiiice

client/src/hooks/useAlerts.tsx
```diff
@@ -0,0 +1,25 @@
...
+    alerts: AlertProps[];
+  };
+};
+
+// Push array of alerts to history state on page before
+// E.g. history.push('new-path', { alerts: [{ ...alertProps }] });
+// Optional additional alerts can be passed so that this hook handles all alert creation
+export default (additionalAlerts: AlertProps[] = []) => {
+  const location = useLocation() as LocationType;
+  const alerts = (location.state ? location.state.alerts : []).concat(
```
jlhogan, line 16: i think this will error if `location.state.alerts` is not an array, maybe
```
const alerts = (location.state?.alerts || []).concat....
```

client/src/hooks/useAlerts.tsx
```diff
@@ -0,0 +1,25 @@
...
+    additionalAlerts
+  );
+  if (alerts.length) {
+    window.scrollTo(0, 0);
+  }
+  return alerts.map((alertProps, i) => (
```
jlhogan, line 22: we may want to reconsider how we'd display multiple alerts but this def works for now

jlhogan (approved):
nice i like leveraging built ins from other components to do state management stuff for us

mmazanec22:
@jlhogan I changed this to be `return { alertElements, alerts, setAlerts }`.  I think having addAlert and clearAlerts would potentially put us in the annoying situation of having to clearAlerts every time we wanted to add an alert but not preserve the old alerts.  This way the dev has control over whether they want to preserve the old alerts by spreading them (which I think would be the less frequent use case anyway), or just replace them altogether.  LMK what you think!

client/src/hooks/useAlerts.tsx
```diff
@@ -0,0 +1,30 @@
...
+type LocationType = Location & {
+  state: {
+    alerts: AlertProps[];
+  };
+};
+
+// Push array of alerts to history state on page before
+// E.g. history.push('new-path', { alerts: [{ ...alertProps }] });
+// Optional additional alerts can be passed so that this hook handles all alert creation
+export const useAlerts = (initialAlerts: AlertProps[] = []) => {
```
jlhogan, line 14: not sure what the use case would be for initial alerts but that doesn't mean it'll never happen

jlhogan (approved):
LGTM! makes sense RE setAlerts

client/src/hooks/useAlerts.tsx
mmazanec22: I think we had some in ECE reporter-- on the report submit page, with a button, IIRC

PR #300, opened 2020-09-09
281 identifiers

### LINKED ISSUE #281: Divide child info into two sections
Child info becomes "child identifiers"
Screen Shot 2020-09-04 at 12 13 56 PM

Child info includes:
Screen Shot 2020-09-04 at 12 11 34 PM

mmazanec22:
Closes #281


client/src/containers/AddChild/NewEnrollment.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { useContext, useState } from 'react';
-import { Enrollment, Child } from '../../shared/models';
+import { Enrollment } from '../../shared/models';
```
bamader, line 2: Do we not need to import a Child object anymore because we've split the form into its two parts? Looking at some of the code later down in the file (line 25ish) there's an invocation to a `child?.enrollments`, but this might be part of the comment you left about how this form is pretty complicated.

client/src/containers/EditRecord/Forms/ChildIdentifiers/Form.tsx
```diff
@@ -0,0 +1,91 @@
...
+  BirthCertificateFormFieldSet,
+} from './Fields';
+import { Form, FormSubmitButton } from '@ctoec/component-library';
+import { EditFormProps } from '../types';
+import AuthenticationContext from '../../../../contexts/AuthenticationContext/AuthenticationContext';
+import { Child } from '../../../../shared/models';
+import { apiPut } from '../../../../utils/api';
+import useIsMounted from '../../../../hooks/useIsMounted';
+
+// import { useFocusFirstError } from '../../../hooks/useFocusFirstError';
```
bamader, line 18: If we're not using this should we delete the commented out line for cleanliness?

client/src/containers/EditRecord/Forms/ChildIdentifiers/Form.tsx
```diff
@@ -0,0 +1,91 @@
...
+  hideHeader = false,
+}: EditFormProps) => {
+  const { accessToken } = useContext(AuthenticationContext);
+  const isMounted = useIsMounted();
+  const [saving, setSaving] = useState(false);
+  // TODO: HOW ARE WE HANDLING ERRORS?
+  // const [error, setError] = useState(null);
+
+  // Focus should automatically be on first error on page
+  // useFocusFirstError([error]);
```
bamader, line 32: I guess same comment here about deleting commented-out code, but looking at it I see this might be a placeholder marking something for the future. Up to you whether to take it out or leave it in!

client/src/containers/EditRecord/Forms/ChildInfo/Fields/index.ts
```diff
@@ -1,14 +1,4 @@
-export * from './BirthCertificateFieldSet';
```
bamader, line 1: Just a general question here: what differentiates a 'child identifier' from a 'child demographic'? Did we just decide on a distinction and that's what we're going with? Or does this come from some kind of business logic at OEC?

bamader (approved):
LGTM! The comments I left aren't at all blockers. More food for thought / clarifications for me.

client/src/containers/AddChild/NewEnrollment.tsx
jlhogan: my guess is this is clean up -- unless the type is explicitly used, this is not needed (referencing an object of type `type` does not require importing the type)

client/src/containers/EditRecord/Forms/ChildInfo/Fields/index.ts
jlhogan: a logical separation since the form was getting very long with everything in one place -- just divided based on if the variable is a unique identifier for the child, or demographic info about the child

client/src/containers/EditRecord/Forms/ChildIdentifiers/Form.tsx
mmazanec22: I'm going to leave this in to remind us that we need to implement it whenever we implement validations-- otherwise we'll likely forget, since the thing will still work fine without it.

jlhogan (approved):
Should we try out pushing this to qa for a design review? I believe it should just require a button press in circleCI build!

client/src/containers/EditRecord/Forms/ChildInfo/Fields/index.ts
mmazanec22: I think basically that the child info form was getting too long and we wanted to split it up

PR #312, opened 2020-09-10
Funding and Space Types

### LINKED ISSUE #207: Funding and Space types: I can enter data for all the possible funding and space type combinations
As a user, I want funding and space types to match the format in which I already record them. I want to avoid my batch upload failing because I had the wrong formatted space type. 

**Acceptance criteria**
• Users can see the space types from a link on the data definitions page:
Space types

• All of the below spaces are supported by the data model and listed in the space type page:
**Child Day Care or CDC:**
- 1 or Part time or PT
- 2 or Full time or FT
- 3 or Full time/part time or PT/FT or split time

**Priority School Readiness or Competitive School Readiness or School Readiness or PSR or CSR:**
- 4 or Full-day or Full-day/Full-year or FD
- 5 or School-day or School-Day/School-Year or SD
- 6 or Part-Day or Part-Day/Part-Year or PD
- 7 or Extended day or ED

**State Head Start or SHS**
- 8 or Extended day or ED
- 9 or Extended year or EY

**Smart Start or SS**
- 10 or School day or School-day/school-year or SD

JordanGuinn:
Closes #207.

Adds support for all possible `FundingSource` and `FundingTime` permutations, and creates page documenting accepted formats for each.

It _also_ includes a slight refactor, in which all database models now live under `/models/db`, whereas all other models just live under the root `/models`.


client/src/containers/FundingSourceTimes/FundingSourceTimes.tsx
```diff
@@ -0,0 +1,66 @@
...
+        ) : (
+          <></>
+        )
+    },
+    {
+      name: 'Accepted formats',
+      cell: ({ row }) =>
+        row ? (<td>
+          {row.fundingTimes.map(fundingTime => {
+            return <div className="margin-top-3 margin-bottom-3">{fundingTime.formats.map(x => '"' + x + '"').join(' or ')}</div>
```
bamader, line 38: Nit: I generally try to avoid generic variable names like 'x' where possible in favor of clearer names that convey what the variable holds, but this is the smallest of non-blocking nits

client/src/shared/models/FundingSourceTime.ts
```diff
@@ -0,0 +1,8 @@
+import { FundingSource } from ".";
+import { FundingTimeInput } from "./FundingTime";
+
+export interface FundingSourceTime {
+  displayName: string;
+  fundingSources: FundingSource[];    //  All FundingSources that can have the associated FundingTimes
+  fundingTimes: FundingTimeInput[];   //  All valid FundingTimes for the associated FundingSources 
+}
```
bamader, line 8: What's the difference between a `FundingSourceTime` and a `FundingTime` as far as the business logic? My guess is that a child can be funded for a period of time in which multiple funding sources come and go, but I just want to make sure I understand that right.

bamader (approved):
LGTM!

client/src/containers/FundingSourceTimes/FundingSourceTimes.tsx
JordanGuinn: That's fair!  I definitely just threw that in and forgot about it - happy to update

client/src/shared/models/FundingSourceTime.ts
JordanGuinn: This `FundingSourceTime` interface just represents an association between a `FundingSource` (or in the case of PSR/CSR, `FundingSources`) and all of the valid `FundingTime` values that can be supplied alongside it during upload.  It doesn't reflect an actual entity, but rather lets us defined valid associations between the `FundingSource` and `FundingTime` enums.  I picked `FundingSourceTime` cuz I wasn't sure of another name that captured that too well 😅

bamader: Oh lol gotcha, coder's convenience :P

mmazanec22 (approved):

client/src/containers/FundingSourceTimes/FundingSourceTimes.tsx
```diff
@@ -0,0 +1,66 @@
...
+          <></>
+        )
+    }
+  ];
+
+
+  return (
+    <div className="grid-container margin-top-4">
+      <BackButton />
+      <h1>Funding space types</h1>
```
mmazanec22, line 51: Let's use the h1ref thing here too

client/src/shared/models/FundingSourceTime.ts
jlhogan: might be nice to distinguish btwn like -- data models (child, enrollment, etc) and then metadata models (columnmetadata, this, maybe flattened enrollment?) but at least just add a note here indicating  ^^^ that

client/src/shared/models/FundingTime.ts
```diff
@@ -6,3 +6,9 @@ export enum FundingTime {
   ExtendedYear = 'ExtendedYear',
   School = 'School',
 }
+
+export interface FundingTimeInput {
```
jlhogan, line 10: I think maybe better to define this in a separate thing or in the FundingSourceTime model to be clear that its metadata and not actually anything to do with the FundingTime data concept

jlhogan (approved):

client/src/shared/models/FundingSourceTime.ts
JordanGuinn: I was having a similar thought - I think it might make sense to have all of the data models themselves live within a `db` subdirectory of `models` (or something along those lines), and then all interfaces/classes/etc not specific to an actual db component can just live in `models` root.  Thoughts?

Would be easily done, but would also probably be best suited in an independent PR just for sake of scope.

jlhogan: that structure makes sense to me. I have 0 objection to tossing that in here tho, but i am sacreligious about separation and scope

JordanGuinn: HEH, works for me!

jlhogan (approved):
:tada:

RyanHansz:
@JordanGuinn This may make sense for the roster updates @jlhogan is making, but can we use the acronyms for the roster view (part time is PT, full time is FT etc)?

jlhogan:
RE ryan's comment -- probably easiest to change the actual enum values to the full text values ("full" becomes "full time") , and then do the mappings to short version ("FT") for the roster view. Dont know that there's really value in maintaining a third internal value (i.e. "full")

RyanHansz:
I deleted my old comment, it seems we should have room for the full-text space type version on the roster

JordanGuinn:
As things stand right now, the contract space name largely depends on the funding type.  For example:

- A `Full` contract space with a `CDC` funding type is represented as `Full time`
- A `Full` contract space with a `PSR` funding type is represented as `Full day`

Is this just a matter of differing language across funding types?  Or are `full time` and `full day` entirely distinct contract spaces?  Am I even making any sense right now? 😅

jlhogan:
you're making sense -- i think we should maybe add separate times for these different funding types !! will make querying about things easier

RyanHansz:
> As things stand right now, the contract space name largely depends on the funding type. For example:
> 
> * A `Full` contract space with a `CDC` funding type is represented as `Full time`
> * A `Full` contract space with a `PSR` funding type is represented as `Full day`
> 
> Is this just a matter of differing language across funding types? Or are `full time` and `full day` entirely distinct contract spaces? Am I even making any sense right now? 😅

For now we should treat them as distinct contract spaces. Govt being Govt, they're all different things

PR #315, opened 2020-09-11
Export csv

### LINKED ISSUE #91: I can export my data to get it back out

### LINKED ISSUE #323: When editing a child record, the confirmation alert should go away when I navigate to another section of the record.
For example, when I change care 4 kids status, I see a confirmation alert. When I navigate to child info, the alert disappears so I can see a new alert when I make other changes.

### LINKED ISSUE #324: Empty state for enrollment and funding
When there are no active enrollments for a child record, change the call to action section to the following:

Screen Shot 2020-09-14 at 12 54 29 PM

bamader:
Closes #91 . Once we have the roster page created we can use this to download a roster CSV easily. Also ignore the super hacky way of forcing in Child ID strings to the backend call. Once we have submission meaningfully set up we can junk that.

Update: Closes #324 
Update: Closes #323


client/src/components/CSVDownloadLink.tsx
```diff
@@ -1,22 +1,48 @@
...
+  /**
+   * Function that downloads information that the backend puts
+   * together in spreadsheet form.
+   */
+  async function downloadTemplate() {
+    await downloadAuthorizedStreamToFile(
+      `export/csv-upload/${submittedIds}`,
+      accessToken || '',
+      `Uploaded Data.csv`
+    ).then((resp) => console.log(resp));
```
mmazanec22, line 29: Instead of doing anything with the response, probably handling a potential error here would be good

client/src/containers/CheckData/CheckData.tsx
```diff
@@ -64,7 +86,7 @@ const CheckData: React.FC = () => {
         <div className="margin-bottom-0">
           <div className="fixed-buttons">
             <Button text="Back to upload" href="/upload" appearance="outline" />
-            <Button text="Send to OEC" href="/success" />
+            <Button text="Send to OEC" onClick={sendData} />
```
mmazanec22, line 89: If this navigates the user to another page, it needs to be an anchor element and have an href (i.e. button with href is fine but the underlying HTML has to be a link if we're navigating pages).

client/src/containers/CheckData/CheckData.tsx
```diff
@@ -30,6 +30,28 @@ const CheckData: React.FC = () => {
...
+   * were persisted to the DB (e.g. we need the child IDs because
+   * it's possible that a user could delete a record while looking
+   * at the CheckData page, and then we wouldn't want to send
+   * that child's ID to the success page for exporting to CSV).
+   *
+   * It's a hacky workaround right now, but it at least lets us
+   * get the component working so we have the functionality.
+   */
+  function sendData() {
+    const idArray = reportData.map((child) => {
```
mmazanec22, line 49: Can we do this by report ID or something instead?  Or something closer to the way the eventual functionality will work (this is on the roster, right? maybe just exporting all children for that site?)?

client/src/components/CSVDownloadLink.tsx
bamader: Oh lol whoops that's totally an artifact from debugging. Will swap out.

client/src/containers/CheckData/CheckData.tsx
bamader: I had those same thoughts, but I think there's a couple problems with that approach. 1) This is used in two places, actually: the roster page, as well as the CheckData page to confirm upload of an enrollment to OEC. That means we can't use report ID for several reasons. a) Once a roster is formed, there ceases to be a meaningful association between a Child and a particular report (since that child could be updated, new children could be added, and a roster could be comprised of children uploaded across multiple reports), and b) in the case where we come from CheckData, say the user modified or even fully deleted a record before submitting; simply re-using the report ID would ignore this change because it would use the same lookup performed by the CheckData page, so these changes wouldn't be persisted. 2) We can and almost certainly will change the way this works after the roster is developed (because sending all of the IDs of relevant children is pretty clunky), but at present we don't have the infrastructure of either the roster page or the submit data to OEC. So that means the pages that would actually be used to click this button don't have any meaningful data to send to the retrieval page right now.

client/src/containers/CheckData/CheckData.tsx
bamader: This is totally on me, but I don't think I understand the terminology there (specifically of 'anchor element'). The onClick function does successfully change the page. Are you saying that just using `history.push` to change the page is bad? I followed the example that we use on the Upload page, where a `history.push()` call is what takes us to CheckData automatically.

bamader: Actually wait is the basic thrust of the comment to just use an href instead of the history.push?

mmazanec22: Anchor element is just a link: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a.  This is an accessibility issue.  We need to use href instead of history push.

bamader: Gotcha. But a button with a link satisfies this?

client/src/containers/CheckData/CheckData.tsx
mmazanec22: I understand that there are changes coming-- my comment was mostly about trying to think through the way this will work in the future (with roster) and how we might be able to do this now so that it's closer to that.  Where in the design will this ultimately fit-- roster or check data?  If it will fit under roster, and roster is just all children at a particular site, then it might make sense to have the export be all children at a site.

bamader: Now I see what you're saying. I guess to me, it doesn't necessarily make sense to organize it by site, but that might be because I already think it's natural to organize in terms of child ID, and that works for both CheckData and Roster. I don't know exactly which of those two it will ultimately live under, but I was kind of working with the assumption that it wouldn't be owned by either container and the same functionality would be used by both. Hence, the general organization. As far as what a roster organization looks like, I think Julia was working on a PR for making a roster, but did you have any other thoughts on potential formats that are worth doing now?

client/src/containers/CheckData/CheckData.tsx
mmazanec22: Under the hood, when we use href on a button component, our component library uses a link instead of a button element-- so yes

client/src/containers/CheckData/CheckData.tsx
mmazanec22: I'm seeing in the invision prototype that the export button only exists on the roster (as far as I can tell): https://skylight.invisionapp.com/console/Full-Data-Collection-Tool-ckeaf1bpi00wn01yhh6f147bf/ckeaf1cjw00x401yh911f60nc/play#project_console

bamader: Oh interesting. The issue I linked (#91 ) actually implies that there are two places this functionality will live, since it specifically mentions the upload page and then links to a different ticket for the roster export. I was hoping to repurpose the code in both places but I think the goal is both pages use the export.

src/controllers/export.ts
```diff
@@ -0,0 +1,329 @@
...
+  var childString: string[] = [];
+  for (let i = 0; i < cols.length; i++) {
+    const c = cols[i];
+    switch (c.formattedName) {
+      case 'Name':
+        childString.push(
+          child.firstName + ' ' + (child.middleName || '') + child.lastName
+        );
+        break;
+      case 'SASID':
```
mmazanec22, line 88: Instead of hard coding all of these fields, which is a little brittle and would break if anything got updated, can we maybe take advantage of existing backend code that generates the template?

client/src/containers/CheckData/CheckData.tsx
mmazanec22: Ah ok.  Well maybe another way to do this, then, would be to include the report ID as state for the success page?  Since the IDs are just what we're getting from the report anyway in this case?

bamader: I thought about that too, but I think there's another problem: if we just use the report ID as the state to fetch all the respective Child IDs, then any deletions that a user made on the EditRecord page will get ignored. Say a user uploaded an enrollment report but then deleted two children who shouldn't have been on there. If we just fetch the report ID, the backend will hand us the same set of names the user saw _before_deleting the erroneous records. Whereas if we use the child IDs themselves, we don't have this problem.

The manual ID compilation will eventually be discarded completely, though. Once we actually define code for what it means to send the data that a user is looking at to OEC, we can just take that exact same data and queue it up as a parameter to download the spreadsheet. We might just be in more of a holding pattern until that infrastructure gets developed? Because once it works, we can just repurpose the data there to replace the ID parsing here.

src/controllers/export.ts
bamader: I actually looked there before I made this giant switch statement. I followed the trail of how things got generated back into the `columnMetadata` controller, and found that the sheet is made of extractions of the `formattedName` property from each column. The problem for us (and the reason for the switch statement, at least at present), is that neither the formattedName nor the propertyName nor the fieldName properties of a columnMetadata object actually correspond to the variable names that live on a Child object. For example, take the field "Town of Birth." That's how it's displayed in the spreadsheet because that's the formatted name. The property name in a column and in a flattened enrollment record (where things get parsed) is just `town`. But on a Child object, it's `birthTown`. So I was running into all kinds of problems because none of the names on any of the entities matched up, and linking them through this switch logic was kind of the easiest way to make that happen.

I definitely agree it's brittle, and leaving as is, is kind of knowingly incurring at least a bit of tech debt going forward. But since we're already going to have to change the model for how we store/parse Children and enrollments (because of fields like 'model' and 'dualLanguageLearner,' which don't even exist on a Child right now), I figured it was okay to not mess with this until we knew what we were doing with our model later on.

Thoughts?

JordanGuinn: That's tough.  I wonder if long-term we want to think about potential associating `ColumnMetadata` entries with the actual columns on the actual tables that providers are gonna be interacting with, instead of the `FlattenedEnrollment` entity?  Which of course is a much larger discussion and certainly doesn't belong in this ticket.

Ideally, the translation from CSV => db and from db => CSV use the same underlying mechanism, but we have some work to do in order to make that happen.

src/controllers/export.ts
```diff
@@ -0,0 +1,370 @@
...
+  });
+  response.contentType('application/octet-stream');
+  response.send(csvStream);
+}
+
+/**
+ * Retrieve ColumnMetadata information for all columns on the
+ * FlattenedEnrollment model.
+ */
+export function getAllEnrollmentColumns(): ColumnMetadata[] {
```
JordanGuinn, line 227: I think this exact function already exists in `controllers/columnMetadata`

client/src/shared/models/Child.ts
```diff
@@ -10,6 +10,7 @@ import { Moment } from 'moment';
 export interface Child {
   id: string;
   sasid?: string;
+  name: string;
```
JordanGuinn: Why does the `Child` model have both a `name` column and `firstName/middleName/lastName/suffix` columns?

client/src/utils/fileDownload.ts
```diff
@@ -4,13 +4,23 @@ import { saveAs } from 'file-saver';
...
  */
 export async function downloadStreamToFile(
   route: string,
-  defaultFileName?: string
+  defaultFileName?: string,
+  authContext?: string
 ) {
-  const res: Response = await apiGet(route, { jsonParse: false });
+  let res: Response;
+  if (!!authContext && authContext != '') {
```
JordanGuinn, line 16: 😎

src/controllers/enrollmentReports/map.ts
```diff
@@ -60,9 +60,9 @@ export const mapFlattenedEnrollment = async (source: FlattenedEnrollment) => {
  * @param source
  */
 const mapOrganization = (source: FlattenedEnrollment) => {
-  if (!source.provider) return Promise.reject('Provider is required');
+  if (!source.providerName) return Promise.reject('Provider is required');
```
JordanGuinn, line 66: This is probably just a merge thing, but I think these initial validations have since gone away

client/src/containers/SubmitSuccess/SubmitSuccess.tsx
```diff
@@ -12,8 +13,23 @@ import LogoWithCheckSrc from '@ctoec/component-library/dist/assets/images/logoWi
...
+ * we can alter this.
+ *
+ * For now, default way of exporting a CSV is to use the
+ * enrollment report ID of the page the user just checked,
+ * since this corresponds to getting a snapshot view of what
+ * the user just did.
+ */
 const SubmitSuccess: React.FC = () => {
   const h1Ref = getH1RefForTitle();
+  // Any cast when extracting param suppresses type error
```
JordanGuinn, line 29: This comment can probably be deleted

src/controllers/enrollmentReports/map.ts
```diff
@@ -207,28 +212,8 @@ const mapFunding = async (
   organization: Organization,
   enrollment: Enrollment
 ) => {
-  const fundingSource: FundingSource = mapEnum(
-    FundingSource,
-    source.fundingType
-  );
-
-  let fundingTime: FundingTime = mapEnum(FundingTime, source.spaceType);
```
JordanGuinn, line 215: Why was all of this deleted?

src/controllers/export.ts
bamader: Yep you're totally right--I even imported it but it must have disappeared in the rebase and merge!

client/src/shared/models/Child.ts
bamader: So this was put in before I rebased and picked up the split the name column up. However, the reason it's there still applies (I think): one of the big complaints we had with the last iteration was that it involved a disgustingly large switch statement. That was because none of the property names in columnMetadata matched the property names on a Child object. Adding a `name` field to a child to correspond to its full name, as displayed on a birth certificate, is a way to cut out a level of switch logic because 'name' is a property field already defined in columnMetadata.

src/controllers/enrollmentReports/map.ts
bamader: Gotcha. Will take them out--I'd actually thought they were supposed to be there from the merge, but must have mixed it up.

client/src/containers/SubmitSuccess/SubmitSuccess.tsx
bamader: Definitely

src/controllers/enrollmentReports/map.ts
bamader: There's some weirdness going on, because that section isn't deleted on my local copy of the branch. Let me try to push it back with some fixes.

bamader: Got it back in. Looks the merge took it out because the refactored variable names weren't updated. But now it's back and with the proper properties.

client/src/shared/models/Child.ts
JordanGuinn: That doesn't apply anymore though, now that we've fully removed that `name` from the `Child` table, right?

My only concern with this is how we'd need to continually keep `name` in sync with the other four columns that now make up a full name, and how that kinda thing will inevitably result in conflict.

bamader: That's totally fair. That just means I have to put one more 'if' case in the CSV mapping function to check if the property is 'name'. If that's okay then this is an easy fix.

bamader: Unless you're saying that our column metadata is also split up by name part?

bamader: Which it totally is because we modified flattened enrollment...I now understand :P

JordanGuinn: Yep!  Ya beat me to it 😁

bamader: So when we upload data in spreadsheet form, does this now mean we have columns for first name, middle name, last name, and suffix instead of just one for name?

JordanGuinn: Yep!

client/src/containers/CheckData/CheckData.tsx
```diff
@@ -67,14 +72,8 @@ const CheckData: React.FC = () => {
       <div className="CheckData__button-container position-fixed bottom-0 width-full">
         <div className="margin-bottom-0">
           <div className="fixed-buttons">
-            <div className="grid-container">
```
JordanGuinn, line 70: What was the motivation for getting rid of this div?  It was added as a part of https://github.com/ctoec/data-collection/pull/318.

JordanGuinn (commented):
LGTM, just one more question

client/src/components/EditForms/EnrollmentFunding/EnrollmentFunding.test.tsx
```diff
@@ -23,13 +24,13 @@ const child = {
     {
       id: 1,
       child: {} as Child,
-      site: { name: 'Site A' } as Site,
+      site: { siteName: 'Site A' } as Site,
```
jlhogan, line 27: i dont think this change is correct

client/src/components/EditForms/EnrollmentFunding/Form/ChangeEnrollmentForm.tsx
```diff
@@ -89,9 +89,15 @@ export const ChangeEnrollmentForm: React.FC<ChangeEnrollmentFormProps> = ({
   return (
     <Card forceClose={closeCard}>
       <div className="display-flex flex-justify">
-        <span className="header-normal font-heading-md">
-          Has {childName}'s age group and/or site changed?
-        </span>
+        {activeFunding == undefined ? (
```
jlhogan, line 92: hypothetically, a child could have an enrollment w no funding (unfunded), not sure if this check should account for that

client/src/utils/fileDownload.ts
```diff
@@ -4,13 +4,23 @@ import { saveAs } from 'file-saver';
...
  * Download a file stream from the route specified and save it to browser,
- * either with the file name specified or as "download".
+ * either with the file name specified or as "download". Optionally
+ * provide an authentication context to access secure routes.
  */
 export async function downloadStreamToFile(
   route: string,
-  defaultFileName?: string
+  defaultFileName?: string,
+  authContext?: string
```
jlhogan, line 13: nit; call this accessToken?

src/controllers/enrollmentReports/parse.ts
```diff
@@ -164,7 +164,7 @@ function parseFlattenedEnrollment(
     }
 
     // Parse zip codes
-    if (prop.match(/zipcode/i)) {
+    if (prop.match(/zip/i)) {
```
jlhogan, line 167: nit: (truly feel free to ignore) but could we rename both of these columns zipcode instead of zip?

src/controllers/export.ts
```diff
@@ -0,0 +1,356 @@
...
+  'enrollments.site',
+  'enrollments.site.organization',
+  'enrollments.fundings',
+];
+
+/**
+ * Allows sorting of an array of nested property values within
+ * an object (i.e. Enrollments nested within a Child).
+ */
+const propertyDateSorter = <T>(
```
jlhogan, line 32: lets pull this out into a util instead of writing 2x

src/controllers/export.ts
```diff
@@ -0,0 +1,356 @@
...
+  return childrenToMap;
+}
+
+/**
+ * Determines all relevant sites a user with the provided User ID
+ * has access to. Everything is handled in one transaction to minimize
+ * calls to the DB.
+ * @param userId
+ */
+export async function getUserWithSites(userId: number) {
```
jlhogan, line 150: no need to do this user stuff here -- this is all added to the user during the authentication middleware. pass in the user object from the request.

src/routes/export.ts
```diff
@@ -0,0 +1,57 @@
...
+);
+/**
+ * /csv-upload-user/:userId GET
+ * Ask the backend to compile a CSV of child objects formatted according
+ * to the data template. Children are found by looking up all sites the
+ * user has access to, finding all Children with enrollments at these sites,
+ * filtering for duplicates, and compiling a report.
+ */
+exportRouter.get(
+  '/csv-upload-user/:userId',
```
jlhogan, line 44: we don't let the user tell us who they think they are -- we get the user information from the decoded JWT. pass in `req.user` to get all the site/org perm data.

jlhogan (commented):
the renaming looks great! My thoughts on the data access pattern turned into a whole thing: https://github.com/ctoec/data-collection/pull/337, but pretty much rehtinking what enrollment report is makes it very easy to pull the child-based report's worth of up to date data. Then, exporting the report's data would be something like GET to `enrollemnt-reports/:reportId/csv`. 
also gotta make sure user ID is determined from the request user, otherwise there's not actually any access controll happening. And leverage that all the site/org ids are already added to that user instance!

client/src/components/EditForms/EnrollmentFunding/EnrollmentFunding.test.tsx
bamader: Could you explain? It's just a renaming of the name property to line up with a property name elsewhere.

client/src/components/EditForms/EnrollmentFunding/Form/ChangeEnrollmentForm.tsx
bamader: Yeah it probably should. We could show the header change only if there's no `activeFunding` but also if there's no `currentEnrollment`.

src/controllers/enrollmentReports/parse.ts
bamader: Np, can change that!

src/controllers/export.ts
bamader: Oh duh! For some reason I thought it would be insecure to pass the whole user in the body of the request, but it's totally different from passing in a whole array like I was before

bamader:
Good comments, all, Julia. I fixed up those changes you mentioned (the nits were probably good, too, because zip has multiple meanings so zipcode is clearer). I think I got what you were saying about the access being determined by the request user, so I ended up stringify-ing the user and directly parsing it in the backend as a way to avoid the user telling the router what ID to use.

bamader:
Also, as far as the change to access pattern, I think your revision makes  A LOT of sense. There was always something that bugged me  about the way Flattened Enrollment was getting used, and I think you finally put words to it. It sounds like after you merge that PR in, we can go back and refactor a bit to make the snapshot enrollment export use the new setup.

client/src/components/EditForms/EnrollmentFunding/EnrollmentFunding.test.tsx
jlhogan: you're right i was confused when first reading this PR :) all good!

src/controllers/export.ts
```diff
@@ -0,0 +1,277 @@
...
+ * of various Child fields into appropriate formats to display
+ * in cells of the CSV to export.
+ * @param value
+ */
+function formatStringPush(value: any) {
+  if (typeof value == 'boolean') {
+    return value ? 'Yes' : 'No';
+  }
+  if (isMoment(value)) {
+    return value.toDate().toLocaleDateString();
```
jlhogan, line 156: `value.format("MM/DD/YYYY")`

src/routes/export.ts
```diff
@@ -0,0 +1,56 @@
...
+);
+/**
+ * /csv-upload-user/:user GET
+ * Ask the backend to compile a CSV of child objects formatted according
+ * to the data template. Children are found by looking up all sites the
+ * user has access to, finding all Children with enrollments at these sites,
+ * filtering for duplicates, and compiling a report.
+ */
+exportRouter.get(
+  '/csv-upload-user/:user',
```
jlhogan, line 44: so, you don't need to pass in the user at all. By making an authenticated request to the api (i.e. with accessToken as authorization header), the server can securely determine the identity of the requesting user. The authenticate middleware decodes the JWT from header, and uses the "sub" claim from the token to look up the user in our system! passing in the  user means that anyone with valid credentials could ask for _any_ user's data. We want users to only be able to access _their_ data. That means this would look something like:
```
exportRouter.get(
  '/csv',
  passAsyncError( async (req, res) => {
    try {
      const childrenToMap = await controller.getChildrenBySiteIds(req.user.siteIds);
      res.send(controller.streamUploadedChildren(res, childrenToMap))
    } catch () {
      ...
    }
  })
```

src/routes/export.ts
```diff
@@ -0,0 +1,56 @@
+import express from 'express';
+import { Response, Request } from 'express';
+import * as controller from '../controllers/export';
+import { passAsyncError } from '../middleware/error/passAsyncError';
+import { getManager } from 'typeorm';
+import { BadRequestError, NotFoundError } from '../middleware/error/errors';
+import { EnrollmentReport, User } from '../entity';
+import { getSites } from '../controllers/sites';
+
+export const exportRouter = express.Router();
```
jlhogan, line 10: these route names are v confusing to me. I think maybe something like `/export/roster` and `export/enrollment-report/:id` would make more sense to me

src/routes/export.ts
bamader: My mind just exploded with the level of awesome that is. So that means I don't actually have to pass in _any_ body in the call to `downloadFileStream` (just the route, file name, and accessToken) because when `api.ts` sends out the call, it'll send a request that already includes information decoded from the JWT that the middleware parsed from `accessToken`?

client/src/containers/CheckData/CheckData.tsx
bamader: Literally none. I'm chalking it up to a merge oversight!

src/routes/export.ts
bamader: Yep just changed the implementation and I completely see what you mean!

jlhogan: when you send the api request from the client (api.ts), it sends an `Authorization` header, with the value like 
```Authorization: Bearer eyJhbGciOiJSUzI1NiIsI....```. On the backend, when the API recieves that request, it passes it off to the `authenticate` middleware, which decodes that token, uses the result to look up the user in the app and augments that user with some permissioning/authorization data (i.e. siteIds, orgIds)

PR #316, opened 2020-09-11
Child POST returns id for fetch pattern consistency

jlhogan:
For consistency with other API access patterns, make children POST return just 201 created with id, then have the API do another GET.
Remove childId from create-record path because it is not actually needed (page maintains child obj as state var, which has id once it exists). Having id in the path would enable the user to navigate back here if they lost connection etc during middle of create flow _BUT_ this won't work because organization is required, and is passed in via Link navigation to the page. So at some point, we want to decide what should happen if the user navigates to the add record page not from the right place (should they get an informative message? or just be redirected to the roster?) 

also:
- adds back incorrectly removed defaultValue null for birthdate (shouldn't default to today!)
- uses just created determination in new flow if user navigates back to that form section while still in new flow
- pulls steps definition out of the file for readability

### DIFF
--- a/client/src/containers/AddChild/AddChild.tsx
+++ b/client/src/containers/AddChild/AddChild.tsx
@@ -1,17 +1,14 @@
 import React, { useState, useEffect, useContext } from 'react';
 import { BackButton } from '../../components/BackButton';
-import { StepList, Button, StepProps } from '@ctoec/component-library';
+import { StepList } from '@ctoec/component-library';
 import { Child, Organization } from '../../shared/models';
 import { apiGet, apiPost } from '../../utils/api';
 import AuthenticationContext from '../../contexts/AuthenticationContext/AuthenticationContext';
-import { useLocation, useHistory, useParams } from 'react-router-dom';
-import { EditFormProps } from '../../components/EditForms/types';
-import { ChildInfoForm, FamilyInfoForm } from '../../components/EditForms';
-import { NewEnrollment } from './NewEnrollment';
-import { NewFamilyIncome } from './NewFamilyIncome';
-import { ChildIdentifiersForm } from '../../components/EditForms/ChildIdentifiers/Form';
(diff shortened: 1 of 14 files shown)


mmazanec22:
@jlhogan probably we should be grabbing org differently/from the user like we talked about, I forgot to make the ticket for that but will now.  Then we don't need to worry about the navigation to/from being an issue

client/src/containers/AddChild/AddChild.tsx
```diff
@@ -20,148 +16,65 @@ type LocationType = Location & {
...
 
   const activeStep = location.hash.slice(1);
   const history = useHistory();
+  const steps = listSteps(history);
+
+  // On initial load, set url hash to first step hash
   useEffect(() => {
     if (!activeStep) {
-      history.replace({ ...location, hash: steps[0].key });
+      history.replace({ hash: steps[0].key });
```
mmazanec22, line 29: Why get rid of location spread here?  That preserves the organization in the link state

client/src/containers/AddChild/AddChild.tsx
```diff
@@ -188,8 +101,8 @@ const AddChild: React.FC = () => {
         ],
       });
     } else {
-      setRefetchChild((r) => r + 1);
-      history.replace({ ...location, hash: steps[indexOfCurrentStep + 1].key });
+      triggerRefetchChild();
+      history.replace({ hash: steps[indexOfCurrentStep + 1].key });
```
mmazanec22, line 105: Same comment-- why not maintain the rest of location?

mmazanec22 (approved):

client/src/containers/AddChild/AddChild.tsx
jlhogan: location is preserved, even without the spread -- replace will only replace the portion of the location that you provide.

PR #319, opened 2020-09-11
92 validations

mmazanec22:
This is a WIP but I wanted to solicit some feedback on directional correctness before moving on to the UI

### DIFF
--- a/client/src/containers/AddChild/NewEnrollment.tsx
+++ b/client/src/containers/AddChild/NewEnrollment.tsx
@@ -31,7 +31,11 @@ export const NewEnrollment = ({ child, onSuccess }: EditFormProps) => {
 
   const onSubmit = (_enrollment: Enrollment) => {
     setSaving(true);
-    console.log(_enrollment)
+    if (!Object.values(_enrollment).reduce((a, b) => a || b)) {
+      // If all of the values are null or undefined, don't block
+      onSuccess();
+      return;
+    }
     apiPost(
       `children/${child.id}/change-enrollment`,
       { newEnrollment: _enrollment },
@@ -60,7 +64,6 @@ export const NewEnrollment = ({ child, onSuccess }: EditFormProps) => {
         data={enrollment}
         onSubmit={onSubmit}
       >
-        <h2>Enrollment and funding</h2>
         <SiteField<Enrollment>
           sites={sites}
           accessor={(data) => data.at('site')}
--- a/client/src/containers/AddChild/NewEnrollment.tsx
+++ b/client/src/containers/AddChild/NewEnrollment.tsx
@@ -31,6 +31,7 @@ export const NewEnrollment = ({ child, onSuccess }: EditFormProps) => {
 
   const onSubmit = (_enrollment: Enrollment) => {
     setSaving(true);
+    console.log(_enrollment)
     apiPost(
       `children/${child.id}/change-enrollment`,
       { newEnrollment: _enrollment },
@@ -60,12 +61,10 @@ export const NewEnrollment = ({ child, onSuccess }: EditFormProps) => {
         onSubmit={onSubmit}
       >
         <h2>Enrollment and funding</h2>
-        {/* <h3 className="font-heading-md margin-bottom-0">Site</h3> */}
         <SiteField<Enrollment>
           sites={sites}
           accessor={(data) => data.at('site')}
         />
-        {/* <h3 className="font-heading-md margin-bottom-0">Start date</h3> */}
         <EnrollmentStartDateField<Enrollment>
           accessor={(data) => data.at('entry')}
         />
--- a/client/src/containers/AddChild/NewEnrollment.tsx
+++ b/client/src/containers/AddChild/NewEnrollment.tsx
@@ -68,16 +68,13 @@ export const NewEnrollment = ({ child, onSuccess }: EditFormProps) => {
         <EnrollmentStartDateField<Enrollment>
           accessor={(data) => data.at('entry')}
         />
-        <h3 className="font-heading-md margin-bottom-0">Age group</h3>
         <AgeGroupField<Enrollment> accessor={(data) => data.at('ageGroup')} />
-        <h3 className="font-heading-md margin-bottom-0">Funding</h3>
         <FundingField<Enrollment>
           fundingAccessor={(data) => data.at('fundings').at(0)}
           getEnrollment={(data) => data.value}
           fundingSpaces={fundingSpaces}
           reportingPeriods={reportingPeriods}
         />
-        <h3 className="font-heading-md margin-bottom-0">Care 4 Kids</h3>
         <CareForKidsField />
         <FormSubmitButton
           text={saving ? 'Saving...' : 'Save'}
--- a/client/src/containers/AddChild/NewEnrollment.tsx
+++ b/client/src/containers/AddChild/NewEnrollment.tsx
@@ -31,7 +31,7 @@ export const NewEnrollment = ({ child, onSuccess }: EditFormProps) => {
 
   const onSubmit = (_enrollment: Enrollment) => {
     setSaving(true);
-    if (!Object.values(_enrollment).reduce((a, b) => a || b)) {
+    if (!Object.values(_enrollment).every((value) => !value)) {
       // If all of the values are null or undefined, don't block
       onSuccess();
       return;
--- a/client/src/containers/EditRecord/EditRecord.tsx
+++ b/client/src/containers/EditRecord/EditRecord.tsx
@@ -96,6 +96,8 @@ const EditRecord: React.FC = () => {
     reportingPeriods,
   };
 
+  console.log(rowData)
+
   return (
     <div className="margin-top-4 grid-container">
       <BackButton />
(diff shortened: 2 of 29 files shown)


client/src/containers/EditRecord/EditRecord.tsx
```diff
@@ -94,6 +94,8 @@ const EditRecord: React.FC = () => {
     reportingPeriods,
   };
 
+  console.log(rowData)
+
```
bamader, line 98: Is this an artifact from debugging or is this intentional to keep? Seems like we don't need it.

src/entity/Child.ts
```diff
@@ -28,48 +32,64 @@ export class Child implements ChildInterface {
...
 
   @Column()
+  @IsNotEmpty({ message: 'Last name is required' })
   lastName: string;
 
   @Column({ nullable: true })
   suffix?: string;
 
   @Column({ nullable: true, type: 'date', transformer: momentTransformer })
+  @IsNotEmpty() // do we need this?
```
bamader, line 49: I feel like it's probably a good idea to be robust with the validations checks? So maybe we could stick in a message like with the previous two columns, in the vein of `{ message: 'Age is a required field' }`.

src/entity/Child.ts
```diff
@@ -28,48 +32,64 @@ export class Child implements ChildInterface {
...
+  // TODO: what are the actual age parameters?
+  @MinDate(moment().add(-7, 'years').toDate(), { message: 'Birth date must be within last 7 years' })
+  @MaxDate(moment().add(-6, 'weeks').toDate(), { message: 'Child must be older than 6 weeks' })
```
bamader, line 52: Def check me on this, but I have a vague memory of 0 to 15 being used for something (I know there are programs that run 0-5 and birth to 3, but I have an impression that birth to 15 also corresponds to something). Might be a question for Ryan, though.

src/entity/Child.ts
```diff
@@ -28,48 +32,64 @@ export class Child implements ChildInterface {
...
   birthdate?: Moment;
 
   @Column({ nullable: true })
+  @IsNotEmpty({ message: 'Birth town is required' })
   birthTown?: string;
 
   @Column({ nullable: true })
+  // TODO: do we account for birth certs from outside the US?
+  @Length(2)
```
bamader, line 61: Is there even a standardized way we could do this? I feel like we don't account for this, but for non-US countries I'm not sure how we would standardize things. Province or county, maybe? But what if a country doesn't have either and it's just a birth town?

src/entity/decorators/childGenderValidation.ts
```diff
@@ -0,0 +1,24 @@
...
+const childGenderSpecifiedMessage = 'Child gender must be specified';
+
+export function ChildGenderSpecified(validationOptions?: ValidationOptions): PropertyDecorator {
+  return function (object: Object, propertyName: string) {
+    registerDecorator({
+      name: 'gender',
+      target: object.constructor,
+      propertyName: propertyName,
+      options: { message: childGenderSpecifiedMessage, ...validationOptions },
+      constraints: [{ childRace: childGenderSpecifiedMessage }],
```
bamader, line 16: Why is the `childGenderValidation` constrained by the `childRace` field?

src/entity/decorators/fundingOverlapValidation.ts
```diff
@@ -0,0 +1,59 @@
...
+const fundingPeriodOverlap = 'Cannot claim a child twice in a reporting period';
+
+export function FundingDoesNotOverlap(validationOptions?: ValidationOptions): PropertyDecorator {
+  return function (object: Object, propertyName: string) {
+    registerDecorator({
+      name: 'fundingOverlap',
+      target: object.constructor,
+      propertyName: propertyName,
+      options: { message: fundingPeriodOverlap, ...validationOptions },
+      constraints: [{ childRace: fundingPeriodOverlap }],
```
bamader, line 16: Same question as with the `childGenderSpecified` code. Is there an aspect of childRace that functions as a constraint on all of these validations?

bamader (commented):
Basically looks good. There's just a few comments I had addressing some inline code comments you put in the files that it's probably good to talk through. Otherwise, it's basically ready for approval.

client/src/containers/EditRecord/EditRecord.tsx
mmazanec22: Yeah this is a WIP, still have to do the whole front end piece, so I left the cruft in for now-- mostly wanted feedback on whether the decorator stuff I have going on is what we're after

src/entity/decorators/childGenderValidation.ts
mmazanec22: Lol good catch

src/entity/decorators/fundingOverlapValidation.ts
mmazanec22: No, this is me messing up-- childRace isn't a field, just an arbitrary key because the constraint needed a key of some kind, and then I copied that one for these other validations, hence this weirdness

src/entity/decorators/childRaceValidation.ts
```diff
@@ -0,0 +1,32 @@
...
+  raceFields.forEach(field => {
+    const val = child[field];
+    if (val) {
+      return true;
+    }
+  });
+  return false;
+}
+
+export function ChildRaceIndicated(validationOptions?: ValidationOptions): PropertyDecorator {
```
JordanGuinn, line 17: I don't know if this is a thing that TypeORM allows or not, but would this decorator be best served at the entity level?  Especially if the check run against each race column evaluates all of the other columns anyway.

JordanGuinn (commented):
Overall this seems like exactly how we wanna use TypeORM decorators 🔥🎉 BIG fan!

src/entity/decorators/childRaceValidation.ts
mmazanec22: Seems like we can't do decorators at the entity level with this library: https://github.com/typestack/class-validator/issues/182

jlhogan:
agreed this looks and feels really good to me!

mmazanec22:
Ok team I added a few header fixes that I must have missed elsewhere (or maybe there was a regression?? idk) and also made it possible to save without entering any enrollment data.  I'd like to merge this and start on the UI stuff in another branch.  Can someone take another look at this and approve if it's ready?  @jlhogan @JordanGuinn @bamader

client/src/containers/AddChild/NewEnrollment.tsx
```diff
@@ -31,6 +31,11 @@ export const NewEnrollment = ({ child, onSuccess }: EditFormProps) => {
 
   const onSubmit = (_enrollment: Enrollment) => {
     setSaving(true);
+    if (!Object.values(_enrollment).reduce((a, b) => a || b)) {
```
jlhogan, line 34: nit: 
```
Object.values.(_enrollment).every((prop) => !prop)
```

jlhogan: is the path here that if all values are null/undefined, the user hits save but does not actually create any objects in the API?

src/entity/Child.ts
```diff
@@ -28,57 +38,78 @@ export class Child implements ChildInterface {
...
   @Column({ nullable: true })
+  // No validation bc the only way to set it to false is to select and unselect
   hispanicOrLatinxEthnicity?: boolean;
 
   @Column({ nullable: true, type: 'simple-enum', enum: Gender })
+  @ChildGenderSpecified()
   gender?: Gender;
 
   @Column({ nullable: true })
+  // No validation bc the only way to set it to false is to select and unselect
```
jlhogan, line 104: what does this mean? we need to somehow confirm that these values get true/false values

src/entity/decorators/childGenderValidation.ts
```diff
@@ -0,0 +1,24 @@
+import { registerDecorator, ValidationOptions } from 'class-validator';
```
jlhogan, line 1: maybe worth moving these into `decorators/validations/[child | enrollment....]` dirs so names can be shorter and easier to navigate around

jlhogan (approved):

client/src/containers/AddChild/NewEnrollment.tsx
mmazanec22: Yeah-- other option was not requiring site on enrollment but that felt pretty wrong

src/entity/Child.ts
mmazanec22: Oh you know what I think I was wrong about this for ethnicity since it's a radio button-- but for some of the checkbox ones (homelessness, foster), the only way to get a false value is to select and then unselect it.  So it needs to default to false if we want to have a false value.

src/entity/Child.ts
```diff
@@ -28,57 +38,76 @@ export class Child implements ChildInterface {
...
+  @IsNotEmpty()
+  // No default value because this is a radio button
   hispanicOrLatinxEthnicity?: boolean;
 
   @Column({ nullable: true, type: 'simple-enum', enum: Gender })
+  @ChildGenderSpecified()
   gender?: Gender;
 
-  @Column({ nullable: true })
+  @Column({ default: false })
```
mmazanec22, line 104: @jlhogan how's this for a fix for the checkbox fields-- just defaulting them all to false?

jlhogan: mucho dig !!!!

client/src/containers/AddChild/NewEnrollment.tsx
jlhogan: makes sense!

PR #322, opened 2020-09-14
Repin component-library & update test snapshots

JordanGuinn:
- Repin the `component-library` dep to the `base` branch and not a since-removed feature branch.
- Use a specific time zone when running front-end tests, so the snapshots are consistent regardless of the machine the snapshots are generated on
- Update test snapshots

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@ctoec/component-library": "https://github.com/ctoec/component-library#base",
+    "@ctoec/component-library": "https://github.com/ctoec/component-library#static-date-for-example",
     "@openid/appauth": "^1.2.7",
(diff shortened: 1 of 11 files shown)


bamader (approved):
Lol like all of our PRs have the yarn and package.json fixes in them right now

client/package.json
```diff
@@ -43,7 +43,7 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "CI=true react-scripts test --env=jest-environment-jsdom-sixteen",
+    "test": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen",
```
JordanGuinn, line 46: So that different snapshots aren't generated depending on the machine run from (PST vs EST to whatever timezone CircleCI uses).

PR #326, opened 2020-09-15
Split up Name column

### LINKED ISSUE #258: Separate first/middle/suffix/last name in the data model.

JordanGuinn:
Closes #258.


jlhogan (approved):

src/controllers/enrollmentReports/map.ts
```diff
@@ -82,44 +82,19 @@ const mapSite = (source: FlattenedEnrollment) => {
...
   organization: Organization,
   family: Family
 ) => {
-  const SUFFIXES = ['sr', 'jr', 'ii', 'iii'];
-  const normalizedName = source.name
-    ? source.name.trim().replace(/\s+/, ' ')
-    : '';
-  const nameParts = (normalizedName || '').split(' ');
-  if (nameParts.length < 2) {
+  if (!source.firstName || !source.lastName) {
```
mmazanec22, line 92: Maybe not for this PR but should we add validations to flattened enrollment?  And just check if it's valid here?

mmazanec22 (approved):

src/controllers/enrollmentReports/map.ts
JordanGuinn: Oooooo, that makes much more sense to me - will do.

jlhogan: I dont know that this is a real validatoin -- i jjust put this here! Let's check with @RyanHansz what we're thinking. Overall probably better to do _all_ validations at one level i.e. on the child model and not spread them out over flattened enrollment + real db models

RyanHansz: Is the potential validation rule that first name and last name are required?

jlhogan: specifically that they are required when doing a bulk upload -- they'll def be required before data is submittable, but what should happen to a bulk upload row that is missing a first or last name? i think easiest is -- nothing, let the user deal with that

RyanHansz: Hmm, yes, we'll need to flag it somehow. I have stuff to show that gets at this tomorrow in backlog refinement

JordanGuinn: Gotcha.  Then for the purposes of this ticket, I'm thinking I'll just remove these field-specific validations that take place as a part of the file upload process and defer that logic to the `CheckData` component.

PR #327, opened 2020-09-15
268 roster

### LINKED ISSUE #268: Group roster: I can see enrollments grouped by age group
# [Designs and icons](https://share.goabstract.com/b5de041e-5083-4536-b5c1-43433094ea28)
• The page header is my org name with number of children and sites
• The roster is grouped by age group with total number of children per each age group
• I can show/hide age groups by clicking the show/hide links
• I can sort each column, which is sorted by last name A-Z by default

Grouped Roster

## Roster header sections:
Screen Shot 2020-09-10 at 2 10 42 PM
Border: 1px lightest-gray (#DCDEE0)

jlhogan:
closes #268


client/src/containers/GettingStarted/GettingStarted.tsx
```diff
@@ -43,6 +43,7 @@ const GettingStarted: React.FC = () => {
       </div>
       <div className="grid-row margin-top-2">
         <Button text="Go to file upload" href="/upload" />
+        <Button text="Go to roster" href="/roster" />
```
jlhogan, line 46: until we redo the getting started page, just added this button

client/src/containers/Roster/TableColumns.tsx
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
```
jlhogan, line 1: TODO: put this in a place where it can be reused instead of defined 2x (here, and in "check data") BUT i tihnk it might not end up being used in "check data" if that becomes a more interactive resolve flow so felt like duplicating for now was fine

src/routes/index.ts
```diff
@@ -18,7 +18,7 @@ export const router = express.Router();
 router.use('/column-metadata', columnMetadataRouter);
 
 /* AUTHENTICATED ROUTES */
-router.use('/users', authenticate, userRouter);
+router.use('/users', authenticate, usersRouter);
```
jlhogan, line 21: rider: renamed for consistency

src/routes/users.ts
```diff
@@ -0,0 +1,17 @@
...
+import * as controller from '../controllers/users';
+
+export const usersRouter = express.Router();
+
+/**
+ * /users/current GET
+ *
+ * Returns the user associated with the
+ * authenticated request, augmented with all
+ * read/write sites and orgs
```
jlhogan, line 11: this could be optional, but i figure we will pretty much always want this info -- now, in the client all the sites/orgs for the user live in the usercontext!!

src/utils/getReadAccessibleOrgIds.ts
```diff
@@ -0,0 +1,18 @@
+import { User, Site } from '../entity';
+import { getManager } from 'typeorm';
+
+/**
+ * Gets all orgIds that the user has READ access to -- that is,
+ * all orgIds they have permissions for, and all orgIds for sites
+ * that they have permissions for.
+ * @param user
+ */
+export const getReadAccessibileOrgIds = async (user: User) => {
```
jlhogan, line 10: should we just always add these IDs to the user on auth ? (like siteIds, organizationIds -- which are implicitly write accessible); do we need to be more explicit about what "permissions" a user has? Not yet i dont think

RyanHansz: I have a hunch that Hartford BOE may be one of our first users, and that role has access to multiple orgs, so this may come up soon.

jlhogan:
open design Qs:
- should roster sections be expanded by default? Should that happen if they are below a certain length? can easily expand/collapse based on some criteria
- should we do a different thing besides showing 0s while the data is loading?

RyanHansz:
> open design Qs:
> 
> * should roster sections be expanded by default? Should that happen if they are below a certain length? can easily expand/collapse based on some criteria
> * should we do a different thing besides showing 0s while the data is loading?
> * should the accordion be full page width, and the inner tables inside grid container?
> 
> known issues:
> 
> * fix swapped expand/collapse copy

Cool to see the roster again! 

1 - Lets start with roster expanded by default up to 50 records
2 - Yes, but not sure what exactly, I'll revisit loading states later
3 - The accordions are fine lined up with the roster items

Here are some changes I noticed, let me know if any of them should be a separate ticket:
1 - Reduce age group title font-size from x-large to large
2 – Sort the roster by last name by default (same as ECE reporter)
3 – All column headers should be sortable
4 – Age group column doesn't need to show in the roster
5 – If a user has access to multiple Orgs, the organization column should show up on the roster
6 – Change all table-data font-size to 2xs (14px)
7 – Truncate text (Ex: Hogwarts childca...) if it will span two rows

jlhogan:
thanks ryan! can do those here, no other ticket needed

RyanHansz:
Can we put margin-bottom: 4rem; on each table to give more space between age groups? 
And can the names be [Lastname, Firstname] 
After that I think this is good to go!

JordanGuinn (approved):
LGTM! 👍🏽👍🏽👍🏽👍🏽

client/src/routes.ts
```diff
@@ -52,6 +53,11 @@ export const routes: RouteConfig[] = [
     component: FundingSourceTimes,
     unauthorized: true,
   },
+  {
+    path: '/roster',
+    component: Roster,
+    unauthorized: false,
```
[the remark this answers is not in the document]
jlhogan: it does indeed! Unauthorized == false so authorized == true :)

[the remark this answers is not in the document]
JordanGuinn: LOL, my bad - sometimes I forget the true === not true bit 😅🙈

jlhogan:
> Can we put margin-bottom: 4rem; on each table to give more space between age groups? 
so there's 4rem between the bottom border of the table content and the top border of the next accordion header? should both of these borders (bottom of table AND top of next header) exist?

PR #334, opened 2020-09-17
Auto-generate users present in WingedKeys but missing from Fawkes

JordanGuinn:
### Summary
I know this solution isn't necessarily _good_, but I've been trying to figure out if it introduces any vulnerabilities and I honestly can't think of any.  That being said, we should def do away with this once we have an actual user management solution in place. 😅

If somebody makes a request with a JWT that, when decoded, refers to a user that Fawkes isn't familiar with, we ask Winged Keys for more details about the user.  If there is in fact a Winged Keys user corresponding to the decoded JWT's sub that isn't present in Fawkes, just _add them to Fawkes_.

If there's a huge vulnerability that I'm overlooking here, PLEASE let me know and I'll blow this all away. 🙃

### Associated PRs
- https://github.com/ctoec/winged-keys/pull/54
- https://github.com/skylight-hq/ctoec-devops/pull/89


src/middleware/authenticate.ts
```diff
@@ -94,7 +118,7 @@ const getUser = async (wingedKeysId: string) => {
 };
 
 /**
- * Full authnetication middleware chains together decodeClaim and addUser,
+ * Full authentication middleware chains together decodeClaim and addUser,
```
jlhogan, line 121: hehe ty

src/middleware/authenticate.ts
```diff
@@ -34,8 +36,30 @@ const decodeClaim = jwt({
...
+        });
+
+        if (res && res.data && res.data.sub && res.data.sub === req.claims.sub) {
+          const _user = getManager().create(User, {
+            wingedKeysId: req.claims.sub,
+            firstName: res.data.given_name,
+            lastName: res.data.family_name,
+          });
+          
+          user = await getManager().save(_user);
```
jlhogan, line 59: i see no issue with this security-wise, however we will still need to create permissions for this user in order for it to actually _show_ you anything. One possibility that I don't mind at all here, since this is not going to be how real user management happens, is just create org permissions for all organizations in the fawkes DB for this kind of user?

jlhogan (approved):

src/middleware/authenticate.ts
JordanGuinn: Works for me!  _Especially_ right now, since it's largely just us using it, and we don't really have any data points in WingedKeys (at least that I'm aware of) to help us determine who should have access to what.

jlhogan: we do not, and that was intentional because ID server was envisioned as the auth service for any/all OEC web apps. We wanted to keep app-specific authorization in the app, and have ID server act only as the authentication provider

JordanGuinn: Oh _totally_ - WingedKeys opens the door, Fawkes shows you around (or some similar metaphor)

bamader (approved):
Lol much like that great commit name you have, this sure is easy...to approve 😛

mmazanec22 (approved):

PR #337, opened 2020-09-18
Better enrollment report

jlhogan:
been thinking a lot about the unit of a "report" and how/when/at what points data is accessed and stored in the DB. Settled on this, which is:
- a "report" actually stores a list of childIds that appeared in that report
  - REASON: easy to get the set of data that the user added via that report, but ensuring that data is the edited (and when this exists, reconciled) version of the data (as opposed to option of pulling FlattenedEnrollment records, which are never updated and so are not to be shown to the user ever!)
  - can also produce a point in time version of the report (pre successive edits or reconciliation or whatever ((i dont know why we would need this))) by using the report upload timestamp to do time-versioned query for child, enrollment,family, funding,etc data
  - but also means needed to create a join table for this mapping (alternative -- we could store a LIST of child guids in the report enrollment table, but that just felt wrong-er). Creating a join table just _wasnt_ working with the custom naming strategy I had introduced, so we're back to the weird inconsistent default naming that happens with typeorm, which is snake case `table_names`, and `columnNames` inherited directly from property names, so camel case. Is it ugly? Yes. Is it worth spending MORE TIME in typeorm source code to figure out? I think not... but I welcome anyone who is interested to bring back the `pascalCaseNamingStrategy` and `src/ormconfig.ts` and play around. B/c of this, I recreated db again :) 

- no longer storing FlattenedEnrollment AND mapped data in the DB. This has been feeling wrong and confusing to me conceptually for some time so i like that it is going away

- FlattenedEnrollment is now EnrollmentReportRow, and it has all the same metadata decoration but is not a typeorm entity

- to compensate for no more typeorm entity, had to add default values to  props on EnrollmentReportRow (b/c ts isn't actually a typed language its just compile stuff so to do any of that introspection at run time default values and their implicit constructors are necessary) 

- EnrollmentReportRow now lives in a top-level "/template" directory, which i feel is pretty random but it no longer felt like it should live in "/entity". I Settled on it's own directory b/c it is a pretty top-level concept in the app. The column metadata decorators are also there. 
```
src
 |
...
├── template
│   │   ├── decorators
│   │   ├── EnrollmentReportRow.ts
│   │   ├── getAllColumnMetadata.ts
│   │   └── index.ts
```
- made "column-metadata" route a "template" route, with "column-metadata", "csv", and "xlsx" subroutes

- also, made a couple fields on db models nullable and optional, since we decided to enforce all things via validations
- as we think about what "submitted" means, maybe we put a `submittedAt` timestamp on enrollment report (and possibly the other ad hoc report data elements as well, like in hedwig but that's a later thing)


src/controllers/enrollmentReports/parse.ts
```diff
@@ -1,34 +1,28 @@
...
-  getColumnMetadata,
+  EnrollmentReportRow,
+  SECTIONS,
   DATE_FORMATS,
   REPORTING_PERIOD_FORMATS,
-} from '../../entity/decorators/columnMetadata';
-import moment from 'moment';
+} from '../../template';
 import { BadRequestError } from '../../middleware/error/errors';
+import { getAllColumnMetadata } from '../../template/getAllColumnMetadata';
```
jlhogan, line 11: todo: export this from /template index

src/controllers/enrollmentReports/parse.ts
```diff
@@ -69,8 +55,21 @@ export function parseUploadedTemplate(file: Express.Multer.File) {
...
+  Object.entries(new EnrollmentReportRow()).forEach(([prop, value]) => {
+    if (typeof value === 'boolean') {
+      booleanProperties.push(prop);
+    }
+    if (moment.isMoment(value)) {
+      dateProperties.push(prop);
+    }
+  });
+
+  console.log('datePrperties', dateProperties);
```
jlhogan, line 69: remove

src/routes/template.ts
```diff
@@ -0,0 +1,35 @@
+import express from 'express';
+
+import * as controller from '../controllers/template';
+import { getAllColumnMetadata } from '../template/getAllColumnMetadata';
+
+export const templateRouter = express.Router();
+/**
+ * /enrollment-reports/template/column-metadata GET
```
jlhogan, line 8: todo: remove `/enrollment-reports` from path in comment no longer true

client/src/components/Forms/FamilyIncome/Form/index.tsx
```diff
@@ -10,7 +10,7 @@ export const FamilyIncomeForm: React.FC<EditFormProps> = (props) => {
   const [showRedeterminationForm, setShowRedeterminationForm] = useState(false);
   const [currentIsNew, setCurrentIsNew] = useState(false);
 
-  if (!child) return <></>;
+  if (!child || !child.family) return <></>;
```
mmazanec22, line 13: Should we throw an error here instead?

mmazanec22 (approved):

client/src/components/Forms/FamilyIncome/Form/index.tsx
bamader: I would agree with this, since family is nested within child

src/controllers/enrollmentReports/map.ts
```diff
@@ -33,7 +33,7 @@ import { getManager } from 'typeorm';
  * TODO: Cache Org/Site, since they are definitely reused a lot in an enrollment report
  * @param source
  */
-export const mapFlattenedEnrollment = async (source: FlattenedEnrollment) => {
+export const mapFlattenedEnrollment = async (source: EnrollmentReportRow) => {
```
bamader, line 36: Nit: should we change the name of the function since it's no longer mapping a `flattenedEnrollment` object? Maybe to `mapReportRow` or `parseReportRow` or something?

bamader (approved):
LGTM! Nice, clean refactoring.

PR #356, opened 2020-09-22
Squash export

### LINKED ISSUE #91: I can export my data to get it back out

### LINKED ISSUE #323: When editing a child record, the confirmation alert should go away when I navigate to another section of the record.

### LINKED ISSUE #324: Empty state for enrollment and funding

bamader:
Closes #91 . 
Closes #324 
Closes #323

CSV exporting that functions from both the roster page (which obtains all child records a provider is authorized to see) and the SubmitSuccess page (which downloads a snapshot of the data just uploaded).


client/src/components/CSVDownloadLink.tsx
```diff
@@ -1,22 +1,60 @@
...
+
+  return (
+    <Button
+      appearance="unstyled"
+      onClick={downloadTemplate}
+      className="text-bold margin-bottom-3 display-block"
+      external
+      text={
+        <TextWithIcon
+          text="Download.csv"
```
mmazanec22, line 52: Smallest of nitpicks: I think that download in this case is supposed to be a verb, not a filename-- so there should be a space between Download and .csv

client/src/components/Forms/CareForKids/CareForKidsField.tsx
```diff
@@ -14,7 +14,7 @@ import { Child } from '../../../shared/models';
 export const CareForKidsField: React.FC = () => {
   return (
     <FormField<Child, RadioButtonGroupProps, boolean>
-      getValue={(data) => data.at('recievesC4K')}
+      getValue={(data) => data.at('receivesC4K')}
```
mmazanec22, line 17: Good catch!

client/src/components/Forms/CareForKids/Form.tsx
```diff
@@ -12,10 +19,16 @@ import { apiPut } from '../../../utils/api';
...
   child,
   onSuccess,
+  setAlerts,
 }) => {
   const { accessToken } = useContext(AuthenticationContext);
   const [isSaving, setIsSaving] = useState(false);
 
+  // Clear any previously displayed alerts from other tabs
+  useEffect(() => {
+    setAlerts([]);
```
mmazanec22, line 29: Hmm we prob need to address this in the way use alerts works, rather than having to add this to each form (not necessarily an issue for this PR, just noting)

client/src/components/Forms/EnrollmentFunding/Form/ChangeEnrollmentForm.tsx
```diff
@@ -24,6 +24,7 @@ import {
   ReportingPeriodField,
   FundingField,
 } from '../Fields';
+import { isNullOrUndefined } from 'util';
```
mmazanec22, line 27: Unused import?

client/src/components/Forms/EnrollmentFunding/Form/ChangeEnrollmentForm.tsx
```diff
@@ -88,9 +89,15 @@ export const ChangeEnrollmentForm: React.FC<ChangeEnrollmentFormProps> = ({
   return (
     <Card forceClose={closeCard}>
       <div className="display-flex flex-justify">
-        <span className="header-normal font-heading-md">
-          Has {childName}'s age group and/or site changed?
-        </span>
+        {activeFunding == undefined && currentEnrollment == null ? (
+          <h2 className="header-normal font-heading-lg">
```
mmazanec22, line 93: This should not be a header-- it is not used to organize content.

client/src/components/Forms/FamilyIncome/Form/index.tsx
```diff
@@ -1,19 +1,26 @@
...
+  // Clear any previously displayed alerts from other tabs
+  useEffect(() => {
+    setAlerts([]);
+  }, []);
+
   if (!child?.family) {
     throw new Error('Family income rendered without family');
   }
 
+  if (!child) return <></>;
```
mmazanec22, line 22: We shouldn't need this here because if there's no child, the expression on line 19 will evaluate to undefined and an error will be thrown

client/src/containers/CheckData/CheckData.tsx
```diff
@@ -40,32 +45,31 @@ const CheckData: React.FC = () => {
...
-              id="enrollment-report-table"
-              rowKey={(row) => row.id}
-              data={reportData}
-              columns={tableColumns}
-            />
-          </PerfectScrollbar>
-        ) : (
-          'Loading...'
-        )}
+        <div className="margin-x-4">
```
mmazanec22, line 48: Instead of adding this div with an x margin, we should change line 47 to just be margin-4 rather than margin-top-4

client/src/containers/Roster/Roster.tsx
```diff
@@ -102,17 +104,7 @@ const Roster: React.FC = () => {
...
-            className="margin-left-2"
-            appearance="unstyled"
-            text={
-              <TextWithIcon
-                Icon={FileDownload}
-                text="Export current view (PLACEHOLDER)"
-              />
-            }
-          />
+          <CSVDownloadLink reportId={undefined} />
```
mmazanec22, line 107: Prob don't need to explicitly specify reportId as undefined here?

client/src/shared/models/db/Family.ts
```diff
@@ -6,7 +6,7 @@ export interface Family extends ObjectWithValidationErrors {
   streetAddress?: string;
   town?: string;
   state?: string;
-  zip?: string;
+  zipcode?: string;
```
mmazanec22, line 9: Nit: we should prob make zipcode zipCode for camel case consistency

src/entity/Funding.ts
```diff
@@ -8,6 +8,7 @@ import { FundingSpace } from './FundingSpace';
 import { UpdateMetaData } from './embeddedColumns/UpdateMetaData';
 import { ValidateNested, IsNotEmpty } from 'class-validator';
 import { LastReportingPeriodAfterFirst } from './decorators/Funding/lastReportingPeriodValidation';
+import { Moment } from 'moment';
```
mmazanec22, line 11: Unused import?

src/utils/propertyDateSorter.ts
```diff
@@ -0,0 +1,20 @@
...
+  a: T,
+  b: T,
+  accessor: (_: T) => Moment | null | undefined
+) => {
+  const aDate = accessor(a);
+  const bDate = accessor(b);
+
+  if (!aDate) return 1;
+  if (!bDate) return -1;
+  if (aDate < bDate) return 1;
```
mmazanec22, line 17: Since we're using moments, you can use aDate.isSameOrBefore(bDate) here

client/src/components/CSVDownloadLink.tsx
bamader: OH that's why that's there. It was originally like that and I was like "why is there a space between a the filename and the extension." Now I get it.

client/src/components/Forms/CareForKids/Form.tsx
bamader: Yeah I would agree with that. I think it's beyond the scope of this PR, but if we could treat alerts as having some sort of "state" functionality with a `setState` function we could pass around that might solve things?

client/src/components/Forms/EnrollmentFunding/Form/ChangeEnrollmentForm.tsx
bamader: Good catch! Sometimes I forget to go back and clean up stuff I was using for debugging.

client/src/components/Forms/EnrollmentFunding/Form/ChangeEnrollmentForm.tsx
bamader: Is it not? I was interpreting it as organizing the current enrollments for a Child. It's just that if a Child has no current enrollments, there's nothing to display in that section (i.e. no content), so the header becomes a vacuous one. To me that was still organizational, but I can definitely understand if that's not really the case.

mmazanec22: When a user clicks the expand card button, it pops open the card expansion, which starts with the "new enrollment" h3 (since this whole component is nested underneath another h2, in line 39 of EnrollmentFunding/Form/index).  Header levels should be a structure that would make sense as an outline-- so we currently have:

I. [Edit Record Child Name]
  A. Enrollment and funding
    1. New enrollment
    2. Previous enrollment
    2. Current enrollment

And this change would make it:

I. [Edit Record Child Name]
  A. Enrollment and funding
  B. Has {childName}'s age group and/or site changed?
    1. New enrollment
    2. Previous enrollment
    2. Current enrollment

bamader: Ah that makes sense. That example is helpful in clarifying how we want to think of headers ad delimiting parts of a hierarchy. Since I was (mistakenly) just using it for the text bolding, I've now replaced the `<h2>`'s with a `div` whose className is the aptly labeled `text-bold`.

mmazanec22: I went through our headers with a fine toothed comb a few weeks back and this was something I changed to be more accessible and closer to the design, which does not indicate an altered font weight for this text: https://app.abstract.com/share/ee1fd572-a677-4e4a-b86e-3da923b143ed?collectionId=cf20235a-2761-4fa5-82c4-e62b60c3eb6f&collectionLayerId=90181a02-b908-4948-a5cd-c884aa8564a0&mode=build&selected=2712098272-3CCF91A4-0B55-49CF-8A48-856E37D44D8A&sha=ed049b31113448ca997e7042786d326af43d99d8

bamader: Good catch! In the original copy I looked at, it was bolded and larger, so I defaulted to a header when I shouldn't have.

src/controllers/export.ts
```diff
@@ -0,0 +1,277 @@
...
+}
+
+/**
+ * One of two ways to export a CSV of Children objects. Uses a
+ * transaction manager to find all Children who were uploaded
+ * in a particular enrollment report, sorts their properties,
+ * and returns the result.
+ * @param report
+ */
+// export async function getChildrenByReport(report: EnrollmentReport) {
```
JordanGuinn, line 59: Is this getting deleted?  Or is this in preparation for some future work?

src/routes/export.ts
```diff
@@ -0,0 +1,66 @@
...
+          'children.enrollments.fundings',
+        ],
+      });
+
+      if (!report) throw new NotFoundError();
+
+      res.send(controller.streamUploadedChildren(res, report.children));
+    } catch (err) {
+      console.error('Unable to download exported enrollment data: ', err);
+      throw new BadRequestError('Could not create spreadsheet to download');
```
JordanGuinn, line 41: Nit: Is this error (and the equivalent in `/export/roster` _really_ indicative of a bad request?  Wondering if 500 error responses might be better suited for catch-all blocks like these.

JordanGuinn (approved):
👍🏽👍🏽👍🏽👍🏽👍🏽

src/controllers/export.ts
bamader: Good catch. Yes, that's a delete candidate. Since Julia restructured enrollment report to throw out `FlattenedEnrollment`, the children objects live directly on the report now and we don't need the indirection.

src/routes/export.ts
bamader: You're probably right, especially because 500 is intended to be more of a catch all block. I'll whip up a 500 error in the middleware and then use that here instead.

PR #358, opened 2020-09-23
341 tabnav child info

mmazanec22:
Closes #341

PR #362, opened 2020-09-24
PR comments

mmazanec22:
Addresses PR comments that I missed before merging #358

### DIFF
--- a/src/entity/Child.ts
+++ b/src/entity/Child.ts
@@ -6,7 +6,7 @@ import {
   OneToMany,
 } from 'typeorm';
 import moment, { Moment } from 'moment';
-import { Length, ValidateNested, IsNotEmpty } from 'class-validator';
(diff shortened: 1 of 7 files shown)


src/entity/Child.ts
```diff
@@ -129,4 +129,7 @@ export class Child implements ChildInterface {
 
   @Column(() => UpdateMetaData, { prefix: false })
   updateMetaData: UpdateMetaData;
+
+  // Had to add this to make typescript happy, even though this implements an interface that extends ObjectWithValidationErrors ???
+  validationErrors: ValidationError[];
```
jlhogan, line 134: I believe this is b/c extending an interface w/ optional properties does not define those properties on the extending class?

jlhogan (commented):
Can you also just explain the array case for `getValidationStatusForFields` here https://github.com/ctoec/data-collection/blob/base/client/src/utils/getValidationStatus.ts#L67

jlhogan (approved):

PR #363, opened 2020-09-24
347 roster alert

### LINKED ISSUE #347: Missing info alert: I see an alert at the top of my roster when there is information missing that is needed to submit my data to OEC.
image

mmazanec22:
Closes #347

### DIFF
--- a/client/src/containers/Roster/Roster.tsx
+++ b/client/src/containers/Roster/Roster.tsx
@@ -6,6 +6,7 @@ import {
   Table,
   PlusCircle,
   Button,
+  AlertProps,
 } from '@ctoec/component-library';
 import AuthenticationContext from '../../contexts/AuthenticationContext/AuthenticationContext';
 import { Child, AgeGroup } from '../../shared/models';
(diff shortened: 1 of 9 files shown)


client/src/containers/Roster/Roster.tsx
```diff
@@ -110,6 +131,8 @@ const Roster: React.FC = () => {
           </Link>
           <CSVDownloadLink />
         </div>
+        {alertElements}
+        {/* TODO: revise accordion to have less stuff in the heading-- needs to be usable for quick navigation */}
```
jlhogan, line 135: im interested please tell me more!

jlhogan (approved):

client/src/containers/Roster/Roster.tsx
mmazanec22: Oh yeah I should have made a ticket sorry-- basically headers are used by screen readers (and search engines, but that's not really a use case for us) to understand page content and structure-- so a blind user may navigate through just the headers to get a sense of what all is on the page and what content they care about/want to delve into, the same way that sighted users pick up on font size/weight and text proximity cues for the same information.  I *think* (but someone should do some googling and verify this when they go to tackle this issue) that as we have it right now, the header for this section will be a little confusing sounding because it will include all of the text nested in this h2, incl the expand button text.  So the action item (if I'm right about this) is that we should modify the accordion to only wrap organizational text needed to make sense of the page in a header, and not all of the other text, even though they're grouped together spatially.

jlhogan: ahhh that totally makes sense. I think i forgot i had pushed stuff into a heading element! thanks for the ticket

PR #367, opened 2020-09-24
Funding types

### LINKED ISSUE #328: Change "private pay" labels to "unfunded"

### LINKED ISSUE #329: Change funding source options from acronyms to acronyms + full text: "CDC" to "CDC – Child Day Care"
Screen Shot 2020-09-15 at 4 50 08 PM

- CDC – Child Day Care
- PSR – Priority School Readiness
- CSR – Competitive School Readiness
- SS – Smart Start
- SHS – State Head Start
and
Change "funding source options" to "funding type"

bamader:
Closes #328 
Closes #329

### DIFF
--- a/client/src/components/Forms/EnrollmentFunding/Fields/Funding/NewFunding.tsx
+++ b/client/src/components/Forms/EnrollmentFunding/Fields/Funding/NewFunding.tsx
@@ -21,7 +21,7 @@ import {
   ChangeEnrollment,
 } from '../../../../../shared/payloads';
 
-const UNFUNDED = 'unfunded';
+const UNFUNDED = 'Unfunded';
 
 type FundingFieldProps<T> = {
   fundingSpaces: FundingSpace[];
@@ -80,7 +80,7 @@ export const FundingField = <
       options={[
         {
           render: (props: RadioOptionRenderProps) => (
-            <RadioButton {...props} text="Unfunded" />
+            <RadioButton {...props} text={UNFUNDED} />
           ),
           value: UNFUNDED,
         },
--- a/client/src/components/Forms/EnrollmentFunding/Fields/Funding/NewFunding.tsx
+++ b/client/src/components/Forms/EnrollmentFunding/Fields/Funding/NewFunding.tsx
@@ -21,7 +21,7 @@ import {
   ChangeEnrollment,
 } from '../../../../../shared/payloads';
 
-const PRIVATE_PAY = 'private-pay';
+const UNFUNDED = 'unfunded';
 
 type FundingFieldProps<T> = {
   fundingSpaces: FundingSpace[];
@@ -76,13 +76,13 @@ export const FundingField = <
       name="funding-source"
       legend="Funding source options"
       showLegend
-      defaultValue={PRIVATE_PAY}
+      defaultValue={UNFUNDED}
       options={[
         {
           render: (props: RadioOptionRenderProps) => (
-            <RadioButton {...props} text="Private pay" />
+            <RadioButton {...props} text="Unfunded" />
           ),
-          value: PRIVATE_PAY,
+          value: UNFUNDED,
         },
         ...fundingSourceOptions.map((fundingSource) => ({
           render: (props: RadioOptionRenderProps) => (
--- a/client/src/containers/Roster/__snapshots__/Roster.test.tsx.snap
+++ b/client/src/containers/Roster/__snapshots__/Roster.test.tsx.snap
@@ -74,12 +74,12 @@ exports[`Roster matches snapshot 1`] = `
                   <div
                     class="margin-top-2 margin-right-2 font-body-sm text-base-darker"
                   >
-                    CDC  — 1/3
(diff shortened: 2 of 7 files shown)


client/src/shared/models/FundingSource.ts
```diff
@@ -1,7 +1,7 @@
 export enum FundingSource {
-  CDC = 'CDC',
-  CSR = 'CSR',
-  PSR = 'PSR',
-  SHS = 'SHS',
-  SS = 'SS',
+  CDC = 'CDC - Child Day Care',
```
jlhogan, line 2: this totally works, and if we ever need to display just the short or long version, we can very easily split this string by `-`

jlhogan (approved):

jlhogan (changes requested):
Just realized -- changing this enum requires a change to the DB since enum types are enforced via check constraint on string values. It's not the best set up but it's the easiest way to do the least custom stuff. if you have other ideas about how this could work (or want to do a little googling and see what other folks do) thats very welcome but not necessary by any means

bamader:
Oh good catch. I didn't even think of constraint checking. I'll do a bit of research to see what other people have used for it, but I'm not gonna spend too long on it. If there's an elegant solution, great, if not, we minimize custom code at the cost of a teeny bit of string checking.

client/src/components/Forms/EnrollmentFunding/Fields/Funding/NewFunding.tsx
```diff
@@ -76,13 +76,13 @@ export const FundingField = <
       name="funding-source"
       legend="Funding source options"
       showLegend
-      defaultValue={PRIVATE_PAY}
+      defaultValue={UNFUNDED}
       options={[
         {
           render: (props: RadioOptionRenderProps) => (
-            <RadioButton {...props} text="Private pay" />
+            <RadioButton {...props} text="Unfunded" />
```
kmagida110, line 83: Naive Javascript question here but is there a reason this can't be the const UNFUNDED?

client/src/containers/Roster/__snapshots__/Roster.test.tsx.snap
```diff
@@ -262,7 +262,7 @@ exports[`Roster matches snapshot 1`] = `
               <td
                 class="font-body-2xs"
               >
-                PSR
+                PSR - Priority School Readiness
```
kmagida110, line 265: Where is the matching change for this?

kmagida110 (approved):
Approved, just a few questions for my own knowledge

client/src/components/Forms/EnrollmentFunding/Fields/Funding/NewFunding.tsx
bamader: Yep that's definitely the right move!

client/src/containers/Roster/__snapshots__/Roster.test.tsx.snap
bamader: When you say matching change, could you elaborate? This is a test snapshot that I had to update because I changed an element of the UI. It got automatically changed by the `docker-compose exec client yarn run test -u` command I mentioned in your PR.

bamader:
I did a little more research than I initially planned because it seemed like there would be some promising leads, but in the end I was pretty dissatisfied with all the solutions I find. I'm sticking with our string checks in the DB. I realized this ALSO means we need to modify the map function and a little bit of how enrollment data is processed when spreadsheets are uploaded. Since we don't want providers to have to fully type out 'CDC - Child Day Care' every time (they should be able to just do CDC), we want to make sure we modify the mapping function to not be like "oh you only typed CDC and not Child Day Care, I have no idea what this is".

bamader:
Update: DB changes have been made in constraint checks, and parsing for Source types has been updated so that providers don't have to enter in the full text in their spreadsheets.

src/controllers/enrollmentReports/map.ts
```diff
@@ -307,7 +308,8 @@ const mapEnum = <T>(
     | typeof FundingTime
     | typeof SpecialEducationServicesType,
   value: string | undefined,
-  firstLetterComparison: boolean = false
+  firstLetterComparison: boolean = false,
```
jlhogan, line 311: would be cool if this was just a regex you pass in to match against instead of `firstLetterComparison` and `isFundingSource`

jlhogan (approved):
one potential improvement to increase reusability (if we end up changing more enum values) but not a blocker

src/controllers/enrollmentReports/map.ts
bamader: I don't actually know why `firstLetterComparison` is there (maybe from some other check), but are you saying we should use a regex to decide what type of enum we're looking at?

jlhogan: yeah no ignore me -- i thought it was doing a different thing and we could just use a regex to compare more programatically. but more complicated than i thought upon further inspection

Issue #338, opened 2020-09-18
Enable differentiation of validation error display btwn Create & Edit flows 


RyanHansz:
blocked by: #365 

thoughts: create a context that uses location to determine if validations are errors (create flow -- they look red) or warnings (edit flow -- they look yellow)

For info required to enroll:
<img width="405" alt="Screen Shot 2020-09-18 at 7 59 53 AM" src="https://user-images.githubusercontent.com/63253539/93595029-f171fc80-f984-11ea-8072-f040fafc7a40.png">

RyanHansz:
Blocking:
- Name
- Date of birth
- Gender
- Site
- Organization
- Age Group
- Enrollment date
- Funding type
- Contract space
- First funding period
- Household size (unless foster)
- Annual household income (unless foster)
- Determination date (unless foster)
- Race
- Ethnicity
- Street Address (unless experiencing homelessness)
- Town
- State
- ZIP Code

mmazanec22:
@RyanHansz which non blocking fields need warning messages, and which ones are truly optional?  (Off the top of my head-- assuming that middle name and suffix are truly optional.)

Also-- do all "blocking" errors prevent form submission, or only some of them?  If some of them do and some of them don't, should we distinguish those in any way?  (Thinking about accessibility-- we probably should only use the HTML attribute required on the ones that do block submission, and consider whether moving focus to the first element is for submission-blocking fields or all fields with errors.)

jlhogan:
We need to regroup on "blocking" validations. Would it be acceptable to scrap them, and only have validations of the non-blocking variety?

RyanHansz:
I'd like to confirm this with Julie, but I believe there's a bare minimum amount of data that we need for a record to be worth collecting (like a record with no names). Since we're starting with a point in time count, it doesn't make sense to me to let people create incomplete enrollments.

jlhogan:
one way we can get to "blocking" validations with the class-validator package is to create a [validation "class"](https://github.com/typestack/class-validator#custom-validation-classes) for blocking validations, and run this validation on PUT/POST body before writing to db, and opting to fail the req + not write to DB if blocking validation is encountered. this means we don't have to check those validations when users do bulk upload, but can easily enforce them for individual record reation via API.

jlhogan:
>OEC isn't allowed to collect data on private pay children

[...]

PR #369, opened 2020-09-25
365 add child info required

### LINKED ISSUE #365: Make missing info block progress through AddChild flow
- `onSuccess` logic becomes `moveToNextStep`
- `onSuccess` just is refetch child
- refetchChild conditionally calls moveToNextStep based on if the `hasValidationErrors` func for the current step says no validation errors!
- style errors as errors instead of warning
- use focus first error

part of #338

mmazanec22:
Closes #365 

This was so much less bad than I thought it would be thanks to our chat Julia!

### DIFF
--- a/client/src/components/Forms/CareForKids/Form.tsx
+++ b/client/src/components/Forms/CareForKids/Form.tsx
@@ -17,7 +17,7 @@ export const doesC4kFormHaveErrors = (child: Child) =>
  */
 export const CareForKidsForm: React.FC<EditFormProps> = ({
   child,
-  afterDataSave: onSuccess,
+  afterDataSave,
 }) => {
   const { accessToken } = useContext(AuthenticationContext);
   const [isSaving, setIsSaving] = useState(false);
@@ -27,7 +27,7 @@ export const CareForKidsForm: React.FC<EditFormProps> = ({
   const onSubmit = (updatedChild: Child) => {
     setIsSaving(true);
     apiPut(`children/${child.id}`, updatedChild, { accessToken })
-      .then(() => onSuccess())
+      .then(() => afterDataSave())
       .catch((err) => {
         console.log(err);
       })
--- a/client/src/components/Forms/CareForKids/Form.tsx
+++ b/client/src/components/Forms/CareForKids/Form.tsx
@@ -27,7 +27,7 @@ export const CareForKidsForm: React.FC<EditFormProps> = ({
   const onSubmit = (updatedChild: Child) => {
     setIsSaving(true);
     apiPut(`children/${child.id}`, updatedChild, { accessToken })
-      .then(() => afterDataSave())
+      .then(afterDataSave)
       .catch((err) => {
         console.log(err);
       })
--- a/client/src/components/Forms/CareForKids/Form.tsx
+++ b/client/src/components/Forms/CareForKids/Form.tsx
@@ -17,7 +17,7 @@ export const doesC4kFormHaveErrors = (child: Child) =>
  */
 export const CareForKidsForm: React.FC<EditFormProps> = ({
   child,
-  onSuccess,
+  afterDataSave: onSuccess,
 }) => {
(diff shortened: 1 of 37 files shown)


client/src/components/Forms/CareForKids/Form.tsx
```diff
@@ -17,7 +17,7 @@ export const doesC4kFormHaveErrors = (child: Child) =>
  */
 export const CareForKidsForm: React.FC<EditFormProps> = ({
   child,
-  onSuccess,
+  afterDataSave: onSuccess,
```
jlhogan: nit: don't rename to onSuccess?

client/src/components/Forms/FamilyIncome/Form/index.tsx
```diff
@@ -11,13 +11,13 @@ const incomeDetFields = ['numberOfPeople', 'income', 'determinationDate'];
...
-      incomeDetFields
-    )
+        family.incomeDeterminations,
+        incomeDetFields
+      )
     : true;
 
 export const FamilyIncomeForm: React.FC<EditFormProps> = (props) => {
-  const { child, onSuccess } = props;
+  const { child, afterDataSave: afterDataSave } = props;
```
jlhogan, line 20: nit: don't need to cast to same name

client/src/containers/AddChild/AddChild.tsx
```diff
@@ -104,15 +91,41 @@ const AddChild: React.FC = () => {
...
+  useEffect(() => {
+    if (!childId) return;
+
+    apiGet(`children/${childId}`, {
+      accessToken,
+    })
+      .then((updatedChild) => {
+        updateChild(updatedChild);
+        if (!updatedChild || !steps || !indexOfCurrentStep) return;
+        // todo: make step list less opinionated
```
jlhogan, line 108: :)

client/src/containers/AddChild/NewEnrollment.tsx
```diff
@@ -45,7 +45,7 @@ export const NewEnrollment = ({
     setSaving(true);
     if (!Object.values(_enrollment).every((value) => !value)) {
       // If all of the values are null or undefined, don't block
```
jlhogan, line 47: is this still what we want to do here? I would guess that in addition to not allowing ppl to enter data that's missing fields, we don't want to let them create data without an enrollment (and _probably_ also want to enforce they add a funding) Maybe a question for @RyanHansz

client/src/containers/AddChild/NewEnrollment.tsx
```diff
@@ -45,7 +45,7 @@ export const NewEnrollment = ({
     setSaving(true);
```
jlhogan, line 45: nit: in line 40 above, use `getCurrentEnrollment` util?

client/src/utils/getValidationStatus.ts
```diff
@@ -40,7 +56,7 @@ export function getValidationStatusForField<
   if (!validationError) return;
   const { constraints } = validationError;
   return {
-    type: 'warning',
+    type: 'error',
```
jlhogan, line 59: so does this make all displayed validation errors look like errors, even in edit flow? not blocker just clarifying. And we have https://github.com/ctoec/data-collection/issues/338 to deal with display as warning/error based on where the user is interacting with the form so not something to address here regardless :)

jlhogan (approved):
🎉  woooo

client/src/utils/getValidationStatus.ts
mmazanec22: Yes it does, but this can be manually overridden.  Still think that context will be the best way to go if we decide to differentiate warning/error

client/src/containers/AddChild/NewEnrollment.tsx
mmazanec22: I'll take this out-- agree that sounded like what we were after

PR #387, opened 2020-09-25
Debug add child

mmazanec22:
Deal with errors that were preventing add child and messing with error focus

### DIFF
--- a/client/src/containers/AddChild/AddChild.tsx
+++ b/client/src/containers/AddChild/AddChild.tsx
@@ -25,7 +25,7 @@ const AddChild: React.FC = () => {
   const activeStep = hash.slice(1);
   const history = useHistory();
   const steps = listSteps(history);
-  const indexOfCurrentStep = steps.findIndex((s) => s.key === activeStep);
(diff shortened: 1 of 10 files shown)


kmagida110 (approved):
Looks good!

client/src/containers/AddChild/AddChild.tsx
```diff
@@ -102,7 +102,7 @@ const AddChild: React.FC = () => {
     })
       .then((updatedChild) => {
         updateChild(updatedChild);
-        if (!updatedChild || !steps || !indexOfCurrentStep) return;
+        if (!updatedChild) return;
```
bamader, line 105: Nit: Should we do these conditional checks before you invoke `updateChild` to modify the state? That way we don't replace anything if the server doesn't send back a correctly updated child?

bamader (approved):
LGTM. Just one question on the ordering of a null check vs updating state.

client/src/containers/AddChild/AddChild.tsx
jlhogan: we can expect that updatedChild will always have a value if we're in the `then` block

jlhogan (approved):

PR #412, opened 2020-09-28
Simpler enums

jlhogan:
simplifies enums so no DB changes are required to change enum value strings or add new values to enums (migrations would still be required to remove or rename an existing enum value)


bamader (approved):
Good find! Glad I saw this before I put up the PR for the `CareModel` enums.

PR #415, opened 2020-09-29
Define and support CareModel Enum

### LINKED ISSUE #223: add "model type" to enrollment (virtual, in person, hybrid)
"model" exists on FlattenedEnrollment, but not Enrollment.

**AC**
- Create enum for CareModel with values "In person", "Virtual", "Hybrid"
- Add `careModel` field to Enrollment, of type `CareModel`

### LINKED ISSUE #343: Add "not disclosed" option to race and ethnicity questions

### LINKED ISSUE #382: Special circumstances section: Group Disability, Dual Language and Foster youth into a section.
- All fields are required.
- rename receivingSpecialEducationServices to Special Education Services. Remove mention of IEP and IFSP in description.
- [maybe delete specialEducationServicesType]
- Add dual language learner to child model.
- Add "unknown/not disclosed" option for each. 
- Update form to look like:
Screen Shot 2020-09-28 at 4 26 37 PM

### LINKED ISSUE #383: Update homelessness form element
- make homelessness and all other booleans in template undefine-able (in template, and in downstream data models)

Field is required and link is added
Screen Shot 2020-09-25 at 4 21 12 PM

Link goes to https://eclkc.ohs.acf.hhs.gov/sites/default/files/learning-modules/homelessness-v2/module-4/story_content/external_files/HL%20Module%204%20Decision-Tool_Final%204_20_18.pdf
in a new tab.

### LINKED ISSUE #389: Update SR space types
Screen Shot 2020-09-28 at 3 48 18 PM

- Full-day "4" or "Full-day" or Full-day/Full-year" or "FD"
- School-day "5" or "School-day" or "School-day/School-year" or "SD"
- Part-day "6" or Part-day" or Part-day/Part-year" or "PD"
- Extended-day "7" or "Extended-day" or "Wraparound" or "ED"

### LINKED ISSUE #394: Remove Care 4 Kids tab on child records
Need to collect provider C4K ID as part of onboarding. 

and remove c4k bool on template and child model

### LINKED ISSUE #453: Update csvdownload link is success page to download user's entire roster (not based report id)
At least for point in time data collection, users will download their entire roster so link on success page should do that

### LINKED ISSUE #468: File download is implemented correctly – Downloading roster instead of enrollment report

bamader:
Adds a `CareModel` enum to the enrollment model and performs edits to boolean fields to allow them to be null (and to allow nulls to be interpreted as an 'Unknown/Not disclosed or collected' value). Moves several fields into a Special Circumstances page that allows children receiving disability services to have that recorded in the DB (also allows these values to be null).
Repairs CSV download link to export a provider's roster and not just an enrollment report, and updates SR funding types.

Note: I know we said we were going to stop adjusting the initial migration, but since this branch is modifying default values of fields used later on, updating the initial values (and permissions, where applicable, to allow nulls) in the first migration seemed the most appropriate.

Closes #223 
Closes #383 
Closes #343 
Closes #389 
Closes #453 
Closes #382 
Closes #468 
Closes #394


src/template/EnrollmentReportRow.ts
```diff
@@ -319,7 +320,7 @@ export class EnrollmentReportRow {
     example: 'Yes',
     section: SECTIONS.FAMILY_INFO,
   })
-  homelessness?: boolean = false;
+  homelessness?: boolean = null;
```
jlhogan, line 323: can these default to undefined or is there a specific reason we need to use null? Does ORM cast undefined and null to null in the DB?

client/src/components/Forms/FamilyAddress/Fields/Homelessness.tsx
```diff
@@ -1,20 +1,67 @@
...
+            ),
+            value: 'No',
+          },
+          {
+            render: (props) => (
+              <div>
+                <RadioButton text="Unknown/Not collected" {...props} />
+              </div>
+            ),
+            value: 'null',
```
jlhogan, line 62: should figure out if this value if used by screen readers -- if it is it should be something like `Unknown` ?

client/src/shared/models/CareModel.ts
```diff
@@ -0,0 +1,5 @@
+export enum CareModel {
```
jlhogan, line 1: does care model need a `Not disclosed option`?

client/src/components/Forms/FamilyAddress/Fields/Homelessness.tsx
```diff
@@ -1,20 +1,67 @@
...
+          else if (data == false) return 'No';
+          else return 'null';
+        }}
+        parseOnChangeEvent={(e) => {
+          if (e.target.value != 'null') return e.target.value == 'Yes';
+          else return null;
+        }}
+        inputComponent={RadioButtonGroup}
+        name="homelessness"
+        legend=""
```
mmazanec22, line 32: We cannot leave the legend blank for a fieldset.  It should be "Homelessness status" and we should remove the p tag on lines 15-17 (which was only there as a workaround for our weird single checkbox situation).

src/controllers/enrollmentReports/parse.ts
```diff
@@ -58,7 +58,7 @@ export function parseUploadedTemplate(file: Express.Multer.File) {
   const booleanProperties: string[] = [];
   const dateProperties: string[] = [];
   Object.entries(new EnrollmentReportRow()).forEach(([prop, value]) => {
-    if (typeof value === 'boolean') {
+    if (typeof value == 'boolean') {
```
mmazanec22, line 61: Why this change?

src/template/EnrollmentReportRow.ts
bamader: From my experiments and trials yesterday, we do need null instead of undefined because of the way that we want to do defaults. A migration can handle a `DEFAULT NULL` but defaulting to an undefined seemed to cause issues in the table (not sure if it was because of a constraint or the way things get cast/allocated). Happy to research more on this though.

client/src/components/Forms/FamilyAddress/Fields/Homelessness.tsx
bamader: Mmm good call. Frankly `unknown` is a better descriptor than null anyway.

client/src/shared/models/CareModel.ts
bamader: Good question...I don't actually know the answer. Should we check with Ryan? It's not hard to put one in.

client/src/components/Forms/FamilyAddress/Fields/Homelessness.tsx
bamader: Ooooh didn't know it was a weirdness hack. Can def change the legend. (I also noticed some other places there should be a legend instead of a `p` tag, so I'll swap those out as I clean up more variables).

src/controllers/enrollmentReports/parse.ts
bamader: Cuz I was experimenting with how to get the DB to accept the null and forgot to change this back!

src/template/EnrollmentReportRow.ts
jlhogan: the migration is sql, so there is no such thing as undefined. but making the DB table have a null default is separate from the default value of the mapped entity property. I was just wondering if leaving the migration and DB table as is, this entity property could just be `homelessness?: boolean` without the null default

client/src/components/Forms/FamilyAddress/Fields/Homelessness.tsx
mmazanec22: Yeah *someone* should have totally left a comment when they implemented this.  (It was me, I should have left a comment-- thanks for cleaning this up :smile: )

src/template/EnrollmentReportRow.ts
bamader: Oh my b, I thought you meant something different. I'm pretty sure it can. I only made that change because it was initially a default of false which is now different than our unknown state. Will get rid of the null!

bamader: Actually, it seems like I can't get rid of that. I've run through all of these PR review changes, and eliminating the default null causes errors when trying to parse an uploaded file. Do you think that's a problem or just a quirk that's okay to leave alone?

client/src/shared/models/CareModel.ts
```diff
@@ -0,0 +1,6 @@
...
+  inPerson = 'In-person',
```
jlhogan, line 2: I believe wejust heard from OEC about a language change here, right @RyanHansz 
they are calling "virtual" something else

src/entity/Child.ts
```diff
@@ -110,22 +112,17 @@ export class Child implements ChildInterface {
...
 
-  @Column({ default: false })
-  receivesC4K?: boolean = false;
+  @Column({ nullable: true, default: null })
+  foster?: boolean;
 
-  @Column({ default: false })
-  receivesSpecialEducationServices?: boolean;
+  @Column({ nullable: true, default: null })
+  receivesC4K?: boolean;
```
jlhogan, line 122: remove this column while you're in here making changes to the SQL? :)

jlhogan (commented):
Let's confirm with ryan where the special circumstances stuff should go. Unclear frm ticket, and looks like you've added it to it's own tab. My interpretation was that it should go in the child demographics tab. @RyanHansz 

Other than that, one small data change that I think would be grate to make along with all of these database changes  (removing `recievesC4k` from child) and one small copy change that doesn't need to be merged at the same time as these because changing the value of an enum string no longer requires a DB change!! :)

client/src/shared/models/CareModel.ts
RyanHansz: Yep its the child info tab. And I forgot to copy the definition from the zoom chat so I had to email to follow up

src/entity/Child.ts
bamader: You got it! Only reason I didn't pull it out yet was because I didn't know what provider level to put it at, but that has nothing to do with the fact that child records don't need the tab anymore!

client/src/shared/models/CareModel.ts
bamader: Whoops my bad. Back into child info they go!

src/entity/Child.ts
bamader: Should I also delete the care4kids fieldset and form? Or just comment it out and leave it in there for whatever provider level needs it?

jlhogan: delete, we are not exposing any provider-level data collection via the application

client/src/components/Forms/ChildInfo/Fields/DualLanguageLearner.tsx
```diff
@@ -0,0 +1,59 @@
...
+        else return null;
+      }}
+      inputComponent={RadioButtonGroup}
+      name="dualLanguageLearner"
+      legend="Dual Language Learner"
+      showLegend
+      options={[
+        {
+          render: (props) => (
+            <div>
```
mmazanec22, line 33: Why put the radio buttons in divs?

client/src/components/Forms/ChildInfo/Fields/Ethnicity.tsx
```diff
@@ -18,33 +18,39 @@ export const EthnicityField: React.FC = () => {
...
           render: (props) => (
             <RadioButton text="Not Hispanic or Latinx" {...props} />
           ),
-          value: 'no',
+          value: 'No',
         },
         {
           render: (props) => (
-            <RadioButton text="Hispanic or Latinx" {...props} />
+            <RadioButton text="Unknown/Not Collected" {...props} />
```
mmazanec22, line 51: Capitalization isn't consistent with other unknown/not collected value.  @RyanHansz we're using sentence casing right?  So "Unknown/not collected"?

mmazanec22: Maybe worth sticking a const somewhere in utils with this text so we don't have to retype it

client/src/components/Forms/ChildInfo/Fields/Foster.tsx
```diff
@@ -0,0 +1,59 @@
...
+  FormField,
+  RadioButtonGroupProps,
+  RadioButtonGroup,
+  RadioButton,
+} from '@ctoec/component-library';
+import { Child } from '../../../../shared/models';
+import { getValidationStatusForField } from '../../../../utils/getValidationStatus';
+
+export const Foster: React.FC = () => (
+  <>
```
mmazanec22, line 12: We don't need to wrap this in a fragment <></> component bc it no longer has sketchy nested p tag trying to look like a label!

client/src/components/Forms/ChildInfo/Fields/Race.tsx
```diff
@@ -37,20 +38,38 @@ const raceOptions: { label: string; field: RaceField }[] = [
...
+    !!record.at('nativeHawaiianOrPacificIslander') ||
+    !!record.at('white')
+  );
+};
+
 /**
  * Component for entering the race of a child in an enrollment.
  */
 export const RaceField: React.FC = () => {
+  const [notDisclosed, setNotDisclosed] = useState(true);
```
mmazanec22, line 61: This is really nicely done!

client/src/components/Forms/ChildInfo/Form.tsx
```diff
@@ -70,7 +75,12 @@ export const ChildInfoForm = ({
       <RaceField />
       <EthnicityField />
       <GenderField />
-      <FosterCheckbox />
+
+      {!hideHeader && <h2>Special circumstances</h2>}
```
mmazanec22, line 79: This would be an h3 since it's nested under child info-- and not necessary to hide bc it's not duplicative of anything (which is what's going on with the other hide header-- it duplicates the h2 provided by step list).  But really grouped form elements should go in a fieldset together with a legend.

src/entity/Child.ts
```diff
@@ -76,29 +76,31 @@ export class Child implements ChildInterface {
...
 
-  @Column({ nullable: true })
+  @Column({ nullable: true, default: false })
   @ChildRaceIndicated()
   white?: boolean;
 
-  @Column({ nullable: true })
-  @IsNotEmpty()
-  // No default value because this is a radio button
+  @Column({ nullable: true, default: true })
```
mmazanec22, line 99: Hmm, so by default someone has to uncheck this box to check another box?  Seems like they should all default to undefined maybe?

jlhogan: i think we should leave the data values as they are, because we want this to be the deafult state. I think the fix should be in the form of improving how the checkboxes work -- that is, automatically un-checking this one if any of the others are clicked, and automatically checking it if none of the others are checked. I am working on checkbox gruop refactor to make this easier, hoping to land it after the weekend.

client/src/components/Forms/ChildInfo/Fields/DualLanguageLearner.tsx
bamader: I think that's how they were in the form I copied the basic structure of the button group from. Do you think I should pull them out of the divs and just have them by themselves?

client/src/components/Forms/ChildInfo/Fields/Ethnicity.tsx
bamader: Good call. A constant stuck in client utils will resolve both of these things at once.

client/src/components/Forms/ChildInfo/Fields/Foster.tsx
bamader: YOU HAVE NO POWER HERE SKETCHINESS

client/src/components/Forms/ChildInfo/Fields/Race.tsx
bamader: Thanks 😄 ! I was actually trying to create more of the functionality Julia described below (with checking and unchecking other boxes) but this wound up creating state variables to track state variables to track state variables, so I was like ehhh something here is probably better left to a checkbox change.

src/entity/Child.ts
bamader: @mmazanec22 This value is default only in the sense that if a provider uploads data with no race fields indicated, then the `raceNotDisclosed` field is the only one that's true. If a provider uploads data where any of the race fields at all are checked, then this default of true is overwritten by the `map` file's creation of a specific child record for the data, because `map` invokes a function called `anyRaceDisclosed` which changes the value of `notDisclosed` to false if any race is provided. I was actually trying for a bit to do what Julia is describing, but it wound up involving state variables to track other state variables and devolved into a gross mess, so the fix probably is in the checkboxes themselves.

client/src/components/Forms/ChildInfo/Fields/DualLanguageLearner.tsx
mmazanec22: Yeah, unless there's some compelling layout case, we should ditch em-- no reason to nest elements if we don't need to.  Excessive nesting => slowness; inconsistent component use => hard to debug styling issues.

src/entity/Child.ts
jlhogan: extends CheckboxProps | FormFieldProps<any, CheckboxProps, any>

bamader: It seems like there's two questions there. One is why can't booleans be `undefined` instead of false. I did a lot of experimenting with how to store values and defaults in the DB since we want to allow interpreting null as something meaningful, and I found that using undefined causes problems with how the client interprets a value that gets stored. I believe (supported by a small sample of research) this is because the migration sets up the field to allow and default to null, but trying to store an `undefined` in the table doesn't register as filling that record, which causes problems when we try to display, for example, the EditRecord for a child. On the other hand, using true and false to initialize the boolean and then overwriting it with whatever value comes in from the upload plays perfectly with the system. It may not be perfect, but that was at least my understanding as I tried the different permutations.

The second question I'm seeing is does `raceNotDisclosed` actually need to be its own field, or can we just interpret the lack of all other race fields as not disclosed. I think here, too, we actually do the need field. There are a number of places in the code where having an explicit field for this property makes things consistent, neater, or easier to integrate. Our validations require explicit field names to make sure appropriate fields get checked. We use field names to generate race options on the page, extracting labels from the fields themselves. We also use the explicit field label of `raceNotDisclosed` to control the state variables that enable or disable other checkboxes. It could def be possible to rewrite these without the explicit field (though the disable control is difficult because each checkbox doesn't know about the others), but to me this seemed the easiest and most consistent way.

Thoughts?

jlhogan: sounds fine. I think this needs to get merged, since it's closing a bajillion tickets at this point. we can tweak it later if necessary

jlhogan (approved):

mmazanec22 (approved):

PR #416, opened 2020-09-29
407 org provider site

### LINKED ISSUE #407: MultiOrganization entity
Create a DB entity that relates multiple organizations

thought: should we just rename the existing "community" entity, as this is _not_ part of point in time launch and already has related permissions object etcc

Currently blocked by OEC's lack of definition.

jlhogan:
closes #407 

THe last commit is just a hand wave at what might be a good place to add convenience interpretation of the permissions from the server; whenever we get to actually implementing features relying on this stuff (deleting roster data with new upload, displaying multi-provider/multi-site rosters, creating new enrollments) I would expect the UserContext stuff to change again depending on what is actually needed and most useful!


mmazanec22 (approved):

src/controllers/users.ts
```diff
@@ -1,12 +1,14 @@
-import { User, Site, Organization } from '../entity';
+import { User, Site, Provider, Organization } from '../entity';
 import { getManager } from 'typeorm';
 
 export const addDataToUser = async (user: User) => {
-  const [sites, organizations] = await Promise.all([
+  const [sites, providers, organizations] = await Promise.all([
     getManager().findByIds(Site, user.siteIds || []),
+    getManager().findByIds(Provider, user.providerIds || []),
```
kmagida110, line 7: Random question, why are there no orgs here?

client/src/shared/models/db/Community.ts
```diff
@@ -1,7 +0,0 @@
-import { Organization } from '.';
```
kmagida110, line 1: Will we be adding a model for a geographic location once we figure out what that would be, ZIP/Town/County etc.

kmagida110 (approved):
From what I know this looks like the right structure, I have a few questions but nothing blocking.

src/controllers/users.ts
jlhogan: b/c i messed up :)

client/src/shared/models/db/Community.ts
jlhogan: Not sure, right now in the system we only are collecting regions for sites which enabled funding reporting. Ultimately if this is going to be the system of record for all thngs provider, then yes we will need to track addresses here as well. I think in a dream world, this is the system of record for enrollment info and there is a separate one for provider info, and this talks to that. But that's so many sprints into a future that i think the answer is yes we should probably capture address data for all entities ? @RyanHansz

PR #436, opened 2020-09-29
Handle jwt error for file upload

### LINKED ISSUE #209: jwt expired error
After time away, upload fails with this message again:
Screen Shot 2020-08-24 at 5 00 55 PM

mmazanec22:
Closes #209 

This does the bare minimum to address the issue.  Some open questions:
- After someone logs in, should we redirect them to getting started (current behavior) or take them back to what they were trying to access?
- Should we somehow poll for log-in-ed-ness and automatically redirect people to the logged out screen after a certain period of time?
- How far do we want to go on winged keys branding to make this whole experience less confusing?

### DIFF
--- a/client/src/containers/Upload/Upload.tsx
+++ b/client/src/containers/Upload/Upload.tsx
@@ -6,6 +6,7 @@ import { ReactComponent as Arrow } from '@ctoec/component-library/dist/assets/im
 import { apiPost } from '../../utils/api';
 import { getErrorHeading, getErrorText } from '../../utils/error';
 import { getH1RefForTitle } from '../../utils/getH1RefForTitle';
+import { handleJWTError } from '../../utils/handleJWTError';
 
 const Upload: React.FC = () => {
   // USWDS File Input is managed by JS (not exclusive CSS)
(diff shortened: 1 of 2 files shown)


jlhogan (approved):
RE: polling, i think we could do that in the auth context -- we make check and optionally do a refresh token request if necessary on every render (what exactly triggers a render in a context I do not know -- a render in a component that uses the context?) anyway maybe there's some way to determine if he token is expired in that path

RE: redirecting back to the path you were trying to access, i vaguely poked at that when i was implementing the not signed in redirect and it was one of those things that i thought would be straightforward-ish, and then turned out to not be as clear to me as i assumed it would be. that doesn't mean it's not possible, but probably (definitely) not in scope for dec launch IMO 

RE: improving ID Server UI, we discussed i think doing some is v good idea, esp before launch but i would limit how much effort we put in b/c hopefully eventually we use an OEC-maintained IdentityServer instance

bamader (approved):
LGTM. I would agree that we should send people back to the getting started page. In fact, if we're confident the JWT only occurs after a certain period of being logged in, we could even display an alert when `handleJWTError` pushes them back to the login page, something like "Your session has expired. Please login again." or something like that. Of course, if JWT isn't strictly caused by that, it might not be completely correct, but from a user perspective it might at least explain why they're being redirected.

I'll defer to Julia on the polling because I don't know much about the contexts. For the winged keys branding, it seems like one of the things Ryan has called "nice to have but not necessary," so maybe we can make a low priority ticket for it and just stick at the bottom of the backlog for after we handle other features?

RyanHansz:
Re: Winged Keys branding, changing the header to say "log in to ECE reporter" would be helpful.
Screen Shot 2020-09-30 at 11 17 56 AM

mmazanec22:
@RyanHansz yay thx for the design there, glad we're improving that screen!  I'm going to open a separate ticket for those styling changes if that's ok.  I'll look into whether passing an alert can be done easily, but my hunch is that it'll be more work than it's worth for now

Issue #350, opened 2020-09-18
Batch upload replaces roster: When I upload a spreadsheet, any data, if there is any, in my roster is replaced with the spreadsheet data.


RyanHansz:
Instead of a separate verify data table for the batch upload, replace the roster data with the spreadsheet data that has just been uploaded.

If the user is uploading a spreadsheet when data already exists in their roster, then show a confirmation modal telling them that all the data will be overwritten. 

![6 2 Batch Upload Messages Modal](https://user-images.githubusercontent.com/63253539/94480576-81dde780-01a4-11eb-8c62-0e0bc2929b0e.png)
![6 2 Batch Upload Messages Modal Copy](https://user-images.githubusercontent.com/63253539/94480575-81dde780-01a4-11eb-8da4-7fbcf1d54b6e.png)

jlhogan:
need to make sure we have a good definition of what is going to be replaced -- just the data that was uploaded by this specific user? Or all the data that this user has access to ? This may be an implementation detail, but requires understanding of how the user accounts are expected to exist. so the actual question is: will the sets of data each account has access to be mutually exclusive for the initial point in time collection?

RyanHansz:
All sites/orgs the user has permissions for. 

- So if I'm a subcontractor that only has access to one site of an org, I would only replace my data for that site (and I wouldn't see data for the rest of the Org).
- If I have access to multiple orgs, I would replace the data for all of them.

RyanHansz:
@jlhogan this one needs a new size estimate

PR #437, opened 2020-09-29
350 replace roster

### LINKED ISSUE #396: Last age group in roster gets cut off
Screen Shot 2020-09-28 at 11 45 59 AM

jlhogan:
TODO: Backend - make the server parse & use the `overwriteSite` id
query params in enrollment report upload

addresses #350  but does not close; still need to do UI stuff
closes #396 
implements mutually exclusive checkbox logix that _may_ apply to https://github.com/ctoec/data-collection/issues/343 (re race not disclosed option) BUT may totally not given form fields


client/src/containers/Upload/CheckReplaceData.tsx
```diff
@@ -0,0 +1,127 @@
...
+          onChange={(e) => {
+            if (e.target.checked) {
+              setSiteIdsToReplace([]);
+            }
+          }}
+        />
+      ),
+    },
+    ...(user?.sites || []).map(
+      (site) =>
```
bamader, line 55: Do we want to allow the user to choose which sites get replaced? I personally really like this functionality, just checking from a design perspective since we initially said just replace everything. This feels like an elegant addition and not something that should cause an issue, though.

client/src/containers/Upload/Upload.tsx
```diff
@@ -24,25 +26,56 @@ const Upload: React.FC = () => {
...
         });
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [file]);
+  }, [postUpload]);
+
+  useEffect(() => {
+    // wait until we know if the user already has a roster or not
+    // before allowing them to submit data
+    if (userRosterCount === undefined) return;
```
bamader, line 69: Potentially dumb question, but should this be just `==` instead of `===`? The distinction was giving me some trouble in the backend for interpreting nulls as booleans, so wondering if that might cause a similar issue for undefined. It might be a totally separate thing, though.

bamader (approved):
Front end stuff looks good. I particularly like allowing the user to select which sites they'd like to replace for. Feels ~fancy~ 🧐

client/src/containers/Upload/CheckReplaceData.tsx
jlhogan: yup, the AC of the ticket was updated to allow user to select !

client/src/containers/Upload/Upload.tsx
jlhogan: `===` is strict equal, so `undefined === false` is `false`. If you want to check if something is undefined or false, you can check `![the thing that might be undefined or false`. The default linter that the react app uses generates warnings for double `==` so i have opted to use `===` always unless I really truly need to

mmazanec22 (approved):

PR #438, opened 2020-09-29
338 edit validations

### LINKED ISSUE #338: Enable differentiation of validation error display btwn Create & Edit flows

mmazanec22:
Closes #338 

A lot of refactoring for a small win but hopefully this makes things simpler and more consistent!


client/src/components/Forms/commonFormStepInfo.tsx
```diff
@@ -0,0 +1,58 @@
+import {
```
jlhogan, line 1: can we use maybe like `section` instead of `step`, since they're not steps in the edit world?
so this could be like `sectionInfo.tsx` or whatever

client/src/components/Forms/commonFormStepInfo.tsx
```diff
@@ -0,0 +1,58 @@
...
+  doesChildIdFormHaveErrors,
+  doesChildInfoFormHaveErrors,
+  doesEnrollmentFormHaveErrors,
+  doesFamilyAddressFormHaveErrors,
+  doesFamilyIncomeFormHaveErrors,
+} from '.';
+import { Child } from '../../shared/models';
+import { doesC4kFormHaveErrors } from './CareForKids/Form';
+
+export const TAB_IDS = {
```
jlhogan, line 11: and then maybe instead of `TAB_IDS` call these `SECTION_KEYS` ?

jlhogan (approved):
I LOVE this refactor!!!!!! one nit re: naming of the shared form stuff

bamader (approved):
LGTM! Especially like the standardization of passing in a child into the 'does X form have validation errors'.

Issue #397, opened 2020-09-28
Roster back button should always go to getting started


RyanHansz:
Change button copy to read "back to getting started"

mmazanec22:
Closed by #446

PR #446, opened 2020-09-30
Move alert elements and change roster back button

### LINKED ISSUE #395: Alerts should always show below the header.

mmazanec22:
Closes #395 and #397

### DIFF
--- a/client/src/containers/AddChild/AddChild.tsx
+++ b/client/src/containers/AddChild/AddChild.tsx
@@ -138,8 +138,8 @@ const AddChild: React.FC = () => {
     <div className="grid-container">
       <div className="margin-top-4">
         <BackButton />
-        {alertElements}
         <h1 ref={h1Ref}>Add a child record</h1>
+        {alertElements}
         <p className="usa-hint">
           Information is required unless otherwise specified.
         </p>
--- a/client/src/containers/Roster/Roster.tsx
+++ b/client/src/containers/Roster/Roster.tsx
@@ -12,7 +12,7 @@ import AuthenticationContext from '../../contexts/AuthenticationContext/Authenti
 import { Child, AgeGroup } from '../../shared/models';
 import { apiGet } from '../../utils/api';
 import { tableColumns } from '../Roster/TableColumns';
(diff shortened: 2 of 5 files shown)


client/src/containers/Roster/Roster.tsx
```diff
@@ -24,7 +24,6 @@ import pluralize from 'pluralize';
 const MAX_LENGTH_EXPANDED = 50;
 
 const Roster: React.FC = () => {
-  const { goBack } = useHistory();
```
kmagida110, line 27: For my own knowledge, is useHistory built in and how far back does it go?

kmagida110 (approved):
LGTM!

client/src/containers/Roster/Roster.tsx
mmazanec22: Use history is a built in hook to react-router-dom.  This func just goes back to the last page, whatever it was.  More info about react router dom history here: https://reactrouter.com/web/api/history

client/src/containers/AddChild/AddChild.tsx
```diff
@@ -138,8 +138,8 @@ const AddChild: React.FC = () => {
     <div className="grid-container">
       <div className="margin-top-4">
         <BackButton />
-        {alertElements}
         <h1 ref={h1Ref}>Add a child record</h1>
+        {alertElements}
```
bamader, line 142: Just curious, Is it better from an accessibility standpoint to display alerts under the header? I would've thought having them higher up on the page / more towards the top would make them more visually apparent.

bamader (approved):
LGTM!

client/src/containers/AddChild/AddChild.tsx
mmazanec22: Hmm well they all have an h2 element as the header for findability and have an alert role assigned if they indicate an error.  I imagine that being all nested under the h1 like this works because then they're a) following expected header order b) still prominent on the page and c) consistent.

bamader: Gotcha, makes sense to me!

PR #447, opened 2020-09-30
390 error message copy

### LINKED ISSUE #390: Update form error message copy.
[Field name] is required for OEC reporting.

mmazanec22:
Closes #390 
Depends on https://github.com/ctoec/component-library/pull/73

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -1,70 +1,70 @@
 {
-  "name": "client",
-  "version": "0.1.0",
-  "private": true,
-  "dependencies": {
-    "@ctoec/component-library": "https://github.com/ctoec/component-library#base",
-    "@openid/appauth": "^1.2.7",
-    "@testing-library/dom": "^7.24.1",
-    "@testing-library/jest-dom": "^4.2.4",
-    "@testing-library/react": "^10",
-    "@testing-library/user-event": "^7.1.2",
-    "@types/classnames": "^2.2.10",
-    "@types/content-disposition": "^0.5.3",
-    "@types/file-saver": "^2.0.1",
-    "@types/jest": "^24.0.0",
-    "@types/lodash": "^4.14.161",
-    "@types/node": "^12.0.0",
-    "@types/pluralize": "^0.0.29",
-    "@types/query-string": "^6.3.0",
-    "@types/react": "^16.9.0",
-    "@types/react-dom": "^16.9.0",
-    "@types/react-router-dom": "^5.1.5",
-    "class-validator": "^0.12.2",
-    "content-disposition": "^0.5.3",
-    "file-saver": "^2.0.2",
-    "idx": "^2.5.6",
-    "lodash": "^4.17.20",
-    "moment": "^2.27.0",
-    "node-sass": "yarn:sass",
-    "pluralize": "^8.0.0",
-    "query-string": "^6.13.1",
-    "react": "^16.13.1",
-    "react-dom": "^16.13.1",
-    "react-markdown": "^4.3.1",
-    "react-perfect-scrollbar": "^1.5.8",
-    "react-router-dom": "^5.2.0",
-    "react-scripts": "3.4.1",
-    "sass": "^1.26.10",
-    "typescript": "^3.9.5",
-    "uswds": "^2.8.0"
-  },
-  "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
-    "test": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen",
-    "eject": "react-scripts eject"
-  },
-  "eslintConfig": {
-    "extends": "react-app"
-  },
-  "browserslist": {
-    "production": [
-      ">0.2%",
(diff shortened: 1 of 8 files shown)


client/package.json
```diff
@@ -1,70 +1,70 @@
...
-    "jest-axe": "^4.0.0",
-    "jest-canvas-mock": "^2.2.0",
-    "jest-environment-jsdom-sixteen": "^1.0.3",
-    "jest-fetch-mock": "^3.0.3"
-  }
+	"name": "client",
+	"version": "0.1.0",
+	"private": true,
+	"dependencies": {
+		"@ctoec/component-library": "https://github.com/ctoec/component-library#field-status-props",
```
bamader, line 6: Once we merge in the Component Library field status props functionality, we can change this back to base, right?

client/yarn.lock
```diff
@@ -1166,9 +1166,9 @@
   resolved "https://registry.yarnpkg.com/@csstools/normalize.css/-/normalize.css-10.1.0.tgz#f0950bba18819512d42f7197e56c518aa491cf18"
   integrity sha512-ij4wRiunFfaJxjB0BdrYHIH8FxBJpOwNPhhAcunlmPdXudL1WQV1qoP9un6JsEBAgQH+7UXyyjh0g7jTxXK6tg==
 
-"@ctoec/component-library@https://github.com/ctoec/component-library#base":
+"@ctoec/component-library@https://github.com/ctoec/component-library#field-status-props":
   version "0.0.31"
-  resolved "https://github.com/ctoec/component-library#99f2350d7ec77830852239e8d1d6dde86ab2f0f8"
+  resolved "https://github.com/ctoec/component-library#bc7dc529c1e691e3a06538b19db404ee35b0fc17"
```
bamader, line 1171: Same comment here as in `package.json`

client/src/utils/getValidationStatus.ts
```diff
@@ -24,20 +27,31 @@ export type ValidationStatusOptions = {
...
 
+export function drillReactNodeForText(inputNode: ReactNode | string): string {
+  if (typeof inputNode === 'string') {
+    return inputNode;
+  } else if (React.isValidElement(inputNode)) {
+    return drillReactNodeForText(inputNode.props.children);
+  }
+  return '';
+}
+
```
bamader, line 38: For my own understanding, what does this function gain us? I see it's recursively drilling down into an element until we find some string, but is this string a property name of the object? Or something else displayed by the validations?

bamader (approved):
Looks good aside from some changes to `yarn.lock` and `package.json` (basically just changing back the branches).

client/package.json
mmazanec22: Yes!  That def needs to be done before this is merged

client/src/utils/getValidationStatus.ts
mmazanec22: The label for a form input can either be a string or a react node.  Theoretically that react node could contain more nested react nodes, too.  This gets us down to the actual text of the label so that we can use it in the error message.

bamader: Ooooh so this retrieves the actual label that we've been setting up in all of our form elements so we can show the user where the thing broke

JordanGuinn (approved):

client/src/utils/getValidationStatus.ts
```diff
@@ -95,11 +111,7 @@ export function getValidationStatusForFields<
     fields.includes(v.property)
   );
   if (!validationErrors.length) return;
-  const message = validationErrors
-    .map((v: ValidationError) =>
-      v.constraints ? Object.values(v.constraints).join(', ') : ''
-    )
-    .join(', ');
+  const message = 'This information is required for OEC reporting.';
```
kmagida110, line 114: What is the case where we'd ever get down to this?

client/src/utils/getValidationStatus.ts
```diff
@@ -24,20 +27,31 @@ export type ValidationStatusOptions = {
   message?: string;
 };
 
+export function drillReactNodeForText(inputNode: ReactNode | string): string {
+  if (typeof inputNode === 'string') {
+    return inputNode;
```
kmagida110, line 32: Is the inputNode always going to be properly capitalized or should we do that here or in the message that calls this?

kmagida110 (approved):
A few non-blocking questions

client/src/utils/getValidationStatus.ts
mmazanec22: We'd use the default message if a dev forgot to manually pass one in.  It might be better to just force a manual one or use the fieldset legend somehow

client/src/utils/getValidationStatus.ts
mmazanec22: It should always be if we're using sentence casing for labels

kmagida110: Out of curiosity, is that enforced programmatically or are we just avoiding it?

PR #459, opened 2020-10-01
317 add record

### LINKED ISSUE #317: Only allow enrollments for organizations users have access to.
For users with multi-org access, add a dropdown to the add enrollment button that makes the user choose organization before starting enrollment. 

Add a down-carat to the add enrollment button
Screen Shot 2020-09-18 at 4 07 27 PM

mmazanec22:
Closes #317


client/src/components/AddRecordButton.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import {
+  ButtonWithDropdown,
+  PlusCircle,
+  TextWithIcon,
+} from '@ctoec/component-library';
+import { Organization } from '../shared/models';
+import { Link, useHistory } from 'react-router-dom';
+
+// TODO: change after org => providers PR is merged
```
jlhogan, line 10: ok so i marked that PR as draft because the data Julie sent complicated how we understood the provider data model. That pr is on hold, feel free to `s/provider/organization` here b/c i'm not sure where things will end up

client/src/components/AddRecordButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const AddRecordButton = ({
+  providers,
+}: {
+  providers?: Organization[];
+}) => {
+  const history = useHistory();
+
+  if (!providers) {
+    return <></>;
```
jlhogan, line 19: we should implement and error boundary before launch and throw errors consistently instead of returning empty fragments (i am the biggest culprit here!!! but just saying this "out loud" so i remember to get it scoped)

client/src/components/AddRecordButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  return (
+    <ButtonWithDropdown
+      text={<TextWithIcon Icon={PlusCircle} text="Add a record" />}
+      id="select-organization"
+      options={providers.map((p) => ({
+        text: p.providerName,
+        // TODO: better to have button that navigates or links with same href but different behavior?
+        href: 'create-record',
```
jlhogan, line 42: does this need a slash?

jlhogan: oh, does this href not actually do navigation, just enables the element to get focus from keyboard?

jlhogan (approved):

client/src/components/AddRecordButton.tsx
mmazanec22: It does need a slash, thanks for pointing that out!  I'm torn on whether this should actually be a button that navigates (bad) or a link that has the same destination but different text (also bad).  Relevant q that has come up after thinking: are we making org (or provider or whatever) nullable on child?  Because then we can just ditch this and move it all into the bottom of the add child stuff like @RyanHansz originally wanted

jlhogan: i would v prefer not to

client/src/components/AddRecordButton.tsx
mmazanec22: It exists in component library so we really just have to start using it!

client/src/components/AddRecordButton.tsx
jlhogan: one workaround from an accessibility standpoint is to add orgId to the path so that there won't be duplicate destinations in links with different text! or, because we'll still be passing the organization in state, could even add the org name to the path if that is better from an accessibility standpoint?

kmagida110 (approved):

PR #463, opened 2020-10-01
450 access control batch upload

### LINKED ISSUE #450: Check org/site access on batch upload
when uploaded data is being parsed, check that user has access to the matched org  / site as follows:

orgs:
- if user only has access to one org, just use that and don't both to look up
- if user has access to multiple orgs and >0 rows have an org that does not match an org they have access to, then return an error saying "Organization not found"
- when UI receives this error, it displays a list of valid organization names to the user 
Screen Shot 2020-09-30 at 5 29 37 PM

sites:
- when any row has site that does not match what the user has access to, leave it blank and let the user fix that via UI (batch edit)
- make sure site is editable on existing enrollment but only when it does not have a value (currently there is no site field in the edit enrollment form -- needs to be added conditionally)

### LINKED ISSUE #461: fix flex wrap on roster section sub-headers

jlhogan:
closes #450 
closes #461

### DIFF
--- a/client/src/containers/Roster/TableColumns.tsx
+++ b/client/src/containers/Roster/TableColumns.tsx
@@ -26,7 +26,7 @@ export const tableColumns: (_?: boolean) => Column<Child>[] = (
               pathname: `/edit-record/${row.id}`,
             }}
           >
-            {row.lastName}, {row.firstName}{' '}
+           {row.lastName}, {row.firstName}{' '}
           </Link>
         </th>
       ),
--- a/src/controllers/enrollmentReports/map.ts
+++ b/src/controllers/enrollmentReports/map.ts
@@ -1,4 +1,4 @@
-import { getManager } from 'typeorm';
+import { getManager, EntityManager } from 'typeorm';
 import {
   Child,
   Family,
@@ -9,6 +9,7 @@ import {
   Organization,
   FundingSpace,
   ReportingPeriod,
+  User,
 } from '../../entity';
 import {
   Gender,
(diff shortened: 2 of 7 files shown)


jlhogan:
oops never requested reviews on this hahahahaaaaaaaaaaaaaaaa

client/src/containers/Roster/TableColumns.tsx
```diff
@@ -28,8 +28,8 @@ export const tableColumns: (_?: boolean) => Column<Child>[] = (
               }}
             >
               {row.lastName}, {row.firstName}{' '}
-              {(!!row.validationErrors && row.validationErrors.length == 0) ||
-                row.validationErrors == undefined ||
+              {row.validationErrors &&
```
JordanGuinn, line 31: +100 for this simplification 😎

src/controllers/enrollmentReports/map.ts
```diff
@@ -22,65 +23,130 @@ import {
...
  */
-const mapOrganization = (source: EnrollmentReportRow) => {
-  return getManager().findOneOrFail(Organization, {
-    where: { providerName: source.providerName },
-  });
+const lookUpOrganization = (
+  source: EnrollmentReportRow,
+  organizations: Organization[]
+) => {
+  if (organizations.length === 1) return organizations[0];
```
JordanGuinn, line 116: Is this necessary?  Do we want to blindly return the first organization if the org list supplied is a single element, regardless of whether or not it matches the enrollment?

src/controllers/enrollmentReports/map.ts
```diff
@@ -22,65 +23,130 @@ import {
...
+      const child = await mapAndSaveRow(
+        transaction,
+        rows[i],
+        organizations,
+        sites
+      );
+      children.push(child);
+    } catch (err) {
+      if (err instanceof ApiError) throw err;
+      console.log('Unable to parse row, dropping: ', JSON.stringify(rows[i]));
```
JordanGuinn, line 50: So if I'm parsing this correctly:

- Cancel this _whole_ procedure if an API error is thrown
- Simply stop processing this row if some other type of error is thrown

Is that right?

JordanGuinn (approved):
👍🏽👍🏽👍🏽👍🏽👍🏽👍🏽

src/controllers/enrollmentReports/map.ts
jlhogan: if the user adding the data has access to only one organization, we assume all data they enter is for that organization.

src/controllers/enrollmentReports/map.ts
jlhogan: yes -- ApiError is thrown when an unacceptable / unresolvable situation is encountered (no identifiable organization). Otherwise, we just drop the row. Eventually we probably want to let the user know about this. Hopefully it doesnt actually happen

mmazanec22 (approved):

src/controllers/enrollmentReports/map.ts
```diff
@@ -22,65 +23,130 @@ import {
 } from '../../../client/src/shared/models';
 import { FUNDING_SOURCE_TIMES } from '../../../client/src/shared/constants';
 import { EnrollmentReportRow } from '../../template';
+import { BadRequestError, ApiError } from '../../middleware/error/errors';
+
+export const mapAndSaveRows = async (
+  transaction: EntityManager,
```
bamader, line 29: What are the chances that wrapping everything up in a single transaction helps the incorrect count for large rosters? It's probs going to help boost speed for large rosters (since we don't have to query out over and over), but I wonder if for large rosters there are some things that aren't getting executed for some reason and that's leading to an improper count.

src/controllers/enrollmentReports/removeExisting.ts
```diff
@@ -10,10 +10,11 @@ import { getManager, In } from 'typeorm';
...
  */
 export const removeExistingEnrollmentDataForUser = async (
+  transaction: EntityManager,
   user: User,
   siteIdsToReplace: number[]
 ) => {
-  const allChildren = await getManager().find(Child, {
+  const allChildren = await transaction.find(Child, {
     relations: ['enrollments'],
     where: { organization: { id: In(user.organizationIds) } },
```
bamader, line 19: This may be an edge case, but suppose there was a family that had only one child enrolled in a program. If we batch upload and replace our existing data, that family no longer has any children enrolled. Should we also then load relations like family to delete as well? I guess eventually we'll want to do probabilistic checking of new data rather than completely erasing everything that's already there, so maybe not.

bamader (approved):
LGTM!

src/controllers/enrollmentReports/map.ts
jlhogan: hm i'm not sure I understand this question. 

Wrapping everything in a database transaction ensures that it will all get rolled back if any errors are thrown. 
Still have to do the same number of queries.
Not likely that some things are arbitrarily not being executed, and very improbably that that would have anything to do with transactions.

PR #464, opened 2020-10-02
Add permission gate to sites and funding spaces when changing child enrollments

### LINKED ISSUE #234: User Permissions: user can only add/change/delete data about the entities they have permissions for
- on entity create, check org id (or site id for enrollments)
- on entity update:
  - check that org/ site ids user has access to

JordanGuinn:
Closes #234.

This PR ensures that when someone attempts to change a child's enrollment, the sites and funding spaces they have available are only pulled from the child's associated organization.  The sites and funding spaces made available during initial child submission into the system are already limited to the organizations the user has access to.

Let me know if I'm missing anything here, or I've misunderstood how our permissions are intended to work.


bamader:
General question: can a child change the organization they're associated with via changing an enrollment? It makes perfect sense to me that the initial submission options are limited, but if it's possible to change a child's organization I'm wondering if we might want to allow full site and funding space options but just give a warning if a user attempts to pick one they don't normally have access to? Or maybe this is me completely misunderstanding what these permissions are supposed to do.

bamader (approved):
LGTM!

JordanGuinn:
@bamader Per slack conversation, looks like the answer to that question for now is no, but we may explicitly revisit that concept at some point down the line.

src/controllers/children.ts
```diff
@@ -32,13 +37,50 @@ export const getChildren = async (user: User) => {
...
+      'Child either does not exist, or user does not have permission to modify'
+    );
+    throw new NotFoundError();
+  }
+
+  await getManager().save(getManager().merge(Child, child, update));
+};
+
+export const deleteChild = async (id: string, user: User) => {
+  const readOrgIds = await getReadAccessibileOrgIds(user);
```
mmazanec22, line 61: Should we also add something like this to create child?

JordanGuinn: Yes!

client/src/hooks/useFundingSpaces.ts
```diff
@@ -3,12 +3,18 @@ import { FundingSpace } from '../shared/models';
...
 
-export function useFundingSpaces() {
-  // Get site options for new enrollments
+export function useFundingSpaces(organizationId?: number) {
   const { accessToken } = useContext(AuthenticationContext);
   const [fundingSpaces, setFundingSpaces] = useState<FundingSpace[]>([]);
+
   useEffect(() => {
-    apiGet('funding-spaces', { accessToken }).then((_fundingSpaces) =>
+    let path: string = 'funding-spaces';
```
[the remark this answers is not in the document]
JordanGuinn: Yep!  Will remove once this is updated to use the cache

mmazanec22 (approved):

PR #493, opened 2020-10-02
Batch edit and add data cache

### LINKED ISSUE #301: data cache?
contexts or perhaps https://lostpebble.github.io/pullstate/docs/quick-example as lighter weight alternatives to redux

### LINKED ISSUE #351: Batch-Edit: For records with needed or missing info, I can quickly add all of the needed information.
**AC**:
- add title to BatchEdit container
- add back to roster button to BatchEdit container
- add SideNav to BatchEdit container, populated with all records that need information
  - add option to get `/children?needInformation=true` 
 OR
  - create a CacheContext, that will store user's records. popuated on initial GET for Roster, used in BatchEdit. Means we will have to update the data in that context whenever we update an record, but that's very doable especially because we'll be re-fetching each updated record in a single spot
    - do NOT implement updating the cache, having roster read from data context, or having single edit-record or add record update the data context. All of that will happen in a separate tasks if we decide to go this route here. 
- Do not add item content (forms) in this ticket, that is covered by #474  Empty <></> content is fine for now
- Add success content with back to roster button as "noSelectedItemContent" in SideNav

Part of #309 
Screen Shot 2020-09-21 at 9 54 32 AM

jlhogan:
closes #351 
closes #301 

very interested on folks thoughts on this data cache. The hope is to not reinvent too many wheels, but also give us the ability to persist data without redux. Can add sites, fundingspaces, reportingperiods here as well. 

TODO: update batch edit tests


mmazanec22 (commented):
I'm into it!  We should add reporting periods and funding spaces-- rn they're hooks and this would be a better way to do that (lol sorry I just saw that you also said that)

client/src/contexts/DataCacheContext/DataCacheContext.tsx
```diff
@@ -0,0 +1,62 @@
...
+  records: T[];
+  loading: boolean;
+};
+type ReadWriteDataCache<T> = ReadOnlyDataCache<T> & {
+  addOrUpdateRecord: (record: T) => void;
+};
+
+export type DataCacheContextType = {
+  children: ReadWriteDataCache<Child>;
+};
```
bamader, line 16: Why not just use a type `ReadWriteDataCache<Child>` directly? Does this extra type add functionality?

client/src/containers/BatchEdit/BatchEdit.tsx
```diff
@@ -1,11 +1,121 @@
...
+    if (activeRecordIdx === allRecords.length - 1) {
+      // Then look for the first record that is still missing info
+      const firstStillMissingInformationRecordIdx = allRecords.findIndex(
+        hasValidationError
+      );
+
+      setActiveRecordIdx(
+        // If a no records are missing info, then set active record to none
+        firstStillMissingInformationRecordIdx < 0
+          ? undefined
```
bamader, line 48: Does setting the active record index to `undefined` just make it so that the side nav isn't open to any record in particular? And it should only be undefined when everything has the correct information and is ready to submit, right?

client/src/contexts/DataCacheContext/DataCacheContext.tsx
```diff
@@ -0,0 +1,62 @@
...
+      apiGet('children', { accessToken })
+        .then((_children) => {
+          if (_children) setChildren(_children);
+        })
+        .finally(() => {
+          setChildrenLoading(false);
+          setHaveFetchedChildren(true);
+        });
+    }
+  }, [accessToken]);
```
bamader, line 55: It took me a sec to realize what this actually does, but this means that we don't have to keep loading the same child records on like every page anymore. Is the idea to eventually refactor all `useEffect` calls which exist to fetch kids with this cached information?

bamader (approved):
LGTM! One question (which I think is a yes but just making sure). In slack, you gave some links to cache libraries that you said were a bit much for our purposes. The caching context you've developed here is intended to be used in place of those libraries, correct?

JordanGuinn (commented):
The new caching mechanism looks good to me 👍🏽👍🏽👍🏽👍🏽 The one thing that I was _initially_ gonna ask about was the potential for conflicting data if multiple users for the same provider attempted to update the same child record (or set of child records) simultaneously, but I then realized that's almost _assuredly_ not a use case we're expecting, especially this early on.

client/src/contexts/DataCacheContext/DataCacheContext.tsx
jlhogan: so i can add other data collections to the cache :) see updates

client/src/containers/BatchEdit/BatchEdit.tsx
jlhogan: yes !

client/src/contexts/DataCacheContext/DataCacheContext.tsx
jlhogan: we will need to fetch the child after it is updated in the server side, so we can get those changes here but will need make sure that in addition to fetching that child, we update that child in the local cache

jlhogan:
>conflicting data if multiple users for the same provider attempted to update the same child record (or set of child records) simultaneously, but I then realized that's almost assuredly not a use case we're expecting, especially this early on

my vote would be do a simple first write wins situation and just check that like the updated at timestamp on the thing you've sent === the update at in the db. but yes, not something to worry about for now

jlhogan:
> One question (which I think is a yes but just making sure). In slack, you gave some links to cache libraries that you said were a bit much for our purposes. The caching context you've developed here is intended to be used in place of those libraries, correct?

yes, i went with this over existing libraries b/c they seemed to be unnecessarily smart for our needs

client/src/contexts/DataCacheContext/DataCacheContext.tsx
bamader: Ah right, the `refetchChild` methods still need to be called to make sure local stuff stays in sync with the back end. Do you think that's easiest to put here, or just keep the different refetch calls on the various forms?

client/src/utils/formatters/name.ts
```diff
@@ -0,0 +1,15 @@
+import { Child } from '../../shared/models';
+
+export const nameFormatter = (
```
mmazanec22, line 3: Yay

mmazanec22 (approved):

PR #503, opened 2020-10-06
Capture submissions to OEC

### LINKED ISSUE #332: Submit to OEC: I can submit my data for a given point of time to OEC.
When a user hits the "submit to OEC" button on their roster and there are no blocking validations. 
- create an OEC Report object
- has organizationId, updateMetadata, submittedAt timestamp

JordanGuinn:
Closes #332. 

The OEC Report object has been created according to the exact specs of the ticket.  Let me know if there are any additional data points (provider? number of children? etc) that we want to capture and I'll update this accordingly.


bamader (approved):
This looks good to me! I'm wondering if we want to add `provider` to the `oecReport` object, too, so that when we show the view of all providers and if they've submitted or not, we can do that just by checking if there's a report object saved in the DB associated with that provider. I think it could be interesting to store other values in the report object (such as number of children submitted, or other things that could be used to display notification messages such as 'Successfully submitted data for XXX children' or things like that), but to me the provider would be the top of that list.

Thoughts?

src/controllers/oecReport.ts
```diff
@@ -0,0 +1,16 @@
...
+import { getManager } from 'typeorm';
+import { OECReport as OECReportInterface } from '../../client/src/shared/models';
+
+export async function createOecReport(
+  user: User,
+  organizationId: number
+): Promise<void> {
+  const report: OECReportInterface = await getManager().create(OECReport, {
+    organizationId,
+    updateMetaData: {
```
jlhogan, line 11: note to selves: thisssssssss needs to happen everywhere!
express doesn't have a global per-request context that would enable us to do this without having to pass the user in to every controller and remember to pass it along every time unless we use something like https://github.com/pulasthibandara/global-request-context

but also there aren't that many places we pass the user in, so could just do it like this everywhere.

src/migration/1601967024917-AddOecReport.ts
```diff
@@ -0,0 +1,21 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddOecReport1601967024917 implements MigrationInterface {
+  name = 'AddOecReport1601967024917';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
```
jlhogan, line 7: hope getting this migration down to necessary parts wasn't too awful :(

jlhogan (approved):

src/migration/1601967024917-AddOecReport.ts
JordanGuinn: Nah - 'twas super easy, actually 🙂

src/controllers/oecReport.ts
JordanGuinn: In an **extremely** far off future, we could create an abstract base controller class that pulled the meaningful bits out of the req/res, and made 'em available to all subclass controllers that extended it.  Cuz in reality, all controllers _really should_ have access to the corresponding request context that invoked it (in my opinion), but that would require a whooooooole bunch of refactor that we prolly don't quite have time for right now 😅

JordanGuinn: TL;DR agreed, this is def the easier thing to do for the time being

JordanGuinn:
@bamader That was kinda my thinking as well (re: capturing `provider` in the report); seems like just another meaningful piece of metadata to hang on to here.  @jlhogan how do you feel about adding `provider` to the `OECReport` entity, along with any other potentially beneficial data points outside of org, user and date?

PR #522, opened 2020-10-07
Record bug fixes

### LINKED ISSUE #483: Adding a record has no finish button
Currently:
Screen Shot 2020-10-02 at 2 22 51 PM

Desired (add to bottom)
95211111-2a4e0600-07ba-11eb-9ce6-8a6d8395996a

### LINKED ISSUE #495: Back button on submission complete page goes to a page not found

### LINKED ISSUE #507: Race should be empty by default when adding a new record
Currently, the record defaults to unknown/not collected and no race options are selectable.
Screen Shot 2020-10-06 at 1 59 41 PM

Ideally, no option would be chosen until the user selects one. 
Screen Shot 2020-10-06 at 2 02 05 PM

bamader:
Fixes up bug tickets in the current sprint. Repairs the back button the success page to point to Roster instead of the no longer existing CheckData page. Changes the default Race field when creating a new child to be nothing (including _not_ selecting Not Disclosed) while leaving default race parsing untouched. Adds a button to complete the Add Record process.

Closes #507 
Closes #483 
Closes #495


client/src/components/Forms/EnrollmentFunding/Fields/Funding/NewFunding.tsx
```diff
@@ -87,8 +89,7 @@ export const NewFundingField = <
               <ContractSpaceField<T>
                 ageGroup={enrollment.ageGroup}
                 fundingSource={fundingSource}
-                // see comment on line 56
```
jlhogan, line 90: this comment makes me expect something to happen on line 56 with this change as well?

client/src/containers/AddChild/AddChild.tsx
```diff
@@ -86,16 +88,10 @@ const AddChild: React.FC = () => {
 
     const moveToNextStep = () => {
       if (indexOfCurrentStep === steps.length - 1) {
-        history.push('/roster', {
```
jlhogan, line 89: confirmed in https://github.com/ctoec/data-collection/issues/483#issuecomment-704946476 that this change is not needed -- the existing behavior is desired behavior, bug was from older version of app. ( I had a feeling ryan wouldn't be adding in unnecessary button clicks willy nilly)

client/src/containers/Roster/Roster.tsx
```diff
@@ -11,19 +11,49 @@ import { useAlerts } from '../../hooks/useAlerts';
...
+  const editChild = location.state?.editChild;
+  const addChild = location.state?.addChild;
+  const deleteChild = location.state?.deleteChild;
+  useEffect(() => {
+    if (editChild !== undefined && addChild) {
+      apiGet(`children/${editChild}`, {
+        accessToken,
+      })
+        .then((updatedChild) => {
+          children.addOrUpdateRecord(updatedChild);
```
jlhogan, line 48: you should make this call to add / update the record while it is being created in the add record flow -- imagine the user stops mid way and navigates back to the roster without finishing the flow and going through this link. we want the roster to be updated. So i would add `addOrUpdateRecord` call to the place where we refetch the child in the add record flow.

client/src/components/Forms/EnrollmentFunding/Fields/CareModel.tsx
```diff
@@ -22,6 +22,7 @@ export const CareModelField = <T extends Enrollment | ChangeEnrollment>({
   return (
     <FormField<T, RadioButtonGroupProps, CareModel | null>
       getValue={(data) => accessor(data)}
+      preprocessForDisplay={(data) => data}
```
mmazanec22, line 25: Do we need to include the preprocess for display func here?

client/src/components/Forms/EnrollmentFunding/Fields/Funding/NewFunding.tsx
```diff
@@ -61,7 +63,7 @@ export const NewFundingField = <
     );
 
     setFundingSourceOptions(Array.from(_fundingSourceOptions));
-  }, [enrollment, fundingSpaces]);
+  }, [enrollment]);
```
mmazanec22, line 66: Why remove fundingSpaces from the deps array?

client/src/containers/Roster/Roster.tsx
```diff
@@ -11,19 +11,49 @@ import { useAlerts } from '../../hooks/useAlerts';
...
+  const location = useLocation() as LocationType;
   const organizations = user?.organizations;
   const organization = organizations ? organizations[0] : undefined;
   const showOrgInTables = idx(user, (_) => _.organizations.length > 1) || false;
 
   const { children } = useContext(DataCacheContext);
+  const editChild = location.state?.editChild;
+  const addChild = location.state?.addChild;
+  const deleteChild = location.state?.deleteChild;
+  useEffect(() => {
```
mmazanec22, line 42: Why does this live on the roster and not in the forms where this is happening?

client/src/containers/SubmitSuccess/SubmitSuccess.tsx
```diff
@@ -31,9 +31,9 @@ const SubmitSuccess: React.FC = () => {
   return (
     <div className="grid-container margin-top-4">
       <div className="grid-row text-bold">
-        <Link className="usa-button usa-button--unstyled" to="/check-data">
+        <Link className="usa-button usa-button--unstyled" to="/roster">
           <TextWithIcon
-            text="Back"
+            text="Back to Roster"
```
mmazanec22, line 36: Check me on this (with the designs) but I think we're using sentence casing and roster should not be capitalized

client/src/components/Forms/EnrollmentFunding/Fields/CareModel.tsx
bamader: Good catch. We don't; this was a legacy attempt to fix the problem that adding a new record resulted in No Care Model being stored (so I was trying to get it to display Unknown), but this really means that when adding a new record, we should ask providers to select a Care Model type when determining an enrollment.

client/src/components/Forms/EnrollmentFunding/Fields/Funding/NewFunding.tsx
bamader: Because I misidentified the cause of an infinite render and deleted something I shouldn't have... 👀

client/src/components/Forms/EnrollmentFunding/Fields/Funding/NewFunding.tsx
bamader: Ohhhhh yeah. I thought the commented out line with the `&& [code]` was an unwanted part of the filter--didn't realize that should be updated. Fixing now with new passed in `orgId`.

client/src/containers/SubmitSuccess/SubmitSuccess.tsx
bamader: We totally are using sentence casing and I have no idea why I have such a hard-coded instinct to use title case capitalization

client/src/containers/Roster/Roster.tsx
bamader: So my understanding of what a context actually did was suuuuper wrong. This whole trying to push stuff into state to shove into links was an approach I did not at all need to take (yay for learning I guess 🤷‍♂️ ). I talked to Julia this morning and now have a better (and actually correct) way of doing this that uses the cache she built.

client/src/containers/Roster/Roster.tsx
bamader: Good point. I actually noticed funkiness yesterday where if I stopped adding a record in one of the early steps, the number of children on the roster was incremented in the header count, but the child didn't actually appear in the roster list. Probably an artifact of my incorrect cache usage that this should fix.

client/src/components/Forms/EnrollmentFunding/Form/index.tsx
```diff
@@ -44,6 +44,8 @@ export const EnrollmentFundingForm: React.FC<EditFormProps> = ({
       <h2>Enrollment and funding</h2>
       <ChangeEnrollmentForm
         childName={child.firstName || ''}
+        // Child won't actually _not_ have an org because it already existed
+        orgId={child.organization?.id || 0}
```
jlhogan, line 48: we should make this non-nullable in the schema, right? we decided we're not creating children w/out an organization.

client/src/containers/AddChild/AddChild.tsx
```diff
@@ -58,13 +60,15 @@ const AddChild: React.FC = () => {
...
       lastName: '',
       organization,
+      raceNotDisclosed: false,
     };
 
     apiPost('children', placeholderChild, {
       accessToken,
     })
       .then((res) => {
         updateChild(res);
```
jlhogan, line 70: somethng i noticed in a branch i'm working on (had to make this change just for testing other stuff locally) is that we don't need the additional `child` state var (and `updateChild` state var update func)... since `children` is already a state var, we can just have the single child record by getting it from the list `const child = children.records.find((_child) => _child.id === childId) ` 

if folks think it's clearer to have the duplicate state var, im not against it but just a thought

client/src/contexts/DataCacheContext/useReadWriteCache.tsx
```diff
@@ -16,6 +16,12 @@ export const useReadWriteCache = <T extends { id: any }>(apiPath: string) => {
     ]);
   };
 
+  const deleteRecordById = (id: string) => {
```
jlhogan, line 19: nit: should this be `remove` just because the delete has to happen elsewhere? very semantically nit picky not blocking!

jlhogan (approved):

client/src/components/Forms/EnrollmentFunding/Form/index.tsx
bamader: Definitely. It's already in the database migrations as not nullable I'll just make the entity and model match.

client/src/containers/AddChild/AddChild.tsx
bamader: I personally like it because I think it adds some clarity and readability, but if we want to take it out I can make that change.

client/src/contexts/DataCacheContext/useReadWriteCache.tsx
bamader: Yeah it should, helps us be consistent with our naming conventions on methods

client/src/containers/AddChild/AddChild.tsx
mmazanec22: I prefer not to pass variables where we don't need to, since it makes stuff more annoying to write and move, so I'd vote we get the child from the cache.  Plus then it's clearer to see where data is coming from for this form/debug when things are wrong

client/src/containers/Upload/__snapshots__/Upload.test.tsx.snap
```diff
@@ -45,37 +45,11 @@ exports[`Upload matches snapshot 1`] = `
         >
           Choose a file
         </label>
-        <div
+        <input
```
mmazanec22, line 48: Is the USWDS js not loading properly for snapshots or something?

client/src/containers/AddChild/AddChild.tsx
bamader: I'll cut it out then. Since we're still modifying state, we'll have to repeat that find call everywhere we currently use `updateChild`, right? So after creating the child, after refetching due to completing a step, etc.?

client/src/containers/Upload/__snapshots__/Upload.test.tsx.snap
bamader: I have no clue what happened with this snapshot. I didn't touch the upload file at all or modify it's UI, but I had a ton of trouble this morning with the upload snapshot test (it was passing on my local and so not updating, and then failing on CircleCI). This is still an output of `docker-compose exec client yarn run test -u` but I don't know why it made any changes.

jlhogan: it's because we let uswds javascript render the text input thingy. I added a timeout, is it not working in CI? we can just remove this test also -- we're essentially just testing USWDS's javascript, and hopeully theyhave tests so we dont need to

JordanGuinn: I've been experiencing this as well - the past few times I've attempted to update snapshots based on work entirely unrelated to uploads, I've both done and undone this very thing 😕

client/src/components/Forms/EnrollmentFunding/Form/index.tsx
JordanGuinn: We could _also_ theoretically just add `organizationId` to the `Child` db model used by the client, since that column already exists anyway and is inherently supplied when pulling a child row (or rows) out of the table (since we currently query for all columns on row).  We can certainly make `organization` required on the `Child` db model as well, we'd just need to be sure that all APIs returning child rows join on the `organization` table when retrieving said data.

JordanGuinn (commented):
LGTM, once all other PR feedback is addressed 👍🏽👍🏽

client/src/components/Forms/EnrollmentFunding/NewEnrollmentForm/NewEnrollment.tsx
```diff
@@ -73,10 +74,14 @@ export const NewEnrollment = ({
...
           accessor={(data) => data.at('entry')}
         />
+        <CareModelField<Enrollment> accessor={(data) => data.at('model')} />
         <AgeGroupField<Enrollment> accessor={(data) => data.at('ageGroup')} />
         <NewFundingField<Enrollment>
           fundingAccessor={(data) => data.at('fundings').at(0)}
           getEnrollment={(data) => data.value}
+          // Always has an organization because either modifies an
+          // existing child, or creates a new one via org roster
+          orgId={child.organization?.id || 0}
```
kmagida110, line 84: Should there be a different default here rather than 0? Maybe this is a thing where I am new to the language but if there is an organization somewhere shouldn't we use that instead?

kmagida110 (approved):
One new question on my end

client/src/components/Forms/EnrollmentFunding/Form/index.tsx
bamader: That's a good point. I _think_ everything that returns a child satisfies this, but if we'd prefer to go with `organizationId` instead we can just lookup the organization by ID anytime we'd pull it from the child.

client/src/containers/Upload/__snapshots__/Upload.test.tsx.snap
jlhogan: delete it!!!

client/src/components/Forms/EnrollmentFunding/NewEnrollmentForm/NewEnrollment.tsx
bamader: Yes there should! I've addressed this in a change that I just haven't pushed yet, but since we're making organization non-nullable on the client side we won't even need a 'default ' `|| 0` at all.

client/src/components/Forms/EnrollmentFunding/Form/index.tsx
jlhogan: > We can certainly make organization required on the Child db model as well, we'd just need to be sure that all APIs returning child rows join on the organization table when retrieving said data.

would be my vote, ( i think org is already required in the DB model, maybe the shared interface needs to be updated). can enforce the relationship is always hydrated by making it "eager"

client/src/containers/Upload/__snapshots__/Upload.test.tsx.snap
bamader: So remove the Upload.test file (don't wanna delete the wrong thing)?

jlhogan: yes delete the tests, re-run with -u to delete the snapshots or delete them manually

PR #529, opened 2020-10-07
Add batch edit item content for childIdent

### LINKED ISSUE #475: Child identifiers batch edit form

### LINKED ISSUE #476: child demographics batch edit form

jlhogan:
closes #475 closes #476 

I realized that we really should be using the same forms for all this stuff, so i'm throwing this up and then going to work on getting edit enrollment, new enrollment, and change enrollment to all leverage the same base form. That way, if we update form stuff, we don't have to do it in multiple places (i.e. noticed care model was missing from one of the many enrollment forms). 

Idea is we can use the new additional `batchEditProps` to correctly display the form in batch edit vs normal add/edit mode

some things are renamed for clarity to my brain while working thru this

tests not yet reconciled, but wanted to get this in front of brains/eyes

### DIFF
--- a/client/src/components/Forms/ChildIdentifiers/Form.tsx
+++ b/client/src/components/Forms/ChildIdentifiers/Form.tsx
@@ -8,8 +8,8 @@ import {
   DateOfBirthField,
   BirthCertificateFormFieldSet,
 } from './Fields';
-import { Form, FormSubmitButton } from '@ctoec/component-library';
-import { EditFormProps } from '../types';
+import { Form, FormSubmitButton, Button } from '@ctoec/component-library';
+import { RecordFormProps } from '../types';
 import AuthenticationContext from '../../../contexts/AuthenticationContext/AuthenticationContext';
 import { Child } from '../../../shared/models';
 import { apiPut } from '../../../utils/api';
@@ -33,11 +33,15 @@ export const doesChildIdFormHaveErrors = (child?: Child) =>
   child ? !!getValidationStatusForFields(child, childIdentifiersFields) : true;
 
 export const ChildIdentifiersForm = ({
-  child: inputChild,
+  record: inputChild,
   afterDataSave,
   hideHeader = false,
   hideErrorsOnFirstLoad,
-}: EditFormProps) => {
+  batchEditProps = {
+    showField: () => true,
+    SkipButton: <></>,
+  },
+}: RecordFormProps) => {
   const { accessToken } = useContext(AuthenticationContext);
   const isMounted = useIsMounted();
   const [saving, setSaving] = useState(false);
@@ -62,6 +66,7 @@ export const ChildIdentifiersForm = ({
       .finally(() => (isMounted() ? setSaving(false) : null));
   };
 
+  const { showField, SkipButton } = batchEditProps;
   return (
     <Form<Child>
       className="ChildIdentifiersForm usa-form"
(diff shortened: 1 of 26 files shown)


client/src/components/Forms/ChildIdentifiers/Form.tsx
```diff
@@ -33,11 +33,15 @@ export const doesChildIdFormHaveErrors = (child?: Child) =>
   child ? !!getValidationStatusForFields(child, childIdentifiersFields) : true;
 
 export const ChildIdentifiersForm = ({
-  child: inputChild,
+  record: inputChild,
```
mmazanec22, line 36: Why not child?  IMO a more descriptive variable name, and one that matches the model, is more helpful

client/src/components/Forms/ChildIdentifiers/Form.tsx
```diff
@@ -71,29 +76,44 @@ export const ChildIdentifiersForm = ({
...
+            <SuffixField />
+          </div>
+        )}
       </div>
-      <DateOfBirthField />
-      <BirthCertificateFormFieldSet />
+      {showField('birthdate', childIdentifiersFields, child) && (
+        <DateOfBirthField />
+      )}
+      {showField('birthCertificateId', childIdentifiersFields, child) && (
```
mmazanec22, line 109: I wonder if we can instead pass a hideField func to the form field/fieldset, that is then processed via the component library and adds a hidden class?

mmazanec22: We could maybe even pass it to the form and have the form pass it down?

client/src/containers/BatchEdit/listSteps.tsx
```diff
@@ -0,0 +1,59 @@
...
+) =>
+  formSections
+    .map(({ key, name, hasErrors }) => {
+      const Form = newForms.find((s) => s.key === key)?.form || (() => <></>);
+      const hasMissingInfo = hasErrors(record);
+      if (hasMissingInfo) {
+        return {
+          key,
+          name,
+          // TODO: WHY THE F is this "props" object actually the child record???????
```
mmazanec22, line 38: Bc the status funcs need to check it for errors?  Or is this just a complaint about the terrible choice of variable name?

mmazanec22 (approved):

client/src/components/Forms/ChildIdentifiers/Form.tsx
jlhogan: whats the benefit to having the html there but hidden? 
also, not everything is in a fieldset

client/src/containers/BatchEdit/listSteps.tsx
jlhogan: no i just was expecting the `props` argument to that functoin to be the actual form props (like it is in the function signature, and in the StepList code, and in the AddChild/listSteps). For some reason here, props is not the props object, with a child property, props is just the child object directly

jlhogan: unless i'm wildly misunderstanding something

mmazanec22: Ah yeah sorry for the misleading naming choice

client/src/components/Forms/ChildIdentifiers/Form.tsx
mmazanec22: No benefit to that, just trying to think of how we can avoid having to wrap every single field in a function.  Seems like something a form component could deal with

client/src/containers/BatchEdit/listSteps.tsx
jlhogan: we need to circle back on this b/c it is just working differently in the other steps _I THINK_ ?

client/src/components/Forms/ChildIdentifiers/Form.tsx
```diff
@@ -71,29 +76,44 @@ export const ChildIdentifiersForm = ({
...
-        <SasidField />
-      </div>
-      <div className="mobile-lg:grid-col-9">
-        <FirstNameField />
-      </div>
-      <div className="mobile-lg:grid-col-9">
-        <MiddleNameField />
-      </div>
-      <div className="display-flex flex-row flex-align-end grid-row grid-gap">
+      {showField('sasid', childIdentifiersFields, child) && (
```
bamader, line 79: I really like this as a way to use the same base form set for everything

client/src/components/Forms/ChildIdentifiers/Form.tsx
bamader: One (maybe naive) way to do things is to build a parameter into `field` called `isBatchEdit` or something to that effect, and it's simply a boolean value. If that parameter is true and we're doing a batch edit, then we have the field check if there are validation errors in the respective part of the data; if there's no errors, it doesn't need to be shown in batch edit. If `isBatchEdit` is false, we must be coming from add a new record or single editing a record via `EditRecord`, in which case we just always show the field. Does that achieve the goal?

bamader (approved):

client/src/containers/BatchEdit/listSteps.tsx
jlhogan: ooooo i see
sorry, my brain :frowning_face:

PR #544, opened 2020-10-09
541 add missing info column

### LINKED ISSUE #541: Move missing info icon into it's own sortable column
SortedbyMissingInfo

kmagida110:
Fixes #541

### DIFF
--- a/client/src/containers/Roster/TableColumns.tsx
+++ b/client/src/containers/Roster/TableColumns.tsx
@@ -40,7 +40,7 @@ export const tableColumns: (_?: boolean) => Column<Child>[] = (
         !row.validationErrors || !row.validationErrors.length ? 1 : 0,
       cell: ({ row }) => {
         return (
-          <td scope="row" className={tableRowClassName}>
+          <td className={tableRowClassName}>
             {(!!row.validationErrors && row.validationErrors.length == 0) ||
               row.validationErrors == undefined ||
               InlineIcon({ icon: 'incomplete' })}
--- a/client/src/containers/Roster/TableColumns.tsx
+++ b/client/src/containers/Roster/TableColumns.tsx
@@ -35,6 +35,18 @@ export const tableColumns: (_?: boolean) => Column<Child>[] = (
       className: tableColumnClassName,
       name: 'Missing info',
       width: `${shortColumnWidthPercent}%`,
+      sort: (row) => {
+        // Default sort with missing records on top on the assumption that missing records will be the priority
+        if (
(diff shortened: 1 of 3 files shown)


bamader (approved):
LGTM!

client/src/containers/Roster/TableColumns.tsx
```diff
@@ -18,20 +18,41 @@ export const tableColumns: (_?: boolean) => Column<Child>[] = (
...
-              to={{
-                pathname: `/edit-record/${row.id}`,
-              }}
-            >
-              {row.lastName}, {row.firstName}{' '}
-              {(!!row.validationErrors && row.validationErrors.length == 0) ||
-                row.validationErrors == undefined ||
-                InlineIcon({ icon: 'incomplete' })}
-            </Link>
+            {(!!row.validationErrors && row.validationErrors.length == 0) ||
```
mmazanec22, line 53: !row.validationErrors || !row.validationErrors.length should cover all of this I think?

mmazanec22 (approved):

client/src/containers/Roster/TableColumns.tsx
jlhogan: agreed, i have this change in a diff branch but simpler conditionals is always a win IMO!!

client/src/containers/Roster/TableColumns.tsx
```diff
@@ -18,20 +18,41 @@ export const tableColumns: (_?: boolean) => Column<Child>[] = (
...
+        ) {
+          return 1;
+        } else {
+          return 0;
+        }
+      },
+
       cell: ({ row }) => {
         return (
           <th scope="row" className={tableRowClassName}>
```
jlhogan, line 52: this column should not be a th, should be td like the other rows

jlhogan (changes requested):

client/src/containers/Roster/TableColumns.tsx
kmagida110: 👍

client/src/containers/Roster/TableColumns.tsx
kmagida110: Sounds good, copied this from the name, should name be td too?

jlhogan: nope :) name is the header for the row (thus, `scope="row"`)

kmagida110: Fix is pushed, thanks @jlhogan

jlhogan (approved):

PR #547, opened 2020-10-09
Birth certificate

### LINKED ISSUE #325: Birth Certificates: I can enter data for US, non-US and unavailable birth certificates.
# [Updated designs](https://share.goabstract.com/ba1d8607-c43c-41e9-accf-37dfd45afafd)
Screen Shot 2020-10-02 at 1 06 55 PM
Screen Shot 2020-09-28 at 3 39 31 PM
Screen Shot 2020-09-28 at 3 39 25 PM

Will require updates to the data spreadsheet template
– Birth Certificate type data element with acceptable values of US, non-US and unavailable
- Add unknown/not collected option to US birth certificate town and state
-- data is only required if US birth certificate

### LINKED ISSUE #542: Care model doesn't show in enrollment card, so if it is missing, the user can't tell which record has errors
Screen Shot 2020-10-08 at 3 37 36 PM
Screen Shot 2020-10-08 at 3 38 49 PM

bamader:
Adds the ability to select non-US birth certificate types to the child info page, including the ability to choose a birth certificate not available option. Migration added to DB to include a new field for type of birth certificate, along with corresponding enum. Fixes enrollment card to show type of care model so that if there's an error, a user will see it at a glance. Updates `yarn.lock` to grab modal fixes from a new commit to the component library.

Note: The massive diff numbers are super distorted by one snapshot that has a ton of changes because modals got updated. This is NOT actually a thousands-of-changes PR 😝 

Closes #542 
Closes #325


client/src/components/Forms/ChildIdentifiers/Fields/BirthCertificateFieldSet.tsx
```diff
@@ -1,41 +1,74 @@
...
+          data.value,
+          [
+            'birthCertificateType',
+            'birthCertificateId',
+            'birthTown',
+            'birthState',
+          ],
+          {
+            message:
+              'Birth certificate type selection is required for OEC reporting. ' +
```
mmazanec22, line 41: Why string concatenation here?

mmazanec22: We should maybe set the message on the backend validations instead and alter the getValidationStatusForFields func to take "use backend message" or something as an option

client/src/components/Forms/ChildIdentifiers/Form.tsx
```diff
@@ -89,7 +90,7 @@ export const ChildIdentifiersForm = ({
         </div>
       </div>
       <DateOfBirthField />
-      <BirthCertificateFormFieldSet />
+      <BirthCertificateField />
```
mmazanec22, line 93: This is nitpicky af and by no means a blocker but I personally like the clarity of calling things fieldsets when that's what we're using under the hood, even if we've abstracted that away-- I think it makes it easier to look at the form component and understand how it relates to what you see on the page.

client/src/components/Forms/ChildIdentifiers/__snapshots__/ChildIdentifiers.test.tsx.snap
```diff
@@ -1417,88 +1417,76 @@ exports[`EditRecord ChildIdentifiers matches snapshot 1`] = `
...
+              id="US birth certificate"
+              name="birth-certificate-fields"
+              type="radio"
+              value="US birth certificate"
+            />
             <label
-              class="usa-label"
-              for="birthCertificateId"
+              class="usa-radio__label"
+              for="US birth certificate"
```
mmazanec22, line 1446: ID values cannot contain spaces-- this needs to be changed so that the label and input are programmatically associated

client/src/components/Forms/ChildIdentifiers/__snapshots__/ChildIdentifiers.test.tsx.snap
```diff
@@ -1417,88 +1417,76 @@ exports[`EditRecord ChildIdentifiers matches snapshot 1`] = `
...
+            <input
+              class="usa-radio__input"
+              id="Non-US birth certificate"
+              name="birth-certificate-fields"
+              type="radio"
+              value="Non-US birth certificate"
+            />
+            <label
+              class="usa-radio__label"
+              for="Non-US birth certificate"
```
mmazanec22, line 1465: Same here

client/src/components/Forms/ChildIdentifiers/__snapshots__/ChildIdentifiers.test.tsx.snap
mmazanec22: (Realizing this might be due to behavior of component library-- we should change that)

mmazanec22 (approved):

client/src/components/Forms/ChildIdentifiers/Fields/BirthCertificateFieldSet.tsx
bamader: I literally just used concatenation because I was like "this line is running on long, if I use concatenation I can split it in two and not have to scroll over"... terrible reason, I know 😂 

As far as backend validations go, that sounds good, but I'm not super clear on where to make that change. I see a `raceValidation` and a `genderValidation` in the `src/entity/decorators/Child` folder; would we make one for birth certificate there that has the same form?

client/src/components/Forms/ChildIdentifiers/Form.tsx
bamader: Totally fair! Back to fieldset it shall go!

client/src/components/Forms/ChildIdentifiers/__snapshots__/ChildIdentifiers.test.tsx.snap
bamader: I'm pretty sure this is the component library. I made the id of the actual `BirthCertificateField` (soon to be `BirthCertificateFieldSet`) hyphenated, so I think the component library or tests got the spaces in somehow.

client/src/components/Forms/ChildIdentifiers/Fields/BirthCertificateFieldSet.tsx
bamader: We can also make the backend validation change in a separate PR so that way we can group it with other validations code? Actually wait this may be a good time to make a new ticket for something.

mmazanec22: I think validation changes belong on this PR-- bc without a change they will block someone from choosing a non-US birth cert

bamader: Yeah you were totally right, selecting anything besides US birth cert got flagged as an error. I added a backend validator for birth certificate that removes that blocker.

src/entity/decorators/Child/birthCertificateValidation.ts
```diff
@@ -0,0 +1,33 @@
+import { registerDecorator, ValidationOptions } from 'class-validator';
+import { Child } from '../../Child';
+
+const childUSCertificateMessage =
+  'Birth certificate ID, birth town, and birth state are required for US birth certificates.';
+const birthFieldsForUSCert = ['birthCertificateId', 'birthTown', 'birthState'];
+
+function isBirthCertificateValid(child: Child) {
+  return (
+    child.birthCertificateType !== 'US birth certificate' ||
```
jlhogan, line 10: should check against the enum value, not magic string

jlhogan (approved):

client/src/components/Forms/ChildIdentifiers/Fields/BirthCertificateFieldSet.tsx
```diff
@@ -1,41 +1,69 @@
...
-        </div>
-      </div>
-    </FormFieldSet>
+      options={[
+        ...Object.values(BirthCertificateType).map((certificateType) => ({
+          render: (props: RadioOptionRenderProps) => (
+            <RadioButton {...props} text={certificateType} />
+          ),
+          value: certificateType,
+          expansion: certificateType === 'US birth certificate' && (
```
JordanGuinn, line 50: nit: Should reference the enum value here

client/src/shared/models/db/Child.ts
```diff
@@ -12,6 +12,7 @@ export interface Child extends ObjectWithValidationErrors {
   lastName?: string;
   suffix?: string;
   birthdate?: Moment;
+  birthCertificateType: string;
```
JordanGuinn, line 15: Can the value of `birthCertificateType` actually be the `BirthCertificateType` enum instead of `string`?  Especially if the db column in non-nullable, which it appears to be.

src/entity/Child.ts
```diff
@@ -63,17 +64,25 @@ export class Child implements ChildInterface {
...
   birthdate?: Moment;
 
+  @Column({
+    type: 'varchar',
+    length: 30,
+    nullable: false,
+    transformer: enumTransformer(BirthCertificateType),
+  })
+  @ChildBirthCertificateSpecified()
+  birthCertificateType: string;
```
JordanGuinn, line 74: Same for this; can it just be:

`birthCertificateType: BirthCertificateType;`

?

src/template/EnrollmentReportRow.ts
```diff
@@ -103,6 +104,18 @@ export class EnrollmentReportRow {
...
+    formattedName: 'Birth certificate type',
+    required: REQUIRED,
+    definition: 'Type of birth certificate based on country of issue',
+    reason:
+      'Used for a variety of reporting; allows linking to a variety of data sets, including SASID-backed data.',
+    format: Object.values(BirthCertificateType).join(', '),
+    example: 'US birth certificate',
+    section: SECTIONS.CHILD_INFO,
+  })
+  birthCertificateType: string = undefined;
```
JordanGuinn, line 117: Aaaaaaaaaand here as well (`birthCertificateType: BirthCertificateType`).  Also if this is a required field, do we need to specify the fallback `undefined` default value?

src/entity/Child.ts
```diff
@@ -63,17 +64,25 @@ export class Child implements ChildInterface {
...
+
   @Column({ nullable: true })
-  @IsNotEmpty({ message: 'Birth town is required' })
+  @ChildBirthCertificateSpecified()
```
JordanGuinn, line 77: +10000 for the use of these decorators 🔥

src/entity/Child.ts
```diff
@@ -63,17 +64,25 @@ export class Child implements ChildInterface {
   })
   birthdate?: Moment;
 
+  @Column({
+    type: 'varchar',
+    length: 30,
+    nullable: false,
+    transformer: enumTransformer(BirthCertificateType),
+  })
+  @ChildBirthCertificateSpecified()
```
mmazanec22, line 73: This doesn't need to be a custom validation-- just an IsNotEmpty

src/entity/Child.ts
mmazanec22: In this case we just needed validate if rather than a custom decorator: https://github.com/typestack/class-validator#conditional-validation

bamader: Great callout on the conditional validation! I ended up using the custom director over just `IsNotEmpty` only because `IsNotEmpty` registers as having an error if the value is either null or undefined, but those are both interpretable states for our fields. `null` means the "unknown/not collected" box is checked, for example.

PR #554, opened 2020-10-09
E2e tests

mmazanec22:
Sorry for annoying whitespace diff


bamader (approved):
Yay testing!

nightwatch.conf.js
```diff
@@ -0,0 +1,26 @@
...
+  selenium: {
+    start_process: false,
+    host: 'hub-cloud.browserstack.com',
+    port: 80,
+  },
+  test_settings: {
+    default: {
+      desiredCapabilities: {
+        'browserstack.user': 'julia518',
+        'browserstack.key': 'Uq4nUNp12oP9oMoYquej',
```
jlhogan, line 13: this iisn't actually like, top secret but probably best to not check it in to source and get it from env var in circleCI

tests_output/CHROME_85.0.4183.83_WINDOWS_smokeTest.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
jlhogan, line 1: remove this from diff

tests_output/smokeTest.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
JordanGuinn, line 1: Surprised and confused at how these files are even present with `/tests_output` having been added to the `.gitignore` 😅

nightwatch.conf.js
```diff
@@ -0,0 +1,26 @@
...
+        browserName: 'edge',
+      },
+    },
+  },
+};
+
+// Code to copy seleniumhost/port into test settings
+for (var i in nightwatch_config.test_settings) {
+  var config = nightwatch_config.test_settings[i];
+  config['selenium_host'] = nightwatch_config.selenium.host;
```
JordanGuinn, line 23: nit: Does the setting of `selenium_host` and `selenium_port` need to exist within this for loop?  Can't they just live outside and be executed once?

e2e-tests/smokeTest.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  '@tags': ['smoke-test'],
+  'smoke-test': function (browser) {
+    browser
+      .url('https://staging.ece-fawkes.ctoecskylight.com')
```
JordanGuinn, line 5: Less of a blocking concern for this PR and more of a conversation starter - moving forward, do we want to allow for our e2e tests to be run across environments?  Like do we want the successful completion of these tests to be a gate to merge, in the same way that our unit tests are?

nightwatch.conf.js
mmazanec22: I think this will become necessary once we have it set up to test in multiple browsers?

e2e-tests/smokeTest.js
jlhogan: i think that would be probably _ideal_ but IMO just getting them to run as a post-merge pre-prod-release step is a good initial pass at this

nightwatch.conf.js
jlhogan: i would either: also leverage this loop for dupes browserstack creds OR remove the loop and duplicate values everywhere

jlhogan (approved):

PR #569, opened 2020-10-14
Login e2e

mmazanec22:


package.json
```diff
@@ -47,7 +47,7 @@
     "serve": "node dist/src/index.js",
     "serve-debug": "nodemon --inspect dist/src/index.js",
     "start": "npm run serve",
-    "test": "node_modules/nightwatch/bin/nightwatch e2e-tests/smokeTest.js",
+    "test": "node_modules/nightwatch/bin/nightwatch -e default,edge",
```
jlhogan, line 50: this means we're running these tests in CI, which i dont think we want to do right now (b/c i dont think it makes sense to do them for staging on every commit to a feature branch -- they wont test the feature branch!) can we rename this command `test-e2e` or something (b/c i hope to soonly add a few backend unit tests to actually run as part of CI) and revert `test` to be something like `exit 0` for now?

jlhogan (approved):
:tada: yay yay yay yay

package.json
mmazanec22: Sure-- should we give e2e tests their own little directory then?  Having them here made sense to me when we were setting up but if they're just running against staging then it's kinda silly to have the backend have anything to do with them

jlhogan: that makes sense to me! they may eventually run against pre-base branch as a pre-merge-into-base check but we aint there yet

jlhogan: that makes sense to me! they may eventually run against pre-base branch as a pre-merge-into-base check but we aint there yet

e2e-tests/login.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+  '@tags': ['login'],
+  login: async function (browser) {
+    // Initializes with the launch_url value set in config
+    await browser.init();
+
+    // Click login from ECE Reporter
+    await browser.click('xpath', "//*/a[contains(@href,'/login')]");
+    await browser.waitForElementVisible('css selector', 'body');
```
bamader, line 9: The fact that we can write a script that programmatically operates a webpage is blowing my mind 🤯

package.json
bamader: I think this is because I don't fully understand the testing environment / setup, but could you explain the nuance going on here? More specifically, why does renaming the command change how the test executes?

bamader (approved):
LGTM!

package.json
mmazanec22: The "test" command I changed here is in the backend package.json file, so it runs in CI as the backend tests (you can look in the build process-- right now in base the "backend tests"step just exits and doesn't do anything).  Really though, the backend is going to have its own tests, and that's what actually belong under the "test" command in the backend's package.json.

I had figured it would be ok to just shove these in here since we don't have any actual backend tests yet, but Julia's right that it doesn't really make sense that a PR would then have to run the e2e tests-- against staging, and not even the changes the PR actually makes-- before merging.

bamader: Ohhh I understand now. Thanks for the explanation, Melanie, that helped a ton!

PR #585, opened 2020-10-14
581: Cache does not attempt to update, or mark itself as updated, unl…

### LINKED ISSUE #581: Roster does not display records on log-in until the user refreshes the page.
Added one record. 
Screen Shot 2020-10-14 at 2 40 49 PM

Logged out and back in again and saw this. Records showed when I refreshed the page. 
Screen Shot 2020-10-14 at 2 41 00 PM

jlhogan:
…ess someone is signed in!

closes #581


client/src/contexts/DataCacheContext/useReadOnlyCache.tsx
```diff
@@ -14,7 +14,7 @@ export const useReadOnlyCache = <T extends { id: any }>(apiPath: string) => {
     records.find((r) => r.id === id);
 
   useEffect(() => {
-    if (!haveFetched) {
+    if (!haveFetched && accessToken) {
```
bamader, line 17: Potentially dumb question, but how could a user get an access token unless they were signed in? I think I'm missing what the fix actually is because it's so succinct.

bamader (approved):
LGTM!

client/src/contexts/DataCacheContext/useReadOnlyCache.tsx
jlhogan: the fix is only to load the cache data when an authenticated user is navigating the site -- otherwise, the code inside this `if` will fire, and `haveFetched` will be set to true, so then when access token changes on log in, the cache data will not be fetched. An alternative fix here would be to remove the early exit is api wrapper for when access token is null, and only update `haveFetched` in the `then` block after the apiGet, not  `finally`

mmazanec22 (approved):

.circleci/config.yml
```diff
@@ -50,10 +50,10 @@ jobs:
           at: .
       - run:
           name: Run npm audit for back-end dependencies
-          command: npm i --package-lock-only && npm audit
+          command: yarn audit
```
JordanGuinn, line 53: yarn works?! 🎉

client/src/contexts/DataCacheContext/useReadOnlyCache.tsx
JordanGuinn: So this doesn't actually wipe the cache on logout, but rather refreshes the cache on login?

JordanGuinn (approved):

client/src/contexts/DataCacheContext/useReadOnlyCache.tsx
jlhogan: yeah -- this does not close the "clear out cache on logout" feature, it just fixes this specific bug. That work still exists and should be done :)

.circleci/config.yml
jlhogan: yeah i lied heheh whoops

PR #588, opened 2020-10-15
Batch edit address

### LINKED ISSUE #154: Copy Edits to Data Definitions
# Edits

Family income determination – This data is collected to assess eligibility for public funding by calculating state median income percentage. Income determinations must be updated at least once a year.

**From "SASID" to "Unique Identifier"**
- Optional
- Definition: A unique number used to identify children. Can be either an SDE-backed SASID or another unique identifier from a system of record such as Childplus. 
- Reason for collecting: Allows for easy reference with your program's system of record. 

**Birth Certificate ID #**
– Definition: The identification number on the child's birth certificate. 
Ex: 123-20-000000

– Accepted format: Text; Generally an 11-digit number written in XXX-XX-XXXXXX format. Format varies by state. 

**Dual Language Learner**
– Definition: Children who have a home language other than English and are learning to or more languages at the same time, or learning a second language while continuing to develop their first language.

**Experiencing homelessness or housing insecurity** Changed from (Experienced)

**Enrollment exit reason**
– Definition: The reason for ending an enrollment. Options include: Aged out, stopped attending, chose to attend another program, moved within Connecticut, moved to another state, Withdrew due to lack of payment, Child was asked to leave, unknown. 

**Care model** Changed from (Model)

**Space type**
– Definition: The contract space type that funds an enrollment. See [contract spaces](https://staging.ece-fawkes.ctoecskylight.com/funding-space-types) for the full list of accepted space types. 

## CDC Spaces

Full-time – "1" or "Full-time" or "FT"
Part-time – "2" or "Part-time" or "PT"
Part-time/Full-time – "3" or "Part-time/Full-time" or "PT/FT"

## State Head Start Space types
Additional Full-day/full-year spaces – "9" or "+FD/FY"
Additional school-day/school-year spaces – "10" or "+SD/SY"
Additional school-day/school-year/extended-year - "11" or "+SD/SY/EY"
Extended day/year spaces – "12" or "ED/EY"
Extended year spaces - "13" or "EY"
Extended day spaces - "14" or "ED"

### LINKED ISSUE #475: Child identifiers batch edit form

### LINKED ISSUE #477: family address batch edit form

### LINKED ISSUE #584: Family address changes are inconsistently displaying.
Address bug
- On adding a family address, the change is not shown.
- The change is shown when I navigate to another tab and back. 
- On refresh the changes are cleared.

bamader:
Implements batch editing for both child identifiers and family addresses. Also updates copy on a number of data definitions and expands the funding time options for various sources.
 
Closes #475 
Closes #477 
Closes #154


client/src/containers/BatchEdit/ChildIdentifiers/Form.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
jlhogan, line 1: you shouldn't need this special wrapper component, this filtering should happen at the step https://github.com/ctoec/data-collection/blob/9022da4993543589989b2fdd7a02b01ac5a765f3/client/src/containers/BatchEdit/listSteps.tsx#L16 so unless the record has errors for this form we won't even add the step to the step list!!

client/src/containers/BatchEdit/ChildIdentifiers/__snapshots__/Form.test.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
jlhogan, line 1: something seems wrong about this :)

client/src/containers/BatchEdit/listSteps.tsx
```diff
@@ -6,13 +6,21 @@ import {
...
+const forms = [
+  { key: SECTION_KEYS.ENROLLMENT, form: EnrollmentFundingForm },
+  { key: SECTION_KEYS.FAMILY, form: FamilyAddressForm },
+  { key: SECTION_KEYS.IDENT, form: ChildIdentifiersForm },
+];
 
 export const listSteps = (child: Child) =>
   formSections
-    .filter((s) => s.key === SECTION_KEYS.ENROLLMENT)
+    .filter(
```
jlhogan, line 21: this flter was just to keep un-implemented kruft out of showing up in the batch edit flow, but i think it's fine to just delete it at this point and know that some stuff is gonna look very weird until it's all done

src/controllers/children.ts
```diff
@@ -28,6 +28,7 @@ export const getChildren = async (user: User) => {
   const readOrgIds = await getReadAccessibileOrgIds(user);
   return getManager().find(Child, {
     relations: [
+      'family',
```
jlhogan, line 31: i _think_ we should be ok only having this in the `getChild` functionality. As things exist right now, the FE needs to do the individual child fetch by ID to get the version of the record that has all necessary information (family, income determinations, distributed validation errors) when moving into a detailed view (edit record, record is active item in batch edit)`. What we need to do it make sure that this by-id fetch happens when those transitions occur before anything renders for the user, and that after it happens the record in the cache is updated appropriately. Happy to talk thru this with faces and voices if it doesn't make sense

client/src/containers/BatchEdit/ChildIdentifiers/Form.tsx
bamader: Oh good point. This thing literally just `&&`'s completely normal form with a boolean, so I should just be able to import the FamilyAddressForm from `/components/Forms` and use that directly in listSteps instead.

bamader: Oh good point. This thing literally just &&'s completely normal form with a boolean, so I should just be able to import the FamilyAddressForm from /components/Forms and use that directly in listSteps instead.

client/src/containers/BatchEdit/ChildIdentifiers/__snapshots__/Form.test.tsx.snap
bamader: ...I didn't check the snapshot content before pushing but ya I'd say you're right 😝

src/controllers/children.ts
bamader: Mmmm I think I understand. I'm looking at the BatchEdit code proper and I see on line 105 that when we supply a `record` prop for the `BatchEditItemContent`, we actually use the cache to look up the child's more detailed view and then hand that off. You're saying we need to do that same thing when the Roster page hands off a specific child to EditRecord, so that EditRecord works with this detailed view, then updates the appropriate entry in the cache?

client/src/containers/BatchEdit/ChildIdentifiers/__snapshots__/Form.test.tsx.snap
bamader: I guess do we still need a `.test.tsx` and a snapshot file for each of FamilyAddress and ChildIdentifiers? Since I'm just directly re-using those forms, they won't be listed under their own directories anymore.

src/controllers/children.ts
bamader: Actually it looks like EditRecord does look up the child's more detailed view by ID and then updates the record in the cache upon every refetch. So the functionality that I just described should already exist. Are you saying it's a problem with the rendering order then?

jlhogan: yes after I said that I started looking in to it. Have a fix for the fetching on a branch now, but trying to get the edit forms to re-render after re-fetch. EYE ROLL. Anyway, TLDR; i ended up working on this b/c it was confusing to me so we can pick this back up when i open that PR shortly ?

client/src/containers/BatchEdit/ChildIdentifiers/__snapshots__/Form.test.tsx.snap
jlhogan: no then additional snapshot testing is not necesary IMO

src/controllers/children.ts
bamader: Sounds good. I'll just leave that part out of this PR (which is about to end up real short 😛 )

bamader: Actually, I think I'm gonna put this branch on pause for the time being. If you're delving into this in another PR, it's probably best we get the cache refetch --> re-render working before I muck around here, since that'll play into testing the new kinds of BatchEdit.

mmazanec22 (approved):

src/migration/1602852240115-RenameSasid.ts
```diff
@@ -0,0 +1,25 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class RenameSasid1602852240115 implements MigrationInterface {
+  name = 'RenameSasid1602852240115';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "child" DROP COLUMN "sasid"`);
```
jlhogan, line 7: do we want to keep sasid and unique IDseparately ? will make using sasids easier as we won't have to go check if each ID is a sasid or not. @kmagida110, any opinions there?

src/template/EnrollmentReportRow.ts
```diff
@@ -466,7 +467,8 @@ export class EnrollmentReportRow {
   @ColumnMetadata({
     formattedName: 'Space type',
     required: REQUIRED,
-    definition: 'The period during the day when services are provided',
+    definition:
+      'The contract space type that funds an enrollment. See contract spaces for the full list of accepted space types.',
```
jlhogan, line 471: should this be a link? definitions are rendered as markdown in the UI, so can add link here by writing it as a markdown link https://www.markdownguide.org/basic-syntax/#links

jlhogan (commented):
Batch-edit stuff approved, but questions remain on the other things (an argument for keeping PRs separate :) )

src/migration/1602852240115-RenameSasid.ts
bamader: We could always add a boolean field called `isSasid` and then just display it as a checkbox next to the text input field in the UI? That way we don't have to show both a Sasid text field and a Unique Identifier text field. Can switch to other formats if it's easier, though.

src/template/EnrollmentReportRow.ts
bamader: This almost certainly should be a link. I didn't see that in the design image on the issue page. Good catch.

src/migration/1602852240115-RenameSasid.ts
jlhogan: we intend to know per-site whether they use SASIDs or some other UUID, so we can optionally only display the necessary field in the form and template, since all of that is programmatically generate

bamader: That makes a lot of sense. So we'll keep the `sasid` field and then add a `uniqueId`field, make them both nullable in the DB and optional on the model schema, and then display the correct version of the field based on which site the enrollment is coming from. This means we'll have to add a field to the `site` model about whether they use SASIDs or non-SASID unique IDs, right?

bamader: @jlhogan Since we've now got a separate ticket to handle the site-level unique identifiers / SASID-usage, is it okay if I just dump the refactoring I had in this PR? (It's just undoing a rename symbol refactor so we're not losing a lot of work, but this way I can keep this PR more cleanly scoped)

kmagida110: I would definitely say that we should keep, at least in the DB, a SASID field. We can edit it on the front end as needed but having a SASID column and a uniqueID field in the DB will allow us to store the information correctly and self-document for the future. We can also have additional validations if needed on the SASIDs

bamader: All good points. I'm just gonna use this PR for what it's supposed to do and not try to make any changes to the SASID or unique ID fields.

jlhogan: sounds good!

jlhogan (approved):

PR #595, opened 2020-10-16
Simplifies child-detail-requiring flows by using a duplicate child st…

jlhogan:
…ate var

also closes an un-opened bug where add record flow doesn't work right now :) (but maybe work in staging depending on how you try it due to the migration having been run there allowing for nulls) !!

todo: cleanup comments i just made inline + clean up tests -- but i can't right now at 5:30 on a friday :) 

relies on https://github.com/ctoec/component-library/pull/82

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@ctoec/component-library": "0.0.35",
+    "@ctoec/component-library": "https://github.com/ctoec/component-library#update-sidenav-render-children",
     "@openid/appauth": "^1.2.7",
     "@testing-library/dom": "^7.24.1",
     "@testing-library/jest-dom": "^4.2.4",
--- a/client/package.json
+++ b/client/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@ctoec/component-library": "https://github.com/ctoec/component-library#update-sidenav-render-children",
+    "@ctoec/component-library": "0.0.36",
     "@openid/appauth": "^1.2.7",
     "@testing-library/dom": "^7.24.1",
     "@testing-library/jest-dom": "^4.2.4",
--- a/client/src/components/Forms/Enrollment/Form.tsx
+++ b/client/src/components/Forms/Enrollment/Form.tsx
@@ -33,10 +33,9 @@ export const doesEnrollmentFormHaveErrors = (
       : false;
   }
 
-  const result = child?.enrollments?.length
+  return child?.enrollments?.length
     ? !!getValidationStatusForFields(child.enrollments, enrollmentFields)
     : true;
-  return result;
 };
 
 type EnrollmentFormProps = {
--- a/client/src/components/Forms/Enrollment/Form.tsx
+++ b/client/src/components/Forms/Enrollment/Form.tsx
@@ -9,7 +9,7 @@ import {
   NewFundingField,
 } from './Fields';
 import UserContext from '../../../contexts/UserContext/UserContext';
-import { EditFormProps } from '../types';
+import { RecordFormProps } from '../types';
 import { getValidationStatusForFields } from '../../../utils/getValidationStatus';
 import AuthenticationContext from '../../../contexts/AuthenticationContext/AuthenticationContext';
 import { getCurrentEnrollment } from '../../../utils/models';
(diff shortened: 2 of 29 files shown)


client/src/components/Forms/Enrollment/Form.tsx
```diff
@@ -33,22 +33,22 @@ export const doesEnrollmentFormHaveErrors = (
       : false;
   }
 
-  return child?.enrollments?.length
+  const result = child?.enrollments?.length
```
jlhogan, line 36: todo: undo -- left over from testing

client/src/containers/BatchEdit/BatchEditItemContent.tsx
```diff
@@ -41,51 +51,56 @@ export const BatchEditItemContent: React.FC<BatchEditItemContentProps> = ({
...
+    apiGet(`/children/${childId}`, { accessToken })
       .then((updatedChild) => {
-        children.addOrUpdateRecord(updatedChild);
-        if (activeStepKey.length) {
+        setChild(updatedChild);
+        updateRecordInCache(updatedChild);
+        // If child has loaded, steps are created, and first step is active
+        // then attempt to advance after refetch
+        if (activeStepKey && steps) {
+          console.log('updated child', updatedChild);
```
jlhogan, line 65: todo: remove console logs

client/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@ctoec/component-library": "0.0.35",
+    "@ctoec/component-library": "https://github.com/ctoec/component-library#update-sidenav-render-children",
```
bamader, line 6: Dumb question incoming, but we can still pin to a specific version for the audit tests, right? We'll just pin to a different version that has this merged in?

client/src/shared/models/db/Child.ts
```diff
@@ -13,7 +13,7 @@ export interface Child extends ObjectWithValidationErrors {
   lastName?: string;
   suffix?: string;
   birthdate?: Moment;
-  birthCertificateType: BirthCertificateType;
+  birthCertificateType?: BirthCertificateType;
```
bamader, line 16: Do we want this to be optional? I thought we wanted providers to have to specify a type. I guess the nullable in the schema accomplishes this, but is there a reason to prefer one approach over the other?

bamader (approved):
LGTM!

client/package.json
jlhogan: so my plan for merging this PR is:
1. land the component-lib PR
2. update version in package.json and publish a new version on NPM
3. update this PR to reference that version
4. merge this PR

client/src/shared/models/db/Child.ts
jlhogan: our general approach (as far as i understand it) is to allow all fields to be nullable, but enforce they have values with validations at the model level, not this DB level. This means we don't have to fake value for bulk uploaded data, and users will just be required to enter the data before they can submit to OEC.

client/src/components/Forms/Enrollment/Form.tsx
```diff
@@ -123,7 +122,7 @@ export const EnrollmentForm: React.FC<EnrollmentFormProps> = ({
           orgId={child.organization.id}
         />
       )}
-      {CancelButton}
+      {AdditionalButton}
```
mmazanec22, line 125: Would you mind making a note about the possible additional buttons this could end up being?

client/src/containers/CreateRecord/CreateRecord.tsx
```diff
@@ -18,7 +18,7 @@ type LocationType = Location & {
   };
 };
 
-const AddChild: React.FC = () => {
+const CreateRecord: React.FC = () => {
   const h1Ref = getH1RefForTitle();
   const { accessToken } = useContext(AuthenticationContext);
   const { state: locationState, hash } = useLocation() as LocationType;
```
mmazanec22, line 24: We prob need to change this to { organization, hash } now right?

mmazanec22 (approved):

client/src/containers/CreateRecord/CreateRecord.tsx
jlhogan: no, i believe what you pass as the second parameter in the `Link` ends up being the state. So here we had locationState = `{ state: { organization: {...}}`. The change made in the button means this works correctly.

client/src/components/Forms/Enrollment/Form.tsx
jlhogan: added in next branch!

client/src/containers/CreateRecord/CreateRecord.tsx
mmazanec22: Oh wow TIL, thanks for catching!

client/src/shared/models/db/Child.ts
bamader: Oh I see. That makes a lot more sense than trying to use the DB to force everything to be correct, because that's why we have validations. Thanks for clarifying!

Issue #388, opened 2020-09-28
Unique ID Setup: The system knows if I use SASIDs, a different unique ID or no unique ID to track children.


RyanHansz:
- at what level do we need to know this ? 
  - multi-org and/or org, so add a field to both
  - one field or two booleans

- add additional optional uniqueId field to child (in addition to optional SASID)

- make template reflect SASID or other unique ID

- UI will only show the user the field they need, or neither if we know they do not use one

PR #599, opened 2020-10-20
Initial pass at error modal

### LINKED ISSUE #154: Copy Edits to Data Definitions

### LINKED ISSUE #476: child demographics batch edit form

### LINKED ISSUE #486: Missing/Incorrect info in Batch Upload Modal
# [Design in Abstract](https://share.goabstract.com/4b27436d-2787-43c2-9253-52c7ec44ac79?collectionLayerId=c7db850b-608f-40ec-8c99-8af57ecfd7ae&mode=design)

When uploading a spreadsheet with missing/incorrect info, I see a modal that tells me how many records have errors and in what column.

Shows after the modal in #350 if the user sees both.

Screen Shot 2020-10-19 at 8 00 24 AM

bamader:
Creates an error modal that shows users which columns in a batch uploaded sheet have errors, and how many total errors there are. Changes map to take a `save` parameter so that we can create formatted Child objects without persisting them to the DB (leveraging `entityManager.create()`). Analyzes counts of each kind of error in the backend so that client can be kept as simple as possible. Also provides needed updates to the data definitions.

Update: Added a string formatting function to replace the `PLACEHOLDER` text we had in BatchEdit cards before. Added child identifiers/demographics to the forms that can be BatchEdit processed.

Closes #486 
Closes #154 
Closes #476


src/controllers/enrollmentReports/checkErrors.ts
```diff
@@ -0,0 +1,43 @@
...
+  const propertyNameToFormattedName = {};
+  let errorDict = {};
+
+  // Start the count of each type of error at 0 overall
+  cols.map((c) => {
+    errorDict[c.propertyName] = 0;
+    propertyNameToFormattedName[c.propertyName] = c.formattedName;
+  });
+
+  for (let i = 0; i < children.length; i++) {
```
mmazanec22, line 22: This is maybe a personal pref but this could be `children.forEach(child => { const { validationErrors } = child`.

mmazanec22 (approved):

client/src/containers/BatchEdit/listSteps.tsx
```diff
@@ -4,15 +4,37 @@ import {
...
+  { key: SECTION_KEYS.FAMILY, form: FamilyAddressForm },
+  { key: SECTION_KEYS.ENROLLMENT, form: EnrollmentFundingForm },
+];
+
+/**
+ * Util that determines which forms have errors for a particular child
+ * and creates a formated string to display in the BatchEdit cards
+ * below the name
+ */
+export const formatBatchEditPlaceholderString = (child: Child) => {
```
jlhogan, line 26: oops, i was just throwing the string "placeholder" in there as a literal placeholder. I would consider this a `BatchEditErrorDetailsString` ?

Also AC from ECE Reporter v1 was that this string include a count of the fields with missing info, not just the forms. But let's leave this for now and open a ticket to beef this up later

bamader:
@RyanHansz The error modal and data definitions updates have been deployed to QA! Let me know if things don't look quite as we want them on the data requirements page.

client/src/containers/Upload/TableColumns.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { Column } from '@ctoec/component-library';
+import { ErrorObjectForTable } from './Upload';
+
+/**
+ * Tabular column formatter that displays a dictionary of
+ * counts of validation errors as a formatted table within
+ * a batch upload modal.
+ */
+export const tableColumns: () => Column<ErrorObjectForTable>[] = () => {
```
jlhogan, line 10: nit: I would maybe make a new folder for ErrorsModal within `Upload` and put this and the modal in it? Or at least name this `ErrorsModalTableColumns`

client/src/containers/Upload/Upload.tsx
```diff
@@ -32,6 +44,10 @@ const Upload: React.FC = () => {
 
   const [userRosterCount, setUserRosterCount] = useState(undefined);
   const [checkReplaceDataOpen, setCheckReplaceDataOpen] = useState(false);
+  const [errorDict, setErrorDict] = useState<ErrorObjectForTable[] | undefined>(
```
jlhogan, line 47: nit: you don't need to explicitly add undefined as a type here, if you add the type parameter and then just leave it empty, it will be undefined. so:
```
const [errorDict, setErrorDict] = useState<ErrorObjectForTable[]>()
```

src/controllers/enrollmentReports/checkErrors.ts
jlhogan: agreed. I feel like perhaps you were inspired to write this for loop from the mapping stuff .. It's needed there because you can't exit early/break out of a forEach loop, and I wanted to be able to do that if an error was thrown while processing each child. But generally prefer the javascript array iterator funcs

src/routes/enrollmentReports.ts
```diff
@@ -44,6 +45,59 @@ enrollmentReportsRouter.get(
   })
 );
 
+/**
+ * /enrollment-reports/checkForErrors POST
```
jlhogan, line 49: nit: can we name this with snake case like the other route names? so `/check-for-errors` ? I also dig keeping url paths relatively short (antithesis of how i feel about all other names p much) so i would be down with `enrollment-reports/check`

src/routes/enrollmentReports.ts
```diff
@@ -84,10 +138,11 @@ enrollmentReportsRouter.post(
       // Ingest upload by parsing, mapping, and saving uploaded data
       try {
         const reportRows = controller.parseUploadedTemplate(req.file);
-        const reportChildren = await controller.mapAndSaveRows(
+        const reportChildren = await controller.mapRows(
           tManager,
           reportRows,
-          req.user
+          req.user,
+          true
```
jlhogan, line 145: nit: i dont love floating context-less booleans and have learned recently about throwing things into an `opts` dict, so they can be named and thus contextualized. makes reading code much easier. so this would be something like:
```
 const reportChildren = await controller.mapRows(
          tManager,
          reportRows,
          req.user
          req.user,
          { doSave: true }
)
```
I'll throw the function signature in the controller file in a comment as well if you're interested

src/controllers/enrollmentReports/map.ts
```diff
@@ -61,11 +69,12 @@ export const mapAndSaveRows = async (
...
  * @param source
  */
-const mapAndSaveRow = async (
+const mapRow = async (
   transaction: EntityManager,
   source: EnrollmentReportRow,
   userOrganizations: Organization[],
-  userSites: Site[]
+  userSites: Site[],
+  save: boolean
```
jlhogan, line 77: to avoid floating boolean, this could be:
```
const mapRow = async(
  transaction: EntityManager,
  source: EnrollmentReportRow,
  userOrganizations: Organization[],
  userSites: Site[]
  userSites: Site[],
  opts?: { save: boolean } = { save: false } // imo, false default makes sense 
```

and then pass `opts.save` into the internal mapping functions (I think those are fine with floating boolean, they're very much contextualized)

jlhogan (approved):
woohoo!!! so seems like the instantiating without saving did indeed work!! :tada: 
no blockers, a few nitpicky comments

jlhogan:
oh and also this may have some conflicts with the changes i just merged -- sorry! shouldn't be too bad to resolve though.

client/src/containers/BatchEdit/listSteps.tsx
bamader: Lol yeah that's a much better name than something that literally still keeps 'placeholder' in the title 😂

client/src/containers/Upload/Upload.tsx
bamader: I have yet to master how explicit I need to be with typing in all the various circumstances that come up 😝

RyanHansz:
@bamader
Looks like the error modal is only picking up birthdate and gender in this upload that has missing income determinations. Also https://github.com/ctoec/data-collection/issues/611 is probably resulting in mismatched errors as well. 
Here's the file I'm using: [ECE Data Collection Template (5).xlsx](https://github.com/ctoec/data-collection/files/5425827/ECE.Data.Collection.Template.5.xlsx)

Otherwise, the data definitions look good, the contract space rows got misaligned:
Screen Shot 2020-10-22 at 4 58 15 PM

src/routes/enrollmentReports.ts
bamader: Short and snake is how it shall be!

RyanHansz:
Edit: Data definitions:
SASID -> 
"SASID or Unique Identifier"

bamader:
@RyanHansz 

Re: missing income determinations and enrollments: I wonder if those are actually linked by the same issue caused by whatever is making #611 weird. Income determinations are also a nested sort of validation, so I wonder if cleaning up the way we do that on the back-end in that ticket will solve the error here (since I think it's parsing based)?

Re: SASID field name: I think we have another ticket open that's responsible for doing all the stuff to add a unique identifier field. I talked to Julia and Kyle and think that going forward, we're going to want both a SASID field _and_ a unique identifier field to help with data matching, and then just show the appropriate field to a user based on whether their org uses SASID or not. If it's okay with you, I might recommend we push changing the SASID field title into that ticket, since it seems like a part of making both fields work?

Re: contract row misalignment: ... I have literally no clue what caused this but will be delving in after I rewrite some function signatures!

src/routes/enrollmentReports.ts
bamader: Oh yeah that's how we do our API call stuff too (like with accessToken and rawBody for the json). You're right, if you're not looking at the function signature then seeing the boolean actually paired with something makes it a _lot_ more interpretable. Thanks for the signature below, too!

RyanHansz:
Good point, SASID / unique id is in #388
Otherwise, shall we revisit the modal when #611 is resolved to make sure it fixes this as well?

bamader:
Yeah I think revisiting is good! I'm gonna open up a ticket about something Julia commented on in a different place, so I'll make sure and note that we need to check the modal again, too. In the mean time, I'll try to make our text not weirdly spaced.

PR #600, opened 2020-10-20
572 fake file

mmazanec22:
Still need to add images to meet design AC, but I made so many changes that I kind of want to merge this sooner rather than later and save that for cleanup.

PR #604, opened 2020-10-20
573: Create funding without any values + tests

### LINKED ISSUE #573: Update batch upload parsing to account for ALL funding space values
ensure parsing handles:
- acronyms
- full names
- numbers

If we can't match a funding space to funding line item, still create an enrollment with empty funding. 

- Funding space will have to be optional when uploading records. 
- Add validation that funding space is not empty when submitting

jlhogan:
plus made some of the `parse` functionality more modular, with the intention of writing tests (soon, but not in the this PR -- big enough)

closes #573


src/controllers/enrollmentReports/map.test.ts
```diff
@@ -0,0 +1,281 @@
...
+} from '../../../client/src/shared/models';
+import { FUNDING_SOURCE_TIMES } from '../../../client/src/shared/constants';
+import moment from 'moment';
+import { EntityManager } from 'typeorm';
+
+describe('controllers', () => {
+  describe('enrollmentReports', () => {
+    describe('lookUpOrganization', () => {
+      it('returns the organization if only one organization', () => {
+        const inputOrg = { id: 1 } as Organization;
```
mmazanec22, line 27: Once I have #600 merged we should be able to use that fake data for tests too

src/controllers/enrollmentReports/map.test.ts
```diff
@@ -0,0 +1,281 @@
...
+        ]),
+      ])(
+        'can parse any AgeGroup value: %s returns %s',
+        (inputAgeGroup, expectedAgeGroup) => {
+          const parsed = mapEnum(AgeGroup, inputAgeGroup);
+          expect(parsed).toEqual(expectedAgeGroup);
+        }
+      );
+
+      it.each([
```
mmazanec22, line 214: Whoa

src/controllers/enrollmentReports/map.ts
```diff
@@ -153,7 +149,7 @@ const lookUpSite = (
  * respective child's race (in which case, assume not disclosed).
  * @param source
  */
-const raceIndicated = (source: EnrollmentReportRow) => {
+export const getRaceIndicated = (source: EnrollmentReportRow) => {
```
mmazanec22, line 152: I changed the race validation in #600 and made it allow not specified-- we want that right?  Also maybe that should be broken into a helper func and used here too

mmazanec22 (approved):

src/controllers/enrollmentReports/map.test.ts
jlhogan: wooooooo!!!!!!!!!!!!!

src/controllers/enrollmentReports/map.ts
jlhogan: i saw that,! was imagining that would land first and then here I would reduce to a single util that we use everywhere for race determination

client/src/shared/constants/FundingSourceTimes.ts
```diff
@@ -3,6 +3,8 @@ import { FundingSourceTime } from '../models/FundingSourceTime';
 
 //  This enumerates all of the valid funding source/funding time combinations,
 //  as well as all of the acceptable formats for each associated funding time
+// TODO: should fulltime and fullday roll up to the same "time"?
+// TODO: should parttime and partday roll up to the same "time"?
```
bamader, line 7: Potentially dumb question, but what do you mean "roll up" here? I think different funding sources use full day and full time, so it might be a differentiator between them? But if you're saying we've already captured that difference in the source, then yeah maybe we don't need to track them separately here.

src/controllers/enrollmentReports/map.test.ts
bamader: This is intense 👀

src/controllers/enrollmentReports/map.ts
```diff
@@ -289,47 +284,27 @@ const mapEnrollment = (
 /**
  * Create a Funding object from FlattenedEnrollment source,
```
bamader, line 285: I just noticed this comment still references `FlattenedEnrollment` objects and I had a mild heart attack thinking back to those days

bamader (approved):
LGTM!

jlhogan:
@RyanHansz (and whoever else should be tagged for QA -- couldn't find a phoebe-seeming github account) things specifically to check for here that have changed:
- when a user uploads funding data but not fundingSource, an empty funding should be created and when they go to edit it, they should see the funding source radio button with the rest of the funding options nested
- if a user attempts to change enrollment or funding when the active funding is missing funding source and contract space, they are shown text that says in essence: you need to add this information before you can change enrollment/funding. I made up that language, it needs love, but needed to have an alternative here since we can't end a current funding that doesn'thave a funding source
- if a user attempts to withdraw a record with missing information they're shown text in the modal that says you can't withdraw a record with missing information

RyanHansz:
For the unable to withdraw state, I think we can repurpose the missing/incorrect info alert:
Screen Shot 2020-10-22 at 8 59 02 AM

And for errors in uploading funding, I propose this alert be shown when the user opens the record:
Screen Shot 2020-10-22 at 8 56 55 AM

jlhogan:
@RyanHansz for withdrawal, do you mean pop that error thing in the modal? Or display that error when they click withdraw instead of opening the modal? 

For now, i'm going to punt on differentiating the error messages b/c kind of not low lift if thats ok. 
will update can't change enrollment copy!

RyanHansz:
I would put the withdraw alert in the normal place above the header so we don't introduce a new alert style. 

And I'll make a separate ticket for the single vs. multiple error messages

jlhogan:
So, the modal does NOT pop up and the error alert displays in the normal place

RyanHansz:
@jlhogan yes, that's right

PR #606, opened 2020-10-21
Correct file download

mmazanec22:
Discovered this bug while writing tests-- needed to move the example download to an unauthenticated route.


bamader (approved):
Good catch!

RyanHansz:
Non-blocking question: Is there a way to separate different sample files? For example, the sample data we'd show programs probably shouldn't have errors in it, but we need one with errors to test.

mmazanec22:
Yes @RyanHansz I intentionally left the door open for that when I made the fake data.  I think we need to think about 1) how we want to access that as a team-- maybe a staging-only page that just has some links? 2) what files we want to create.  But yeah it should be pretty trivial to make that happen now!

jlhogan (approved):
@mmazanec22 If you are up for it, would be awesome to create that page (truly can be whatever) and update the link in the getting-started page to provide correct, complete example. But obvi not blocking here. can we capture that in a ticket tho to make sure it happens?

RyanHansz:
Added #616, feel free to add more details!

PR #615, opened 2020-10-22
Implement single site user limitations

### LINKED ISSUE #538: Single Site user: I can only see and edit data for a single site of a multi-site provider.
# [Design](https://share.goabstract.com/dec9b5f8-a16a-4884-8c41-ee3c2c5301b0?collectionLayerId=95b78637-c964-4d83-b7af-549c06401153&mode=design)
Single Site User

– I cannot submit data to OEC (only a user with access to all sites can)
– I do not see capacity of funding spaces, only # spaces filled 
– The title changes to the name of the site, with the provider name above it
– If I batch upload, I can only replace/upload data for the site I have access to

bamader:
Implements access and UI restrictions for users with access to a single site that are part of an organization with multiple sites. As a note, I put the determination of whether the user was a single-site user of a multi-site provider in the backend, because it seems like something that user systems shouldn't have access to (i.e. if a user _does_ only have access to one site, we don't want the code that determines all the other sites their org provides for showing up in the front end). 

Closes #538


client/src/containers/Roster/Roster.tsx
```diff
@@ -54,6 +54,24 @@ const Roster: React.FC = () => {
     activeSiteId
   );
 
+  // Use state to determine whether current user is a single-site
+  // accessor of a provider with multiple sites--limit options if so
+  const [isSSUserOfMSProvider, setIsSSUser] = useState<boolean>(false);
```
jlhogan, line 59: so I think we can do this a bit more simply -- my thought was this: 
any user who is given permissions via "sitePermissions" is NOT an organization-level user and thus cannot do things like submit data, whether or not they have access to only one site, or to multiple sites. 

So it think we can remove this, and just look at the user object to see if they have any organization permissions, or if they only have site permissions.  Does that make sense?

jlhogan (changes requested):

client/src/containers/Roster/Roster.tsx
bamader: I think it does, yeah. Lemme just see if I'm interpreting some fields correctly: a `User` object has `sites` and `organizations`, both of which are populated (on the backend, in the `authenticate.ts` file) via `sitePermissions` and `orgPermissions`. So the sites and orgs in those fields on a User are explicitly those the user can actually edit data for, rather than just sites/orgs the user is associated with? Based on how they're used on the roster page, I was thinking the `sites` and `organizations` arrays consisted of those the user was affiliated with (otherwise, a user could have elements in their `sites` array but nothing in their `organizations` array, and then that would make deciding how to format the header weird).

Follow-up (and this may be a question for Ryan): the ticket has a few characteristics in the AC for single-site users of multi-site providers: can't submit, can't see capacity, differently formatted page header, batch upload restrictions. Do we want to apply those four things to _all_ users who don't have orgPerms? Or just users who don't have orgPerms but also have access to only one site? If it's the latter case, I think we do need to know specifically whether the user is a single-site accessor of a multi-site organization. (This might be what you mean in your first paragraph with the other things part, but I just want to double check.)

Sorry for the long reply!

bamader: @RyanHansz Wanted to get your eyes on the second part of this question. Specifically, do we want the four restrictions in the AC of #538 to apply to all users who don't have org-level permissions?

RyanHansz: Yes, all the restrictions apply. 
"– If I batch upload, I can only replace/upload data for the site I have access to" would be (sites) for a multi-org user.

jlhogan: and my clarification/technical implementation addendum is:
-  an "all" site user is actually an "organization" user. we will grant permission/access to "all" sites by granting organization-level permissions to that user.
- a user who arbitrarily has access to "all" sites for an org via site permissions is not an edge case we care about.

bamader: Awesome. Thank you both! Can def simplify this then.

bamader: I like the idea of adding a convenience var to the model that captures access type. It's simple, provides the client the least information it needs to make this change, and keeps things on the backend where (to me at least) it makes the most sense to have that determination.

bamader: Yeah the terminology of the variable names was throwing me off a bit because I didn't realize they were convenience fields until I dug into authenticate and looked at how they got generated. I probs won't do anything with the cache because if we're gonna eventually swap that out with a package, I think it makes more sense to do the access type convenience field instead.

client/src/containers/Roster/rosterUtils.tsx
```diff
@@ -159,7 +159,8 @@ export function getChildrenByAgeGroup(
 
 export function getAccordionItems(
   childrenByAgeGroup: ChildrenByAgeGroup,
-  showOrgInTables: boolean
+  showOrgInTables: boolean,
+  opts: { hideCapacity: boolean } = { hideCapacity: false }
```
jlhogan, line 163: can we put `showOrgInTables` into the `opts` object as well ?

src/middleware/authenticate.ts
```diff
@@ -111,6 +111,9 @@ const getUser = async (wingedKeysId: string) => {
     ])
   );
 
+  // Determine access pattern level of the user
+  user.accessType = allOrgIds.length === 0 ? 'site' : 'organization';
```
jlhogan, line 115: Can we make this determination based off the length of `orgPermissions` as opposed to `allOrgIds`? JIC some time in the future we want to populate site-level users with their parent org (could see that being convenient); wouldn't want access to get messed up from some change like that.

client/src/containers/Roster/Roster.tsx
```diff
@@ -140,19 +145,30 @@ const Roster: React.FC = () => {
...
+              {organizations[0].providerName}
+            </div>
+          )}
           {h1Content}
         </h1>
         <p className="font-body-xl margin-top-1">
           {/* TODO: should this count be just for the thing showing or for all of them? */}
           {childrenCache.loading
             ? 'Loading...'
+            : isSiteLevelUser
```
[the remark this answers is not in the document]
bamader: They definitely can. This conditional is there because in the design image of #538 , there's no "... at X sites" for the user. However, that issue was under the assumption we were only changing the UI for single site users rather than all site-level users. So question for @RyanHansz : can we still show the "...enrolled at X sites" text?

client/src/containers/Roster/Roster.tsx
```diff
@@ -183,9 +199,13 @@ const Roster: React.FC = () => {
         />
         {/* TODO: change when we figure out multi-site entity */}
         <Button
-          text="Send to OEC"
+          text={
+            isSiteLevelUser
+              ? 'Organization permissions required to submit'
```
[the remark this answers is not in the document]
bamader: That would make sense to me, and now that you say that I agree. Will wait for Ryan to confirm though.

src/middleware/authenticate.ts
bamader: Yeah we def can. I just did `allOrgIds` in case there were relevant permissions stemming from the `orgsFromCommunities`.

client/src/containers/Roster/Roster.tsx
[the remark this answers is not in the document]
RyanHansz: Agree! Just hide the "submit to OEC" button in the bottom bar

client/src/containers/Roster/Roster.tsx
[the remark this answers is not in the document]
jlhogan: my gut: If org user OR sites.length > 1, show. else, hide

client/src/containers/Roster/Roster.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect } from 'react';
+import React, { useContext, useEffect, useState } from 'react';
```
jlhogan, line 1: cleanup unused imports

client/src/containers/Roster/Roster.tsx
```diff
@@ -15,7 +15,7 @@ import { getH1RefForTitle } from '../../utils/getH1RefForTitle';
 import { AddRecordButton } from '../../components/AddRecordButton';
 import { useHistory, useLocation } from 'react-router-dom';
 import DataCacheContext from '../../contexts/DataCacheContext/DataCacheContext';
-import { apiPut } from '../../utils/api';
+import { apiGet, apiPut } from '../../utils/api';
```
jlhogan, line 18: same, cleanup

client/src/containers/Roster/Roster.tsx
```diff
@@ -140,19 +144,30 @@ const Roster: React.FC = () => {
       nestedActiveId: activeSiteId,
       activeId: activeOrgId,
     };
+  } else if (isSiteLevelUser) {
+    // User has no organization permissions, is restricted on page
+    h1Content = sites[0].siteName;
+    tabNavProps = undefined;
```
jlhogan, line 150: i'm not sure tabNav should disappear site-level user -- the org nav should, but user should still have sites nav (unless user has access to only one site)

client/src/containers/Roster/Roster.tsx
```diff
@@ -140,19 +144,30 @@ const Roster: React.FC = () => {
...
+              {organizations[0].providerName}
+            </div>
+          )}
           {h1Content}
         </h1>
         <p className="font-body-xl margin-top-1">
           {/* TODO: should this count be just for the thing showing or for all of them? */}
           {childrenCache.loading
             ? 'Loading...'
+            : user?.accessType === 'organization' || sites.length > 1
```
jlhogan, line 169: i think perhaps this ternary is backwards? If the user is organization access or the user has access to more than one site, then include the site in the string. Also, can avoid double nested ternary with something like, but that's just a preference :) 
```
{childrenCache.loading
  ? 'Loading...'
  : `${childrenCache.records.length} children enrolled at ${(user?.accessType === 'organization' || user?.sites.length > 1) ? `${sites.length} sites` : ''}`
```

jlhogan (approved):
some nits but overall LGTM !

client/src/containers/Roster/Roster.tsx
bamader: Good point, forgot to change this conditional from when I was trying to code the specific case of a single-site user. Will update.

client/src/containers/Roster/Roster.tsx
```diff
@@ -140,20 +144,45 @@ const Roster: React.FC = () => {
       nestedActiveId: activeSiteId,
       activeId: activeOrgId,
     };
+  } else if (organizations.length === 1) {
```
mmazanec22, line 147: The base case is a single site user (see line 108) so this should be unnecessary-- we can just change the default h1 to organization provderName

client/src/containers/Roster/Roster.tsx
```diff
@@ -140,20 +144,45 @@ const Roster: React.FC = () => {
...
+    h1Content = organizations[0].providerName;
+    // Will only have a single site if we reach this conditional
+    tabNavProps = undefined;
+  } else if (isSiteLevelUser) {
+    // User has no organization permissions, is restricted on page
+    h1Content = sites[0].siteName;
+    if (sites.length == 1) tabNavProps = undefined;
```
mmazanec22, line 154: If we leave the default as single site, then there's no need to set tab nav props to undefined here, since that's what they default to on line 115.  This can just be
```
if (sites.length > 1) {
  tabNavProps = {
    blablabla
  }
}
```

mmazanec22 (approved):

PR #639, opened 2020-10-26
550 radio

### LINKED ISSUE #550: Make component library not assign ids by default to radio buttons (or anywhere else) bc ids with spaces aren't valid and they break accessibility
Giving this 2 points because figuring out the relationship between form things in the component library is a bit complex-- but basically we just need to pass the parent ID of a RadioButtonGroup to its individual radio buttons as props (and modify it for each one so they're all unique) so they can associate the label and input correctly

mmazanec22:
Closes #550 

This is a refactor of most of our forms, so I'm gonna push this one to QA and run e2e testing against it (yay) before merging it!  Any manual review that anyone is inspired to do would also be nice.

### DIFF
--- a/client/src/components/Forms/ChildInfo/Fields/DisabilityServices.tsx
+++ b/client/src/components/Forms/ChildInfo/Fields/DisabilityServices.tsx
@@ -3,14 +3,12 @@ import {
   RadioButtonGroup, RadioOptionInForm,
 } from '@ctoec/component-library';
 import { Child } from '../../../../shared/models';
-import { getValidationStatusForField } from '../../../../utils/getValidationStatus';
+import { getValidationStatusForFields } from '../../../../utils/getValidationStatus';
 import { UNKNOWN } from '../../../../shared/constants';
 
 export const DisabilityServices: React.FC = () => {
 
-
-  // TODO: further refactor radiogroup to take common option props?
-  const getRadioButtonProps = ({ label, id }: { label: string, id: string }): Pick<RadioOptionInForm<Child>, 'getValue' | 'parseOnChangeEvent' | 'preprocessForDisplay'> => ({
+  const getRadioButtonProps = ({ label, id }: { label: string, id: string }): RadioOptionInForm<Child> => ({
     getValue: (data) => data.at('receivesDisabilityServices'),
     parseOnChangeEvent: (e) => {
       if (e.target.value !== 'disability-unknown')
@@ -21,7 +19,11 @@ export const DisabilityServices: React.FC = () => {
       if (data === true) return 'disability-yes';
       else if (data === false) return 'disability-no';
       else return 'disability-unknown';
-    }
+    },
+    id,
+    text: label,
+    value: id,
+    name: id
   })
 
   return <RadioButtonGroup<Child>
@@ -42,7 +44,12 @@ export const DisabilityServices: React.FC = () => {
         id: 'disability-unknown',
         label: UNKNOWN,
       },
-    ].map(o => getRadioButtonProps(o)}
-    status={getValidationStatusForField}
-  />
+    ].map(o => getRadioButtonProps(o))}
+    status={(data) =>
+      getValidationStatusForFields(
+        data,
+        ['receivesDisabilityServices'],
+        { message: 'Disability services status is required for OEC reporting.' }
+      )
+    } />
 };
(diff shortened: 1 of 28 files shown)


client/src/components/Forms/ChildInfo/Fields/DisabilityServices.tsx
```diff
@@ -1,50 +1,24 @@
...
-} from '@ctoec/component-library';
-import { Child } from '../../../../shared/models';
-import { getValidationStatusForField } from '../../../../utils/getValidationStatus';
 import { UNKNOWN } from '../../../../shared/constants';
+import { Child } from '../../../../shared/models';
+import { TripleBooleanRadio } from '../../../TripleBooleanRadio';
 
 export const DisabilityServices: React.FC = () => (
-  <FormField<Child, RadioButtonGroupProps, boolean | null>
+  <TripleBooleanRadio<Child>
```
jlhogan, line 7: Is tripleBooleanRadio a form field? I guess i missed the PR for that, but if it is a form field i think that should be expressed in the name? Does it use `field` prop to find/update data in the object? Why opt for magic string instead of type safe object driller?

mmazanec22: TripleBooleanRadio is a component added to this project in this PR.  It's not a form field bc radio group is no longer a form field.  There isn't any more magic string-ness going on here than there was before, it's just condensed into one place for any radio button group where we're using the same logic

client/src/components/TripleBooleanRadio.tsx
```diff
@@ -0,0 +1,68 @@
...
+export type TripleBooleanRadio<T> = {
+  field: keyof T;
+  id: string;
+  legend: string;
+  trueOption: TripleBooleanRadioOption;
+  falseOption: TripleBooleanRadioOption;
+  unknownOption: TripleBooleanRadioOption;
+};
+
+// Future podcast title??
```
jlhogan, line 19: :rofl:

client/src/components/Forms/ChildInfo/Fields/DisabilityServices.tsx
jlhogan: don't know why/howi couldn't find it when i first reviewed this PR...... please excuse my brain!

jlhogan (approved):

JordanGuinn (approved):
Looks reasonable to me

client/src/components/Forms/Enrollment/Fields/Site.tsx
```diff
@@ -24,27 +17,25 @@ export const SiteField = <T extends Enrollment | ChangeEnrollment>({
...
+          parseInt(e.target.value) || null,
         value: `${site.id}`,
+        id: `site-${site.id}`,
       }))}
-      status={(data, _, props) =>
-        getValidationStatusForField(
-          enrollmentAccessor(data),
-          enrollmentAccessor(data).at('site').path,
-          props
+      status={(_, dataDriller) =>
```
kmagida110, line 34: How is this related to the radio button or is it not?

kmagida110 (approved):
One quick question but good to go

client/src/components/Forms/Enrollment/Fields/Site.tsx
mmazanec22: The radio button group was being treated as a field, which makes sense data-wise but is kinda messed up html-wise and props-passing-wise bc radio groups are actually a fieldset under the hood.  This PR (and the component library one a few days ago) makes it so that this is a fieldset instead, and therefore needed to take the fieldset status function.  I added the data driller as something passed to the status func via the component library fieldset status func and made a note that it should align better with the field status func in the future.  The data driller is used here to get the appropriate parent object and pass it to the status func.

client/src/components/TripleBooleanRadio.tsx
```diff
@@ -0,0 +1,68 @@
...
+  label: string;
+};
+
+export type TripleBooleanRadio<T> = {
+  field: keyof T;
+  id: string;
+  legend: string;
+  trueOption: TripleBooleanRadioOption;
+  falseOption: TripleBooleanRadioOption;
+  unknownOption: TripleBooleanRadioOption;
```
bamader, line 16: I love this so much ❤️

bamader (approved):
Great stuff! Love that triple boolean refactor.

PR #641, opened 2020-10-27
465 income det validation

### LINKED ISSUE #465: Income determination not required for children living with foster family.
On income determination page:
8 4 Child record  – Family Income
- If a child is living with a foster family, then income determinations are not required. 
- Create backend validation - Ensure there is an income determination if not a foster child
- Add the UI alert for foster youth

mmazanec22:
Closes #465

### DIFF
--- a/client/src/components/Forms/FamilyIncome/Form.tsx
+++ b/client/src/components/Forms/FamilyIncome/Form.tsx
@@ -2,7 +2,6 @@ import React, { useContext, useState } from 'react';
 import { Child, IncomeDetermination } from '../../../shared/models';
 import {
   getValidationStatusForFields,
-  getValidationStatusForField,
 } from '../../../utils/getValidationStatus';
 import { RecordFormProps } from '../types';
 import AuthenticationContext from '../../../contexts/AuthenticationContext/AuthenticationContext';
@@ -12,7 +11,6 @@ import {
   Form,
   FormSubmitButton,
   FormFieldSet,
-  Alert,
 } from '@ctoec/component-library';
 import {
   HouseholdSizeField,
@@ -49,9 +47,9 @@ export const doesFamilyIncomeFormHaveErrors = (
   const incomeDeterminationsHaveError = child?.family?.incomeDeterminations
     ?.length
     ? !!getValidationStatusForFields(
-        child.family.incomeDeterminations,
-        incomeDeterminationFields
-      )
+      child.family.incomeDeterminations,
+      incomeDeterminationFields
+    )
     : false;
 
   return familyHasIncomeDeterminationError || incomeDeterminationsHaveError;
@@ -91,10 +89,10 @@ export const FamilyIncomeForm: React.FC<FamilyIncomeFormProps> = ({
 
   const determination = (type === 'edit'
     ? (incomeDeterminationId
-        ? child?.family?.incomeDeterminations?.find(
-            (d) => d.id === incomeDeterminationId
-          )
-        : idx(child, (_) => _.family.incomeDeterminations[0])) || {}
+      ? child?.family?.incomeDeterminations?.find(
+        (d) => d.id === incomeDeterminationId
+      )
+      : idx(child, (_) => _.family.incomeDeterminations[0])) || {}
     : {}) as IncomeDetermination;
 
   const createDetermination = async (updatedData: IncomeDetermination) =>
--- a/client/src/components/Forms/FamilyIncome/Form.tsx
+++ b/client/src/components/Forms/FamilyIncome/Form.tsx
@@ -83,7 +83,9 @@ export const FamilyIncomeForm: React.FC<FamilyIncomeFormProps> = ({
   const { accessToken } = useContext(AuthenticationContext);
 
   if (child?.foster) {
-    // This shows up for new child
+    // New child is and batch edit both use this form directly
+    // So this alert will show for those two forms
+    // Edit child conditionally shows this form, so this alert is in that container too
     return <FosterIncomeNotRequiredAlert />;
   }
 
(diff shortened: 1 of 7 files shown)


client/src/components/Forms/FamilyIncome/AlertFosterIncomeNotRequired.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { Alert } from '@ctoec/component-library';
+
+export const AlertFosterIncomeNotRequired = () => (
```
jlhogan, line 4: nit: to me this name sounds like an action, i'd prefer `FosterIncomeNotRequiredAlert`

client/src/components/Forms/FamilyIncome/Form.tsx
```diff
@@ -64,6 +74,11 @@ export const FamilyIncomeForm: React.FC<FamilyIncomeFormProps> = ({
   const [loading, setLoading] = useState(false);
   const { accessToken } = useContext(AuthenticationContext);
 
+  if (child?.foster) {
+    // This shows up for new child
```
jlhogan, line 78: what does this mean?

src/entity/Family.ts
```diff
@@ -42,9 +47,12 @@ export class Family implements FamilyInterface {
   @Column({ nullable: true, default: null })
   homelessness?: boolean;
 
-  // TODO: only validate if child is not a foster child
   @OneToMany(() => IncomeDetermination, (det) => det.family)
   @ValidateNested({ each: true })
+  @MinLength(1)
```
jlhogan, line 52: wooooo was going to ask if we could throw this in here, but you already did !

jlhogan (approved):

client/src/components/Forms/FamilyIncome/Form.tsx
mmazanec22: I had to add this in two places bc of how we're conditionally showing forms in edit child-- can make this comment clearer

src/entity/Family.ts
jlhogan: Just checked and this does not seem to validate the length of the nested array -- it's a string length validation. To validate array length, use `@ArrayMinSize(1)`

jlhogan: ok actually, i'm unable to get nested family validations working at all?

src/entity/Family.ts
```diff
@@ -42,9 +47,12 @@ export class Family implements FamilyInterface {
...
+  // If one of the children is not a foster child, must have at least one income det
+  // This would break if we actually had a one to many between family and children, but we create a new family object for each child
+  @ValidateIf((family) => family.children?.some((c: Child) => !c.foster))
```
jlhogan, line 55: as far as I can tell, this does _not_ actually work, because children is not populated on the family object

jlhogan (changes requested):

src/entity/Family.ts
jlhogan: Kept playing around with this because was interested and what I cam up with is HACKY but also it WORKS :) 
in Child.ts:
```
@ValidateNested()
@ValidateIf((child) => { if(child.family) child.family.childIsFoster = child.foster; return true})
@ManyToOne(() => Family)
family?: Family;
```
in Family.ts:
```
@OneToMany(() => IncomeDetermination, (det) => det.family)
@ArrayMinSize(1)
@ValidateIf((family) => {const childIsFoster = family.childIsFoster; delete family.childIsFoster; return !childIsFoster})
incomeDeterminations?: Array<IncomeDetermination>;
```

jlhogan: improvement: write this as a custom decorator that's called something other than `ValidateIf`

JordanGuinn: Are there any situations where the `Family` object wouldn't be present on `Child`?  Also I'm legit AMAZED that attributes set in one validator persist when running another 😎

jlhogan: a lil perusal of the source code shows that the actual object instance is passed around everywhere, so just taking advantage of js pass by reference to update the old boy :) 

family is not present on the child when we do the `getChildren` call, which means family  + income determinations are not validated in the roster view. So ACTUALLY, if family is not present, we need to go pull it + income determinations from DB and validate those  OR just use the full query that we use for `getChildById` in the `getChildren` DB fetch (probably the latter -- easier and not really worth optimizing to save those like KB of data per record prob)

jlhogan: Actually i'm just gonna make that change in my branch b/c i need it to resolve a funky thing in batch edit :)

mmazanec22: This appears to be working for me as is :thinking:

src/entity/Child.ts
```diff
@@ -134,6 +134,10 @@ export class Child implements ChildInterface {
   receivesDisabilityServices?: boolean;
 
   @ValidateNested()
+  @ValidateIf((child) => {
+    if (child.family) child.family.childIsFoster = child.foster;
```
JordanGuinn, line 138: Would there be any potential situation where one child in a family is in foster care while the rest of the children are not?  Or vice versa?  

Wanna confirm that this doesn't introduce any possible edge cases or anything.

jlhogan: if we're gonna overload this validateIf, definitely want to comment !! 

re jordan's comment -- that's definitely a _potential_ situation, but right now there's a one to one child-family relationship (every child gets a new family even if in reality the children are siblings) so doesn't matter.
But even if we were reusing families correctly, and a family had foster and not foster children, this would work as expected.
But allllllso really i dont think thats an edge case we need to worry about.

src/entity/Family.ts
```diff
@@ -42,9 +47,14 @@ export class Family implements FamilyInterface {
   @Column({ nullable: true, default: null })
   homelessness?: boolean;
 
-  // TODO: only validate if child is not a foster child
   @OneToMany(() => IncomeDetermination, (det) => det.family)
   @ValidateNested({ each: true })
+  @ArrayMinSize(1)
+  @ValidateIf((family) => {
```
jlhogan, line 53: comment!

jlhogan (approved):
Add comments to explain the funky usage of validateIf but otherwise all good!

JordanGuinn (approved):

PR #644, opened 2020-10-28
Historical updates

### LINKED ISSUE #591: Identify historical  records in a batch upload (match on child name, DOB, unique id if exists)
- As a user, I can upload multiple rows of data to add historical enrollment data to a child's record.
  - New enrollments
  - Withdrawn enrollments
  - Changes in site
  - Changes in funding
  - Changes in age group
  - Ended funding
  - Started funding
- When I upload data, the rows must have the same identifiers. perfect match only
- Only changes to enrollment and funding are tracked, all other fields will use only the most recent data.
- I must add at least name, DOB, Unique ID (if exists) to all rows with enrollment updates. 
- After uploading, I can see historical enrollments in the enrollment and funding page of the child's record. 
- If there are errors, I will eventually be able to see them in #486 
- For a sample upload, see:
https://docs.google.com/spreadsheets/d/1smR7pz57ALs3689BI1Z17rvbr4WrXhZ5Dz_em8Cyy1w/edit#gid=0

bamader:
Implements historical enrollment/funding tracking within Batch Upload. Multiple rows per child can now be provided in a single sheet, with the most recent row providing demographic/identifying information and subsequent rows adding/updating enrollments and fundings. 

Closes #591


src/controllers/enrollmentReports/map.ts
```diff
@@ -88,36 +181,82 @@ const mapRow = async (
...
-    source,
-    site,
-    child,
-    save
-  );
+  let child = isChildUpdate(source, processedChildren);
+  const isVisistedChild = child !== null;
+
+  // Case where this row creates a brand new child
+  if (!isVisistedChild) {
```
mmazanec22, line 188: Can the brand new vs updated logic be broken out into two separate functions?  I think that would be a little easier to follow.  So here it would just be like
```
if (isVisitedChild) {
  updateChild(whatever params)
  return null
}
createNewChild(whatever params)
return child
```

mmazanec22 (approved):
Aside from some nitpicking on for loops vs js array methods and modularity, woo!

src/controllers/enrollmentReports/map.ts
```diff
@@ -26,6 +26,94 @@ import { BadRequestError, ApiError } from '../../middleware/error/errors';
...
+ * information has already been seen, out of all children
+ * that have thus far been parsed.
+ * @param row
+ * @param processedChildren
+ */
+export const isChildUpdate = (
+  row: EnrollmentReportRow,
+  processedChildren: Child[]
+) => {
+  for (let i = 0; i < processedChildren.length; i++) {
```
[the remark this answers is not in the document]
bamader: You're absolutely right. When I was initially doing this, I was trying to use for loops because I thought I needed to use `Promise.resolve` and `Promise.all` to handle some things, but Julia was able to help me reformulate it in a much cleaner way, and now I can strip out some of these loops!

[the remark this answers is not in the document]
bamader: (My promise-fu is weeeeeeak 😛 )

src/controllers/enrollmentReports/map.ts
```diff
@@ -26,6 +26,94 @@ import { BadRequestError, ApiError } from '../../middleware/error/errors';
...
+};
+
+/**
+ * Determine whether a child with the given identifying
+ * information has already been seen, out of all children
+ * that have thus far been parsed.
+ * @param row
+ * @param processedChildren
+ */
+export const isChildUpdate = (
```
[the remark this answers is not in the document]
bamader: Honestly since I can reformulate the function as a single line, I honestly don't even think it needs to be a function anymore. Scrapping it altogether because I think the code is just as readable without introducing extra code!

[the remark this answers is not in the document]
bamader: Actually I lied, I like the way it looks with the function because I think the code reads more like plain language. Will make naming change and keep it.

src/controllers/enrollmentReports/map.ts
```diff
@@ -26,6 +26,94 @@ import { BadRequestError, ApiError } from '../../middleware/error/errors';
...
   'You uploaded a file with missing information.\nProvider name is required for every record in your upload. Make sure this column is not empty.';
 
+/**
+ * Determine whether a given enrollment report row from an
+ * uploaded sheet has demographic/identifier information matching
+ * a given child.
+ * @param child
+ * @param other
+ */
+export const isIdentifierMatch = (
```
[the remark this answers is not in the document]
bamader: Definitely. There's enough going on in `map` that these can be moved to a new utils file.

src/controllers/enrollmentReports/mapUtils.ts
```diff
@@ -0,0 +1,83 @@
...
+ * dates at the same site.
+ * @param row
+ * @param child
+ */
+export const getExistingEnrollmentOnChild = (
+  row: EnrollmentReportRow,
+  child: Child
+) => {
+  if (!child.enrollments) return null;
+  console.log('CHILD HAS ENROLLMENTS IN GET EXISTING');
```
JordanGuinn, line 51: nit - we can probably remove all of these all caps log statements 😅

src/controllers/enrollmentReports/map.ts
```diff
@@ -366,7 +456,7 @@ export const mapFunding = async (
     });
     fundingSpace = fundingSpaces.find((space) => space.time === fundingTime);
 
-    // If no direct match on time and source === CDC, look for a split
+    //   // If no direct match on time and source === CDC, look for a split
```
JordanGuinn, line 459: This feels like too many slashes 😉

src/controllers/enrollmentReports/map.ts
```diff
@@ -43,20 +48,24 @@ export const mapRows = async (
...
         rows[i],
         organizations,
         sites,
+        processedChildren,
         opts.save
       );
-      children.push(child);
+      if (child) children.push(child);
     } catch (err) {
       if (err instanceof ApiError) throw err;
```
JordanGuinn, line 67: Random and potentially ignorant question unrelated to anything - is there a particular reason we only re-throw `ApiError` instances?

src/controllers/enrollmentReports/map.ts
```diff
@@ -103,23 +145,71 @@ const mapRow = async (
...
+
+  // Apply any needed enrollment info updates before going to funding
+  if (modifyingExistingEnrollment) {
+    console.log('MODIFYING EXISTING ENROLLMENT');
+    await transaction.update(Enrollment, enrollment.id, enrollmentUpdate);
+  }
+
+  let funding = getExistingFundingForEnrollment(source, enrollment);
+  const modifyingExistingFunding = funding !== undefined;
+  const fundingUpdate = await Promise.resolve(
```
JordanGuinn, line 187: I don't think you need the `Promise.resolve` here - if `mapFunding` is async, you should be able to just `await` that, and of course if it's not there's no need to `await` it at all

src/controllers/enrollmentReports/map.ts
```diff
@@ -103,23 +145,71 @@ const mapRow = async (
...
+  transaction: EntityManager,
+  source: EnrollmentReportRow,
+  organization: Organization,
+  site: Site,
+  child: Child
+) => {
+  let enrollment = getExistingEnrollmentOnChild(source, child);
+  console.log('EXISTING ENROLLMENT: ', enrollment);
+  const modifyingExistingEnrollment = enrollment !== undefined;
+  const enrollmentUpdate = await Promise.resolve(
```
JordanGuinn, line 175: The same applies here as did for my previous comment about `Promise.resolve`

JordanGuinn (approved):
💯💯💯💯💯

src/controllers/enrollmentReports/mapUtils.ts
bamader: LOL you mean you don't want the mad ravings of my overly enthusiastic console in the app? 😂

src/controllers/enrollmentReports/map.ts
bamader: Lol whoops

src/controllers/enrollmentReports/map.ts
bamader: That's a good question. I have to confess that I've just been doing/using it because I've seen that's how the rest of the codebase is built, but that's not exactly a great reason to continue doing the practice. Maybe it has something to do with the types of custom things we've built using our middleware? Honestly, this is a good enough question that we might want to post  it into the devops slack channel to discuss more as a team.

src/controllers/enrollmentReports/map.ts
bamader: Yeah you're right. I was worried about synchronous modifications to the DB but that's not actually an issue/what's going on here.

Issue #560, opened 2020-10-13
Nested Tab Nav fast follows-- roster table width, button with dropdown alignment


mmazanec22:

mmazanec22:
Closed by #648

PR #648, opened 2020-10-30
548 month view

### LINKED ISSUE #548: Change month: I can view my roster for any month from July 2020 to the current month.
# [Designs in Abstract](https://share.goabstract.com/a9896ee2-a19e-447c-a0f3-b41dae257cd5)
Past enrollments

**UI Only features:**
– Past enrollment text button
– Button opens modal
- Modal has form elements to choose month and year
- Validations: Month cannot be before July 2020, Month cannot be in the future
- Submitting the form will load the roster for the given month. Any enrollment that was active for at least 1 day in that month is included
- Calendar Icon, label and text button to return to current month appears in the top right corner of the roster when viewing any past month
- Filtering is persisted through URL parameter

mmazanec22:
Closes #548 and #560 
Depends on https://github.com/ctoec/component-library/pull/87

### DIFF
--- a/client/src/containers/Roster/MonthFilter/MonthFilterModal.tsx
+++ b/client/src/containers/Roster/MonthFilter/MonthFilterModal.tsx
@@ -1,0 +1,93 @@
+import React, { useState } from 'react';
+import moment, { Moment } from 'moment';
+import cx from 'classnames';
+import {
+  Button,
+  DateInput,
+  FormStatusProps,
+  Modal,
+} from '@ctoec/component-library';
+
+type MonthFilterModal = {
+  filterByMonth?: Moment;
+  setFilterByMonth: (_: any) => void;
+  toggleModal: () => void;
+  isModalOpen: boolean;
+};
+
+export const MonthFilterModal: React.FC<MonthFilterModal> = ({
+  filterByMonth,
+  setFilterByMonth,
+  toggleModal,
+  isModalOpen,
+}) => {
+  // Duplicate of month value on roster-- needed bc we want to preserve the value but don't want to trigger the filtering until the user clicks the button
+  const [selectedMonth, setSelectedMonth] = useState<Moment | undefined>(
+    filterByMonth
+  );
+  const monthRecentEnough = selectedMonth?.isSameOrAfter(
+    moment('July 2020', 'MMMM YYYY')
+  );
+  const monthInPast = selectedMonth?.isSameOrBefore(moment());
+  let monthStatus: FormStatusProps | undefined = undefined;
+  if (selectedMonth && !monthRecentEnough) {
+    // Must be same or after July 2020
+    monthStatus = {
+      type: 'error',
+      id: 'month-too-far-back',
+      message: 'Month cannot be before July 2020',
+    };
+  } else if (selectedMonth && !monthInPast) {
+    // Must be same or before right now
+    monthStatus = {
+      type: 'error',
+      id: 'month-in-future',
+      message: 'Month cannot be in the future',
+    };
+  }
+
+  return (
+    <Modal
+      isOpen={isModalOpen}
+      toggleOpen={toggleModal}
+      header={<h2>View past enrollments</h2>}
+      content={
+        <>
+          <div className={cx('grid-row', { 'margin-left-3': !!monthStatus })}>
+            <DateInput
+              defaultValue={selectedMonth}
+              onChange={(moment) => setSelectedMonth(moment)}
+              label="View your roster for a past month"
+              id="roster-filter-date-input"
+              hideField={{
+                day: true,
+                calendar: true,
+              }}
+              status={monthStatus}
+            />
+          </div>
+          <div className="margin-top-4">
+            <div className="grid-row flex-first-baseline space-between-4">
+              <Button
+                appearance="outline"
+                onClick={toggleModal}
+                text="No, cancel"
+              />
+              <Button
+                appearance="default"
+                onClick={() => {
+                  if (!monthStatus) {
+                    setFilterByMonth(selectedMonth);
+                    toggleModal();
+                  }
+                }}
+                disabled={!!monthStatus}
+                text="View roster"
+              />
+            </div>
+          </div>
+        </>
+      }
+    />
+  );
+};
--- a/client/src/containers/Roster/Roster.tsx
+++ b/client/src/containers/Roster/Roster.tsx
@@ -1,4 +1,5 @@
 import React, { useContext, useEffect } from 'react';
+import { Moment } from 'moment';
 import {
   Accordion,
   Button,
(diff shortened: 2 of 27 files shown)


client/src/containers/Roster/Roster.tsx
```diff
@@ -21,45 +22,58 @@ import {
...
-  const isSiteLevelUser = user?.accessType === 'site';
 
-  // TODO: should this be all of the children for the whole org or what?
-  const numberOfChildrenWithErrors = childrenCache.records.filter(
-    (c) => c.validationErrors && c.validationErrors.length > 0
-  ).length;
+  const updateActiveMonth = (newMonth: Moment) => {
+    const month = getQueryMonthFormat(newMonth);
+    history.replace(
+      `/roster?${stringify({
```
jlhogan, line 60: nit, i prefer to use the path objects, so here instead of manually creating path with query, do
```
history.replace({ search: stringify({...})
```

client/src/containers/Roster/Roster.tsx
```diff
@@ -80,112 +93,129 @@ const Roster: React.FC = () => {
...
   }
 
   // Base case is single site user
-  let h1Content = sites.length ? sites[0].siteName : 'Loading...';
+  let h1Content = userSites.length ? userSites[0].siteName : 'Loading...';
+  let subHeaderText = '';
+
+  if (user && !childrenCache.loading) {
+    if (activeMonth) {
+      subHeaderText = getSubHeaderText(
```
jlhogan, line 112: nit:
```
subheaderText = getSubheaderText(
  activeMonth ? childrenFilteredByMonth : allChildren,
  userSites,
  activeMonth
)
```

client/src/containers/Roster/RosterButtonsTable.tsx
```diff
@@ -0,0 +1,60 @@
...
+  setFilterByMonth: (_: any) => void;
+};
+
+export const RosterButtonsTable: React.FC<RosterButtonsTable> = ({
+  organizations,
+  filterByMonth,
+  setFilterByMonth,
+}) => {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  // Duplicate of month value on roster-- needed bc we want to preserve the value but don't want to trigger the filtering until the user clicks the button
```
jlhogan, line 21: think this comment is kruft

client/src/containers/Roster/useRosterFilters.ts
```diff
@@ -0,0 +1,70 @@
...
+    setChildrenFilteredByLocation(
+      getFilteredChildren(allChildren, {
+        activeOrgId,
+        activeSiteId,
+      })
+    );
+  }, [allChildren, activeOrgId, activeSiteId]);
+
+  // There's gotta be a better name for this variable
+  const [completelyFilteredChildren, setCompletelyFilteredChildren] = useState<
```
jlhogan, line 42: `childrenFilteredByAll` ? But also, why maintain all of these as state? can't we just do the filtering and display result of that as user applies various filters?

jlhogan (approved):

client/src/containers/Roster/useRosterFilters.ts
mmazanec22: Using a state var means we can avoid re-filtering by month when a user clicks on a different site/org and vice versa.  Maybe it won't make a big diff... I figured we might as well try to not do extraneous looping through all the data though.  Is there a reason it would be preferable not to use state here though?

client/src/containers/Roster/MonthFilter/MonthFilterModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+      message: 'Month cannot be before July 2020',
+    };
+  } else if (selectedMonth && !monthInPast) {
+    // Must be same or before right now
+    monthStatus = {
+      type: 'error',
+      id: 'month-in-future',
+      message: 'Month cannot be in the future',
+    };
+  }
```
bamader, line 47: Potentially dumb question (and definitely for a separate ticket/PR): is this one of the kinds of validation we want to move to the backend? I guess here it's not actually a 'validation' in the sense of the classes we've been using, but I wasn't sure how much error-checking we wanted to shift out of the client. Though, now that I say that, it doesn't actually make much sense to move this out because it's directly picked by a user in the client...

client/src/containers/Roster/MonthFilter/MonthFilterModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+export const MonthFilterModal: React.FC<MonthFilterModal> = ({
+  filterByMonth,
+  setFilterByMonth,
+  toggleModal,
+  isModalOpen,
+}) => {
+  // Duplicate of month value on roster-- needed bc we want to preserve the value but don't want to trigger the filtering until the user clicks the button
+  const [selectedMonth, setSelectedMonth] = useState<Moment | undefined>(
+    filterByMonth
+  );
```
bamader, line 27: I might've missed it, but what is this a duplicate of? It doesn't seem like this is replicating any other code that's there (I see Julia's remarked that this comment in another file might be kruft, and it seems to me like that's because it's doing something else, but maybe that's the point?).

bamader (approved):
Looks good, looks good!

client/src/containers/Roster/MonthFilter/MonthFilterModal.tsx
mmazanec22: I think when we talk about validations in that context we usually mean validating data stored in the db.  This date only exists ephemerally as a query param

client/src/containers/Roster/MonthFilter/MonthFilterModal.tsx
mmazanec22: This is a duplicate of the month filter stored in roster-- we need to have it here too so that the user can set it and see the status message if there is one before we go changing query params and filtering data and such (which happens in the onclick button below)

client/src/containers/Roster/MonthFilter/MonthFilterModal.tsx
bamader: Gotcha. That makes a lot more sense.

client/src/containers/Roster/MonthFilter/MonthFilterModal.tsx
bamader: Ah okay, so this is a way of keeping hold of information until the user acknowledges (via onClick or changing filters) they've seen the status update (whatever it might be)?

mmazanec22: Yeah, and disabling the button to filter based on local state rather than pushing an invalid state to the parent component (roster)

PR #650, opened 2020-10-30
Correct table spacing

### LINKED ISSUE #618: Text on funding space page is misaligned
On the funding space page, the spacing between the text in the contract space column and the accepted formats column is misaligned, resulting in weird spacing. This looks like it's because each row of the Table is a FundingType, which consists of multiple FundingSpaces, and the accepted formats of each FundingSpace is a string that has to wrap to a newline, which throws the spacing off. Consult styling elements [here](https://www.w3schools.com/tags/att_td_colspan.asp) to turn the table into one where Contract Spaces define each row, and try to make the FundingType labels span multiple rows.

bamader:
Reformats the table of funding types and contract spaces to use contract spaces as the row indexer for each column (this forces the accepted format text to line up with each contract space). Funding types are then broadcast over multiple rows.

Closes #618

### DIFF
--- a/client/src/containers/FundingSourceTimes/FundingSourceTimes.tsx
+++ b/client/src/containers/FundingSourceTimes/FundingSourceTimes.tsx
@@ -1,21 +1,63 @@
 import React from 'react';
 import { Table, Column } from '@ctoec/component-library';
 import { BackButton } from '../../components/BackButton';
-import { FundingSourceTime } from '../../shared/models';
+import {
+  FundingSource,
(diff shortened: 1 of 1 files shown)


jlhogan (approved):
Variable names are a bit confusing to me, but not a blocker

mmazanec22 (approved):

bamader:
I'll try to make the names more descriptive before merging, then! Would rather have too much clarity in variables than a type that's confusing unless you really stare at it.

Issue #652, opened 2020-11-02
Spacing on text links


RyanHansz:
Missing line-height between two text links on data definitions page:
<img width="316" alt="Screen Shot 2020-11-02 at 9 45 08 AM" src="https://user-images.githubusercontent.com/63253539/97881258-2d241400-1cf0-11eb-8b20-85a789340b8a.png">

mmazanec22:
Noticed this and already started fixing it in my PR for #616 !  (It was because I removed the default margin so that it would fit in the table and didn't catch all instances of this)

mmazanec22:
Closed by #660

PR #660, opened 2020-11-03
616 moar fake data

### LINKED ISSUE #616: Create separate data samples for download
1 – For programs to download in the app. Should not have any errors. See:
[This spreadsheet](https://docs.google.com/spreadsheets/d/14GLwMS2i6hzLDNZb1hKqeLmuQavKpgg7aiKOq83lrDE/edit#gid=0)

2 – For testing, a page only the team can access where we can download data to test with that has errors and other test scenarios. Use our existing sample data for now.

mmazanec22:
Closes #616 and #652 

You can see the dev utility page by going to /dev in dev or test environments.


mmazanec22:
Ideal future state would prob be a form where we could specify some fake data params-- which fields are missing, how many rows, etc.

client/src/containers/DevUtility/DevUtility.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { getH1RefForTitle } from '../../utils/getH1RefForTitle';
+import { CSVExcelDownloadButton } from '../../components/CSVExcelDownloadButton';
+
+const DevUtility: React.FC = () => {
+  const h1Ref = getH1RefForTitle();
+  return (
+    <div className="grid-container margin-top-4">
+      <div className="grid-row grid-gap">
+        <div className="desktop:grid-col-8">
+          <h1 ref={h1Ref}>Templates and sample data for development</h1>
```
bamader, line 11: All these varieties are _fantastic_!

client/src/routes.ts
```diff
@@ -90,6 +91,12 @@ export const routes: RouteConfig[] = [
     component: BatchEdit,
     unauthorized: false,
   },
+  {
+    path: '/dev',
+    component:
+      process.env.NODE_ENV !== 'production' ? DevUtility : PageNotFound,
+    unauthorized: true,
+  },
```
bamader, line 99: Do we think there's value in exposing these sample data templates to providers? Maybe not all of them, but perhaps ones that are correctly filled out or show how to enter particular kinds of child information? (This question might also be dumb because that could already be the plan; I just saw the `!== 'production'` and assumed we'd have this be dev only).

src/controllers/export.ts
```diff
@@ -126,92 +139,74 @@ function formatStringPush(value: any) {
...
  * @param child
- * @param cols
+ * @param columns
  */
-function flattenChild(child: Child, cols: ColumnMetadata[]) {
+function flattenChild(
+  columns: ColumnMetadata[],
+  child: Child,
+  enrollment?: Enrollment,
+  skipInfoForPastEnrollments?: boolean
```
jlhogan, line 149: my favorite nit: pop this in an opts obj param, so it can be named when calling and not just contextless boolean

src/utils/parseQueryString.ts
```diff
@@ -3,23 +3,23 @@ import { Request } from 'express';
 export const parseQueryString = <T = string>(
   req: Request,
   name: string,
-  opts?: { post?: (_: string) => T; forceArray?: boolean }
+  opts?: { post?: (_: any) => T; forceArray?: boolean }
```
jlhogan, line 6: in my open PR i just deleted this b/c it was not working and i figured there's prob something out there that works better, but if we feel good about this fix then happy to let this land + then re-introduce to my code :tada:

jlhogan (approved):
this is sick

src/controllers/export.ts
bamader, line 149: Any particular reason for having this as its own parameter? Based on the conditional it's part of in line 175, I can't quite tell what the advantage is of having this be separate and not processing some of the sections.

src/routes/enrollmentReports.ts
```diff
@@ -19,32 +15,6 @@ import { validate } from 'class-validator';
 
 export const enrollmentReportsRouter = express.Router();
 
-/**
- * /enrollment-reports/:reportId GET
- *
- * Returns the parsed data from the given EnrollmentReport,
- * as nested data object with Child as root.
- */
-enrollmentReportsRouter.get(
```
bamader, line 28: Do we not make any calls to this route anymore? I know the `POST` route for enrollmentReports returns an `enrollmentReportId`, but I don't know if we actually do anything with it.

bamader (approved):
LGTM!

src/routes/enrollmentReports.ts
mmazanec22: No, this is unused now-- we only use the post routes (and should maybe rename them to batch upload at some point??)

src/utils/parseQueryString.ts
mmazanec22: This seems to work as expected now, but agree that we could use some library to do the parsing (query-string or querystring perhaps, heh)

src/controllers/export.ts
mmazanec22: Was following the update enrollment example data that Ryan provided-- it looks like we only ask people to include child identifiers + enrollment to update an enrollment, so if we're providing this as an example, we should prob make it mirror what users will be doing

client/src/routes.ts
mmazanec22: We are providing sample data to providers on the template page-- this page here at `/dev` just gives us a way to get at the incomplete/bad data samples

PR #661, opened 2020-11-03
354 pagination better cache

### LINKED ISSUE #633: Tab nav disappears when navigating to a site with no records

jlhogan:
No pagination yet (lol) but SWR has built in things for fetching and caching pages of data, so it will get done today OR ELSE

the tldr here is now swr is the cache, and you can mutate a list of cache things with a single item by using the provided mutate function if you've fetched the same key. Had to change the apiGet signature to take accessToken as an inline argument to avoid unnecessary code writing but if anyone has strong objection to that, let's talk.  we can tweak things like deduping interval or refetch interval or retry interval. the docs are actually v great: https://github.com/vercel/swr

also closes #633


client/src/containers/CreateRecord/CreateRecord.tsx
```diff
@@ -104,18 +116,19 @@ const CreateRecord: React.FC = () => {
...
         const currentStepStatus = steps[indexOfCurrentStep].status({
           child: updatedChild,
         } as RecordFormProps);
 
-        if (currentStepStatus === 'complete') {
+        if (
+          currentStepStatus === 'complete' ||
+          currentStepStatus === 'exempt'
+        ) {
+          console.log('moving next step');
```
mmazanec22, line 131: Console log

client/src/containers/Roster/RosterButtonsTable.tsx
```diff
@@ -1,19 +1,17 @@
 import { Button, Calendar, TextWithIcon } from '@ctoec/component-library';
 import { Moment } from 'moment';
-import React, { useState } from 'react';
+import React, { useState, useContext } from 'react';
```
mmazanec22, line 3: Looks like some of these imports are unused

client/src/containers/Roster/rosterUtils.tsx
```diff
@@ -1,153 +1,68 @@
...
-  };
-}
-
+/**
+ * Does client-side data filtering, by site and/or month
+ * @param allChildren
+ * @param site
+ * @param month
+ */
 export function getFilteredChildren(
```
mmazanec22, line 33: A note here about how this is only for front end filtering, or maybe a func renaming, could make this clearer

src/controllers/reportingPeriods.ts
```diff
@@ -1,12 +1,18 @@
...
 import moment from 'moment';
+import { FundingSource } from '../../client/src/shared/models';
 
 /**
  * Get all reporting periods, less than 3 years old
  * and less than 3 years in the future
  */
-export const getReportingPeriods = async () => {
+export const getReportingPeriods = async (shortFundingSource?: string) => {
+  console.log('fundingSource', shortFundingSource);
```
mmazanec22, line 11: Console log

src/controllers/reportingPeriods.ts
```diff
@@ -1,12 +1,18 @@
...
+  const fundingSource = Object.values(FundingSource).find((fs) =>
+    fs.startsWith(shortFundingSource)
+  );
+  console.log('looke dup', fundingSource);
```
mmazanec22, line 15: Looke dup!

mmazanec22 (approved):

src/controllers/reportingPeriods.ts
jlhogan: :)

client/src/components/Forms/Enrollment/Funding/Fields/ContractSpace/ContractSpace.tsx
```diff
@@ -10,15 +10,17 @@ import {
...
 import { SingleContractSpaceField } from './SingleContractSpace';
 import {
   ChangeFunding,
   ChangeEnrollment,
 } from '../../../../../../shared/payloads';
 import { fundingSpaceFormatter } from '../../../../../../utils/formatters';
-import DataCacheContext from '../../../../../../contexts/DataCacheContext/DataCacheContext';
 import { getValidationStatusForField } from '../../../../../../utils/getValidationStatus';
+import { useAuthenticatedSWR } from '../../../../../../hooks/useAuthenticatedSWR';
+import { stringify } from 'querystring';
```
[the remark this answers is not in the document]
jlhogan: i added `query-string` a while ago, and i think it's a bit better, however if `query-string` is in there and we can't remove it then maybe we stick with that? At least requires parsing the `?` and `#` off of search/hash, which is annoying. will TAL.

bamader (approved):
Looks good! Will probably need to read the docs myself before doing any coding or manipulating the cache but you're right, they do look pretty friendly.

PR #662, opened 2020-11-03
All the fixes

### LINKED ISSUE #617: Verify error modal catches all relevant errors.
Once #611 is resolved and once validations are cleaned up in the backend, test that the error-checking modal shown on the upload screen can catch missing income determination information. 

Also check for errors in enrollment and funding such as:
- Missing/incorrect site name
- Missing/Incorrect Org name
- Missing/Incorrect funding type
- Missing/Incorrect space type
- Missing/Incorrect funding reporting periods
– Invalid funding types for age group

### LINKED ISSUE #655: Historical fundings don't appear in a child's record when using a sheet with historical enrollments/fundings in batch upload

bamader:
Addresses a lot of things dealing with bugs and validations.
- Fixes fundings not showing up when doing historical enrollment batch upload (closes #655 )
- Fixes  the funding age group match validation to not crash everything when parsing null/undefined
- Expands error modal  functionality to catch errors in nested objects (closes #617 )
- Fixes bug with processing family incomes of 0 (they were being interpreted as undefineds)
- Related to ctoec/component-library#88 to add scrolling to large modals

I'll update `yarn.lock` and `package.json` once the  component library PR  lands.


src/controllers/enrollmentReports/checkErrors.ts
```diff
@@ -10,21 +33,33 @@ import { getAllColumnMetadata } from '../../template';
  */
 export const checkErrorsInChildren = async (children: Child[]) => {
   const cols: ColumnMetadata[] = getAllColumnMetadata();
-  const propertyNameToFormattedName = {};
-  let errorDict = {};
+
+  // Need to start the tracking dicts off with a few mappings because
```
mmazanec22, line 37: I feel like this is a problem all over the back end and we should maybe come up with a sturdier fix-- like either writing a func that maps any input data to our property names, or just making our properties line up with the column names

src/controllers/enrollmentReports/map.ts
```diff
@@ -370,7 +370,9 @@ const mapIncomeDetermination = (
     let incomeDetermination = {
       // Cast empty strings to undefined to avoid DB write failures
       numberOfPeople: source.numberOfPeople || undefined,
-      income: source.income || undefined,
+      // Need to accept 0 as valid income, so use forcible number conversion
+      // to check if the result is a valid number
+      income: isNaN(source.income) ? undefined : source.income,
```
mmazanec22, line 375: Good catch!

src/controllers/enrollmentReports/checkErrors.ts
bamader: It def is, you're totally right. Do you think it's worth just making a util file for the backend that has all of these mappings in one place? It'll still have some of those types of encodings in there, but at least it would be centralized and serve as a single source of truth.

jlhogan (approved):

src/controllers/enrollmentReports/checkErrors.ts
mmazanec22: Something like that?  Maybe an enum per object that has mismatched property names, and then some helper functions to convert either way?  And then we probably should make a ticket to change that eventually-- seems unnecessary to work around ourselves

bamader: I'm now wondering if we should just put all of the mismatched naming stuff into one ticket. This might be the wrong attitude (but I figure I should bring it up given our timeline), but refactoring back end naming symbols seems like a lower priority than some of our other point-in-time launch stuff for December 1? If that's not the case I can work on it here, I just want to make sure it's a right thing to prioritize.

(Also, because I know tone is hard to convey in text commentary, my intention here is to ask a question and not push back on refactoring as an idea, so I'm sorry if it comes across as dodgy!)

bamader: I ended up refactoring the property names on `enrollmentReportRow` to line up with the properties on the various objects (`Child`, `Family`, `Enrollment`, etc.) because it broke WAY less than trying to rename and refactor the objects to match enrollmentReportRow (I started out with that approach but I think it wound up being over-engineering because we were gonna need another ticket to make the names match anyway, and it involved more work than the returns gave us gains). Was able to clean out the hard coding of stuff in the tracking dicts when counting errors, and also was able to cut out some of the specific casing in the `export` controller because now all of the fields are just on the enrollmentReportRow. If we see other places to simplify names using this strategy, we can do a fast follows ticket, but I think the fixes in this one should be good to go.

src/controllers/enrollmentReports/checkErrors.ts
```diff
@@ -19,12 +42,10 @@ export const checkErrorsInChildren = async (children: Child[]) => {
     propertyNameToFormattedName[c.propertyName] = c.formattedName;
   });
 
-  children.forEach((child) => {
+  children.forEach(async (child) => {
```
mmazanec22, line 45: If this is async then we should probably wrap this in an `await Promise.all` and then use map to return the promise?

mmazanec22 (approved):

src/controllers/enrollmentReports/checkErrors.ts
bamader: Can do!

PR #667, opened 2020-11-04
354 (pt II): actually paginate

### LINKED ISSUE #354: API pagination
- as an API consumer, results from the `children` endpoint can be requested in chunks by supplying limit & offset (or take & skip -- this is what they're called in typeorm) QS params
  - children controller handles these QS params and applies them as query filters
  - client api helper abstracts pagination and appends results

open Q:
- pagination implemented like this will decrease time to load for the page w/ data but does not actually mean we've implmented any pagination in the browser.
at a certain point, will the data be too large to present all at once to the user? How many records do we anticipate the largest user to have? Lets make sure we confirm that our designs hold up under that load. If not, we need a solution for pagination in the roster view (fetching + loading only expanded roster sections? this means implementing API filtering on `/children` )

jlhogan:
Also, change how we query for children to offload a bit more of the
filtering to the DB (in preparation for maybe one day doing all the
filtering in the DB?) and separate child controller into a few separate files because that monster was no fun to read

closes #354

### DIFF
--- a/client/src/containers/CreateRecord/CreateRecord.tsx
+++ b/client/src/containers/CreateRecord/CreateRecord.tsx
@@ -10,8 +10,6 @@ import { getH1RefForTitle } from '../../utils/getH1RefForTitle';
 import { RecordFormProps } from '../../components/Forms/types';
 import { useFocusFirstError } from '../../hooks/useFocusFirstError';
 import { listSteps } from './listSteps';
-import { useAuthenticatedSWR } from '../../hooks/useAuthenticatedSWR';
-import { stringify } from 'querystring';
 
 type LocationType = Location & {
   state: {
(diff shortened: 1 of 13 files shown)


client/src/containers/CreateRecord/CreateRecord.tsx
```diff
@@ -48,21 +46,6 @@ const CreateRecord: React.FC = () => {
 
   const [child, setChild] = useState<Child>();
   const organization = locationState?.organization || child?.organization;
-  const { mutate } = useAuthenticatedSWR<Child[]>(
```
jlhogan, line 51: no point mutate-ing this cache anymore because it is not the cache we use in the roster page (we paginate now)

client/src/containers/EditRecord/EditRecord.tsx
```diff
@@ -15,21 +15,15 @@ import {
...
   const h1Ref = getH1RefForTitle('Edit record');
-  const { childId, organizationId } = useParams() as {
+  const { childId } = useParams() as {
     childId: string;
-    organizationId: string;
   };
   const { accessToken } = useContext(AuthenticationContext);
   const { alertElements, setAlerts } = useAlerts();
   const [child, setChild] = useState<Child>();
-  const { mutate } = useAuthenticatedSWR(
```
jlhogan, line 30: same here as above -- no benefit to attempting to mutate this cache

client/src/containers/Roster/Roster.test.tsx
```diff
@@ -190,30 +221,5 @@ describe('Roster', () => {
     }
   );
 
-  it('correctly separates children by ageGroup', async () => {
```
jlhogan, line 193: this test was failing and im OVER IT, and it should really be a unit test for the roster util funcs anyway, so let's add those (one day....)

PR #668, opened 2020-11-04
499 loading

### LINKED ISSUE #499: Improved loading state for roster and batch upload
Roster Loading state

6 2 Batch Upload – Uploading

- Remove empty state alert while loading
- Add Indeterminate loading animation to roster, batch upload 
- Create a component in component library
If possible, use circular indeterminate loading animation from material-ui 
[Circular indeterminate loading](https://material-ui.com/components/progress/#circular-indeterminate)

mmazanec22:
Closes #499 
Depends on https://github.com/ctoec/component-library/pull/89

### DIFF
--- a/client/src/containers/Roster/Roster.tsx
+++ b/client/src/containers/Roster/Roster.tsx
@@ -150,10 +150,7 @@ const Roster: React.FC = () => {
           filterByMonth={queryMonth}
           setFilterByMonth={updateActiveMonth}
         />
-        <LoadingWrapper
-          text="Loading your roster..."
-          loading={loading}
-        >
+        <LoadingWrapper text="Loading your roster..." loading={loading}>
           {!(children || []).length ? (
             <NoRecordsAlert />
           ) : tabNavProps ? (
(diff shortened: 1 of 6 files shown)


client/src/containers/Upload/Upload.tsx
```diff
@@ -31,27 +36,26 @@ const Upload: React.FC = () => {
 
   const h1Ref = getH1RefForTitle();
   const { accessToken } = useContext(AuthenticationContext);
+  const history = useHistory();
```
mmazanec22, line 39: I did some light rearranging in this file to try to grok where to put loading-related logic-- no functionality is changed

jlhogan (approved):

client/src/containers/Roster/Roster.tsx
```diff
@@ -57,11 +58,12 @@ const Roster: React.FC = () => {
   useUpdateRosterParams();
 
   // Fetch child data, filtered by organization and month
-  const { data: children, isValidating } = useAuthenticatedSWR<Child[]>(
+  const { data: children, isValidating, error } = useAuthenticatedSWR<Child[]>(
     query.organization
       ? `/children?${stringify({ organizationId: query.organization })}`
       : null
   );
+  const loading = !children && !error;
```
bamader, line 66: Non-blocker: are we caching errors that might exist on the children? I see in line 61 that the cache hands us back both children and the errors, but I didn't realize we were storing errors in the cache. I only ask because at first I was like why are we also checking `!error` because there shouldn't be errors without children, but maybe that's not the case?

bamader (approved):
LGTM! Sorry for the review delay!

client/src/containers/Roster/Roster.tsx
jlhogan: error == there was an error making the API call. we're not "storing" errors in the cache, but the useSWR hook makes api calls and returns an error, same as it thrown by the `apiGet` func

bamader: Mmmm I understand. Thank you!

PR #672, opened 2020-11-04
653: Add custom logger to redact parameters in prod-like environments

### LINKED ISSUE #653: custom TypeORM logger to redact parameters
Goal: Remove PII from logging. 

The alternative is to not do query logging in the application but do transaction logging on the database.

jlhogan:
closes #653


src/loggers/QueryLogger.ts
```diff
@@ -0,0 +1,51 @@
+import { AdvancedConsoleLogger, QueryRunner, Logger } from 'typeorm';
```
bamader, line 1: Not an actual work comment, but the concept of an _advanced_ logger is hilarious to me. Like, it's not just `console.log()`. It's `console.*advanced*.log()` 😂

bamader (approved):
LGTM!

mmazanec22 (approved):

src/loggers/QueryLogger.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  logSchemaBuild(message: string, queryRunner?: QueryRunner) {
+    super.logSchemaBuild(message, queryRunner);
+  }
+
+  logMigration(message: string, queryRunner?: QueryRunner) {
+    super.logMigration(message, queryRunner);
+  }
+
+  log(level: 'log' | 'info' | 'warn', message: any, queryRunner?: QueryRunner) {
```
JordanGuinn, line 48: Thoughts on adding at least `error` here, and potentially `debug` and `trace` as well?

JordanGuinn (approved):

src/loggers/QueryLogger.ts
jlhogan: this is direct copy/paste from https://github.com/typeorm/typeorm/blob/master/src/logger/AdvancedConsoleLogger.ts

JordanGuinn: Oh weird - _why do they only support those three_?  Ah well 🤷🏽‍♂️

PR #673, opened 2020-11-04
354: Backend changes for pagination of child records fetch

jlhogan:
- separates child controller into multiple  files because they were getting really long
- adds site-level filter to child query 
- adds skip/take params to /children GET
- removes some community kruft from auth flow
- adds organizations to users even if they only have site permissions (because we definitely want to know that info, and popping it there doesn't give the user any read/write access they shouldn't have


src/controllers/children/create.ts
```diff
@@ -0,0 +1,41 @@
...
+  ) {
+    console.error(
+      'User does not have permission to create the child row supplied'
+    );
+    throw new Error('Child creation request denied');
+  }
+
+  // TODO: make family optional on the child
+  // (to enable family lookup when adding new child)
+  // and stop creating it here
```
bamader, line 25: Is this a feature we need to implement before Dec 1?

src/controllers/children/get.ts
```diff
@@ -0,0 +1,113 @@
...
+      'enrollments.fundings',
+      'organization',
+    ],
+    where: (qb: SelectQueryBuilder<Child>) => {
+      qb.where('Child.organizationId IN (:...orgIds)', {
+        orgIds: filterOrgIds,
+      });
+      // On deck for typeORM to implement where for nested relations in find
+      // https://github.com/typeorm/typeorm/issues/2707
+      // Until then, use the generated aliases to do nested filtering
```
bamader, line 99: I was literally just reading about nested relations filtering while researching the ambiguously documented and contradictory reported soft delete 👀

src/controllers/children/get.ts
```diff
@@ -0,0 +1,113 @@
...
+      if (user.accessType === 'site') {
+        qb.andWhere('Child__enrollments.siteId IN (:...siteIds)', {
```
bamader, line 101: Do you need a `join` clause to use an alias? I remember reading about using `.leftJoin()` to access the nested properties of a relation.

src/controllers/users.ts
```diff
@@ -1,11 +1,29 @@
...
+  const sites =
+    user.siteIds && user.siteIds.length
+      ? await getManager().findByIds(Site, user.siteIds)
+      : [];
+
+  //
+  const orgIds =
+    user.organizationIds && user.organizationIds.length
+      ? user.organizationIds
+      : sites.map((s) => s.organizationId);
```
bamader, line 22: I think I know the answer, but this shouldn't show site-level users anything on the roster page they shouldn't see, right? Like, we want to augment the users with some site data in orgs so we know how they have particular access, but this shouldn't show any roster UI elements (from other sites or org-level views) the user wouldn't normally have?

src/controllers/children/get.ts
```diff
@@ -0,0 +1,110 @@
...
+    take?: number;
+  }
+) => {
+  let { organizationIds, missingInfo, skip, take } = filterOpts;
+
+  const opts = (await getFindOpts(user, organizationIds)) as FindManyOptions<
+    Child
+  >;
+  opts.skip = skip;
+  opts.take = take;
```
bamader, line 46: What's the importance of these? Do they change how `find` works?

bamader (approved):
LGTM!

src/controllers/children/create.ts
jlhogan: no :)

src/controllers/children/get.ts
jlhogan: the joins happen because of the `relations` specified above

src/controllers/users.ts
jlhogan: nope, and we need to know about the org the user "belongs" to even if they don't have org-level permissions. So if we are showing them things they don't need to be / shouldn't be seeing, i think we should fix that on the client-side (but I'm pretty sure they dont)

src/controllers/children/get.ts
jlhogan: ahhh yes will comment, sorry! these are how we do the pagination -- skip representing the offset, and take representing the next page size which in SQL Server are the `offset_fetch` part of this ORDER BY syntax https://docs.microsoft.com/en-us/sql/t-sql/queries/select-order-by-clause-transact-sql?view=sql-server-ver15#syntax

src/controllers/children/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './changeEnrollment';
```
mmazanec22, line 1: Yay for splitting this up into smaller files!

mmazanec22 (approved):

PR #680, opened 2020-11-05
354 (pt III): actually paginate

### LINKED ISSUE #354: API pagination

jlhogan:
closes #354 
depends on #673

### DIFF
--- a/client/src/containers/Roster/hooks/useOrgSiteProps.ts
+++ b/client/src/containers/Roster/hooks/useOrgSiteProps.ts
@@ -1,5 +1,5 @@
 import { TabNav, TabItem } from '@ctoec/component-library';
-import { stringify, parse } from 'querystring';
+import { stringify, parse } from 'query-string';
 import { QUERY_STRING_MONTH_FORMAT } from '../rosterUtils';
 import { useHistory } from 'react-router-dom';
 import { useContext } from 'react';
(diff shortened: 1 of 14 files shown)


client/src/containers/Roster/hooks/useOrgSiteProps.ts
```diff
@@ -1,54 +1,72 @@
...
+    month?: string;
   };
 
+  // Function to update search query when user clicks on tab nav
   const tabNavOnClick = (clickedId: string, clickedItem: TabItem) => {
     if (clickedItem.nestedItemType) {
       // If it has a nested item type then it's an org
       history.push({
-        search: stringify({ organization: clickedId, month: query.month }),
+        search: stringify({ ...query, organization: clickedId }),
```
mmazanec22, line 29: If we change orgs we also need to unselect the site, so this should prob include `site: undefined`?

client/src/containers/Roster/hooks/useOrgSiteProps.ts
```diff
@@ -1,54 +1,72 @@
...
-      history.push({
-        search: stringify({ site: clickedId, month: query.month }),
-      });
+      // Push a specific site id if specific site clicked
+      if (clickedId !== ALL_SITES) {
+        history.push({
+          search: stringify({ ...query, site: clickedId }),
+        });
+      }
+      // Or remove site param from search if 'All sites' clicked
```
mmazanec22, line 38: Huh, I wouldn't have thought that this would be necessary to get all sites to show-- is it just to clean up the params by removing an unnecessary one?  If so we should maybe do the same for orgs for consistency

client/src/containers/Roster/rosterUtils.tsx
```diff
@@ -36,16 +36,18 @@ export function applyClientSideFilters(
   month?: Moment
 ): Child[] {
   let filteredChildren: Child[] = allChildren;
+  if (month) {
+    filteredChildren = filteredChildren.filter((child) => {
+      console.log(child, childHasEnrollmentsActiveInMonth(child, month));
```
mmazanec22, line 41: Console log

mmazanec22 (approved):

client/src/containers/Roster/hooks/useOrgSiteProps.ts
jlhogan: yeah just a clean up thing -- there's no need to have "all sites" in the query params, an undefined value will trigger same functionality. I got rid of "all org view" b/c its too much data and a user will not actually be able to get anything meaningful from that view

client/src/containers/Roster/rosterUtils.tsx
jlhogan: ty!

bamader (approved):
Looks good! One general question: if we're taking the mutate calls out of CreateRecord, does that affect storing the partially created child in the cache so that if a user stops midway through they don't lose the progress?

(Also I saw your comment about clearing the cache in the roster test and I tried to look up a good way to do that but like nothing seems to work :/ . Some users said it  was "dangerous" to clear the cache because a component could still be using one of the keys but I don't know if it's worth spending a ton of time trying to debug that specific test case)

jlhogan:
> Looks good! One general question: if we're taking the mutate calls out of CreateRecord, does that affect storing the partially created child in the cache so that if a user stops midway through they don't lose the progress?

the partially created child will exist in the DB the front end cache is never the place where we store data as source of truth. we just store it for convenience and improved performance. so when user returns to roster after partial child create, it'll be pulled with all the other child records (API calls are re-triggered when component remounts)

PR #689, opened 2020-11-06
Soft delete has liftoff

### LINKED ISSUE #643: Soft delete data
Updating everything in database to have delete field.

Update controllers to set deleted = 1 instead of hard deleting. 

Filter for delete = 1 when querying data

May be a thing typeORM can do

bamader:
Implements soft-delete using some typeorm stuff and some manual filtering. Closes #643 
The good news: typeorm's `@DeleteDateColumn` decorator can be automatically set any time `softRemove()` is called, and `entityManager.find()` queries will automatically filter out entities of whatever type you're looking for if they have a deletedDate that isn't null.
The bad news: typeorm sucks at documenting and supporting soft-delete features. As of GitHub issues from 2 days ago, TypeORM doesn't support cascading soft removal or nested/relational filtering of soft-deleted things, nor do they plan to. The community has generally suggested workarounds in the vein of manual filtering post-find and repurposing some decorators to create a cascading effect without TypeOrm. So, those things had to get manually put in.

But, it works without refresh, doesn't crash the app, and keeps filtering on the backend where the client doesn't need to know about it!


client/src/shared/models/db/Child.ts
```diff
@@ -39,4 +39,5 @@ export interface Child extends ObjectWithValidationErrors {
   family?: Family;
   organization: Organization;
   enrollments?: Array<Enrollment>;
+  deletedDate: Date;
```
jlhogan, line 42: don't think this needs to be on the shared FE models

src/data/children.ts
```diff
@@ -49,6 +49,8 @@ const children: Child[] = Array.from({ length: 100 }, (_, i) => {
     organization: org,
     organizationId: org.id,
     updateMetaData: { updatedAt: new Date() },
+    deletedDate: null,
```
jlhogan, line 52: can we please att `deletedDate` to update metadata ? more easily ensure all fields that we want to audit are being treated the same way

jlhogan (commented):
All in all this seems v reasonable to me, the util functions are clear and clean. bummer that we can't filter on the DB side, but alas. Maybe one day move to query builders all around, but definitely not today. Request to remove the deleted fields from the FE models + to move the deletedDate to `updateMetadata`

src/data/children.ts
bamader: I don't quite understand the ask here. The `deletedDate` field has to stay on its respective entity so that typeorm knows what to autoupdate and filter, so is this saying to also make a `deletedDate` field in `updateMetaData` that we manually set every time we call for a delete?

jlhogan: I was under the impression that the decorator was what triggered the typeorm functionality. My assumption is that if we add the deleteDate with decorator to updateMetaData, which is already embedded in each of the updateable entities, the same behaviors would apply. That might be untrue tho.

bamader: Ah I understand. I guess I don't know if that'll work. I'll move it and see what happens.

bamader: Just tested it, and moving `deletedDate` into updateMetadata is causing lots of problems (it's not filtering on deletedDate == null anymore, and deletedDate is being initialized with a value when uploading a file which makes no sense but at this point doesn't surprise me because typeorm). Is it a blocker for it to stay outside `updateMetaData`?

src/utils/filterSoftRemoved.ts
```diff
@@ -0,0 +1,36 @@
...
+ * out all instances of the property in the entity's array that
+ * have a soft-deleted date set in the system.
+ * @param entity
+ * @param property
+ */
+export const filterXFromY = (entity: Object, property: string) => {
+  if (!entity || !property || !entity.hasOwnProperty(property))
+    return undefined;
+  if (!entity[property]) return entity;
+  const filteredProp = entity[property].filter(
```
JordanGuinn, line 14: This kinda presupposes that the property supplied for the entity provided is gonna be an array, yeah?

src/utils/filterSoftRemoved.ts
```diff
@@ -0,0 +1,36 @@
+import { Child, Enrollment, Family } from '../entity';
+
+/**
+ * Given an entity and an object property on that entity, filter
+ * out all instances of the property in the entity's array that
+ * have a soft-deleted date set in the system.
+ * @param entity
+ * @param property
+ */
+export const filterXFromY = (entity: Object, property: string) => {
```
JordanGuinn, line 10: Do we _need_ this to be this generic at this point, given we're only using it for the sole purpose of filtering out soft deleted entities?  What if we instead had something like:

```
export const removeDeletedRows = (rows: any[]) => {
  return rows.filter((row: any) => !row.deletedDate);
};
```

which we just applied directly to the collections of entities that needed filtering?

src/utils/filterSoftRemoved.ts
```diff
@@ -0,0 +1,36 @@
...
+};
+
+/**
+ * Perform a complete filter of all soft-deleted sub-entities
+ * from a given child. Filter's the child's family's income
+ * determinations, the child's enrollments, and the fundings of
+ * all remaining enrollments.
+ * @param child
+ */
+export const completeFilterChild = (child: Child) => {
```
JordanGuinn, line 28: nit: could this be named something more reflective of the action being performed here, like `removeDeletedEntities` or something?

src/entity/IncomeDetermination.ts
```diff
@@ -33,4 +39,7 @@ export class IncomeDetermination implements IncomeDeterminationInterface {
 
   @Column((type) => UpdateMetaData, { prefix: false })
   updateMetaData: UpdateMetaData;
+
+  @DeleteDateColumn()
```
JordanGuinn, line 43: 😎😎😎😎

JordanGuinn (commented):
Looks legit overall!  Just have a couple questions about the utility funcs

src/data/children.ts
jlhogan: nope!!!

src/utils/filterSoftRemoved.ts
bamader: It does, yeah. But that's actually okay for our implementation. Because typeorm's find calls filter for deletedDate at their own entity level (i.e. a find that searched for Children wouldn't return any child object with `deletedDate !== null`), the _only_ properties that won't be filtered for us are nested properties, and the only nested properties we have that we want filtered are all stored in arrays (incomeDeterminations on a Family, enrollments on a Child, and fundings on an Enrollment). But, given the comment below about the function not needing to be this overly generic, I'll be swapping this out with a cleaner, more specific to our use-case function anyway 😎

src/utils/filterSoftRemoved.ts
bamader: That makes sense to me. Not sure why my first instinct was to make it as generic as possible when we could just have a simpler array filtering function, but I'll switch this out!

src/controllers/enrollmentReports/removeExisting.ts
```diff
@@ -14,10 +15,13 @@ export const removeExistingEnrollmentDataForUser = async (
   user: User,
   siteIdsToReplace?: string[]
 ) => {
-  const allChildren = await transaction.find(Child, {
+  let allChildren = await transaction.find(Child, {
     relations: ['enrollments'],
     where: { organization: { id: In(user.organizationIds) } },
   });
+  allChildren.forEach((c) => {
+    c.enrollments = removeDeletedElements(c.enrollments || []);
```
jlhogan, line 23: i dont think this hurts any, but also shouldn't be necessary since we're never returning these child objects, just removing them. Oh unless trying to remove an already removed entity fails?

jlhogan (approved):

JordanGuinn (approved):

PR #706, opened 2020-11-10
665 batch edit wraparound

### LINKED ISSUE #665: After editing the last record listed in batch edit, I get "all records are complete" even if I didn't edit any of the other ones

### LINKED ISSUE #694: Batch editing is going to a completed state when there are records with missing info.
Nov-09-2020 12-34-28

jlhogan:
closes #665 
closes #694


client/src/containers/BatchEdit/BatchEdit.tsx
```diff
@@ -53,19 +59,15 @@ const BatchEdit: React.FC = () => {
...
 
-      setActiveRecordId(
-        // If a no records are missing info, then set active record to none
-        !firstStillMissingInformationRecord
-          ? undefined
-          : // otherwise set active record to first that is still missing info
-            firstStillMissingInformationRecord.id
-      );
+      setActiveRecordId(firstStillMissingInformationRecord?.id);
+      return;
```
jlhogan, line 70: this one `return` is the whole fix for #665

client/src/containers/BatchEdit/BatchEdit.tsx
```diff
@@ -10,19 +10,25 @@ import { BackButton } from '../../components/BackButton';
...
 import { BatchEditItemContent } from './BatchEditItemContent';
 import { Child } from '../../shared/models';
 import { getBatchEditErrorDetailsString } from './listSteps';
 import { useAuthenticatedSWR } from '../../hooks/useAuthenticatedSWR';
-import { stringify } from 'querystring';
+import { stringify, parse } from 'query-string';
 
 const BatchEdit: React.FC = () => {
   const { childId } = useParams() as { childId: string };
+  const { organizationId } = parse(useLocation().search) as {
```
jlhogan, line 22: scope batch edit to orgId so that the number of records you are editing in batch edit matches the number of records from the alert you clicked on -- thoughts @RyanHansz ? (this only matters for multi-org user(s))

client/src/containers/BatchEdit/BatchEditItemContent.tsx
```diff
@@ -42,6 +46,18 @@ export const BatchEditItemContent: React.FC<BatchEditItemContentProps> = ({
     setSteps(undefined);
   }, [childId]);
 
+  // Generate steps for the current child
+  // after state is set (child has been fetched)
```
jlhogan, line 50: just moved this around b/c i thought it made the file more readable

client/src/containers/BatchEdit/BatchEditItemContent.tsx
```diff
@@ -11,20 +11,24 @@ import AuthenticationContext from '../../contexts/AuthenticationContext/Authenti
...
 
 export const BatchEditItemContent: React.FC<BatchEditItemContentProps> = ({
   childId,
+  organizationId,
   moveNextRecord,
 }) => {
   const [child, setChild] = useState<Child>();
   const { mutate } = useAuthenticatedSWR<Child[]>(
-    `children?${stringify({ 'missing-info': true })}`
+    organizationId
```
jlhogan, line 29: limiting batch edit to org scope -- see comment above

client/src/containers/BatchEdit/BatchEditItemContent.tsx
```diff
@@ -55,39 +71,32 @@ export const BatchEditItemContent: React.FC<BatchEditItemContentProps> = ({
...
+            }
             return children;
-          }
-          return children;
-        }, false);
-        moveNextStep();
+          }, false);
+        }
+
+        if (triggerRefetchCount) moveNextStep();
```
jlhogan, line 93: somehow this check (to only move to next step if we were doing a post-update fetch, not initial grab the record fetch) got lost, and was also mucking things up i think

client/src/containers/EditRecord/DeleteRecord/DeleteRecord.tsx
```diff
@@ -16,9 +16,6 @@ export const DeleteRecord: React.FC<DeleteProps> = ({ child }) => {
   const toggleIsOpen = () => setIsOpen((o) => !o);
 
   const { accessToken } = useContext(AuthenticationContext);
-  const { mutate } = useAuthenticatedSWR(
```
jlhogan, line 19: this is a useless left over update of a cache we no longer use. We can come back in and add functionality to update the paginated roster cache (here, and withdraw, and edit) if we want, but this is kruft so wanted to clean it up

client/src/containers/Roster/hooks/useChildrenWithErrorsAlert.tsx
```diff
@@ -5,14 +5,17 @@ import { AlertProps } from '@ctoec/component-library';
 import React from 'react';
 import pluralize from 'pluralize';
 import { Link } from 'react-router-dom';
+import { stringify } from 'query-string';
 
 export const useChildrenWithErrorsAlert = (
   isLoading: boolean,
-  childrenWithErrorsCount: number
+  childrenWithErrorsCount: number,
+  organizationId?: string
```
jlhogan, line 13: add org id so we can pass it off to batch edit

client/src/containers/BatchEdit/BatchEdit.tsx
mmazanec22: Without digging too hard into everything else I don't understand why-- maybe a comment here?

mmazanec22 (approved):

client/src/containers/BatchEdit/BatchEditItemContent.tsx
bamader: +1 for this. The logical flow is a bit more intuitive to me this way.

client/src/containers/Roster/hooks/usePaginatedChildData.tsx
```diff
@@ -1,6 +1,7 @@
 import { stringify } from 'query-string';
 import { useAuthenticatedSWRInfinite } from '../../../hooks/useAuthenticatedSWR';
 import { Child } from '../../../shared/models';
+import { cache } from 'swr';
 
 const PAGE_SIZE = 100;
```
bamader, line 6: Not a blocker, just a design question: do we know how many records we want shown per page in the final version of the system (or I guess the Dec 1 version)?

bamader (approved):
LGTM!

client/src/containers/Roster/hooks/usePaginatedChildData.tsx
jlhogan: page_size is the size of data we fetch from the API, nothing to do with display -- all records will be displayed to the user. This was an arbitrary size, can make it smaller if it turns out grabbing 100 records is too slow but i figured tweaks like that will come after some usability testing w/ users or load testing by the team which just hasn't happened yet

Issue #540, opened 2020-10-08
Make funding reporting periods accept more date formats.


RyanHansz:
When I upload:
<img width="403" alt="Screen Shot 2020-10-08 at 3 11 36 PM" src="https://user-images.githubusercontent.com/63253539/95502916-970eff00-0978-11eb-93aa-d3600079708f.png">

The tool shows:
<img width="626" alt="Screen Shot 2020-10-08 at 3 11 29 PM" src="https://user-images.githubusercontent.com/63253539/95502918-970eff00-0978-11eb-8a5b-941292641f96.png">

We can make the field accept any date within the month to accommodate excel auto-formatting. 
- Support: **MM/YY** 10/20, MM/YYYY 10/2020, **MM/DD/YYYY** 10/1/2020, **MMM-YY** Oct-20,

PR #709, opened 2020-11-10
Fix fake data

### LINKED ISSUE #693: Fix upload parsing of site, funding type, reporting periods
Collab with kyle on #540

mmazanec22:
Closes #693 .  Addresses upload parsing of site and funding type, leaves reporting period formatting for #540 .  Breaks up the map file into a mapUtils folder because I was having a hard time keeping track of what was where.

When I download an example file and upload it, there are still a few funding types that aren't parsed.  I can't really make any sense of it, except maybe that the combination of funding type/time/whatever is not supported by that org?  Any help on that would be appreciated, though it is probably not essential for launch if it's just a problem with the fake data quality.


src/controllers/export.ts
```diff
@@ -121,28 +121,34 @@ export async function streamUploadedChildren(
...
   }
-  if (typeof value == 'string') {
+  if (typeof value === 'string') {
     return value || '';
   }
-  if (typeof value == 'number') {
+  if (typeof value === 'number') {
     return value.toString();
   }
+  if (propertyName === 'site') {
```
mmazanec22, line 143: This was the actual problem with exporting site data

src/controllers/export.ts
```diff
@@ -121,28 +121,34 @@ export async function streamUploadedChildren(
  * in cells of the CSV to export.
  * @param value
  */
-function formatStringPush(value: any) {
+function formatStringPush(value: any, propertyName: string) {
```
jlhogan, line 124: nit: this function name doesn't mean anything to me

src/controllers/export.ts
jlhogan: nit: could we put all these specific property name chekcs together (i.e. pull the funding space check down here?)

jlhogan:
>When I download an example file and upload it, there are still a few funding types that aren't parsed. I can't really make any sense of it, except maybe that the combination of funding type/time/whatever is not supported by that org? Any help on that would be appreciated, though it is probably not essential for launch if it's just a problem with the fake data quality.

can you provide an example ? Should be easy enough to check if the failure is b/c there is no matching funding space (e.g. is not supported by the org)

jlhogan (approved):

RyanHansz:
@mmazanec22 I don't believe this is a problem with fake data. I'm having batch upload issues with data I enter myself that should be valid.
Example:
[SCenariotesting.xlsx](https://github.com/ctoec/data-collection/files/5519479/SCenariotesting.xlsx)

mmazanec22:
Ok so the anomalous ones were all cases where:
- the child had only one enrollment
- the funding time was AdditionalExtendedSchool

RyanHansz:
> Ok so the anomalous ones were all cases where:
> 
> * the child had only one enrollment
> * the funding time was AdditionalExtendedSchool

Added issue with SHS funding spaces here: 
https://github.com/ctoec/data-collection/issues/718

src/controllers/export.ts
```diff
@@ -121,28 +121,35 @@ export async function streamUploadedChildren(
...
-    return value ? 'Yes' : 'No';
+  if (
+    propertyName === 'firstReportingPeriod' ||
+    propertyName === 'lastReportingPeriod'
+  ) {
+    return reportingPeriodToString(value);
   }
-  if (isMoment(value)) {
-    return value.format('MM/DD/YYYY');
+  // Name matching means some values are nested one more level
```
bamader, line 140: Nit: can delete this since you have the comment about checking for specific names

mmazanec22:
Ok I figured out the issue for #718 -- the funding space table was only allocating 20 characters to funding times.  Migration is included in this PR.

bamader (approved):

jlhogan:
>Ok I figured out the issue for #718 -- the funding space table was only allocating 20 characters to funding times. Migration is included in this PR.

I have a change in an open PR that changes the way we're doing enums to not use varchar with arbitrary length and actually enforce enum value check at DB level so things like this should never happen again :)

PR #710, opened 2020-11-10
235 (Pt I): Add uniqueId to child and uniqueIdType to org

jlhogan:
abandoning moves to make DB enforce enum checks b/c dropping & recreating columns in data-ful envs will fail, and creating migrations that do this neatly is a drag b/c SQL server is annoying about renaming thinks with constraints.

If we decide we want enum values enforced at db-level, we can wipe all data and do the migrations, but then if we ever wanna change the enums we're in a very annoying place again so i think we should stick with enforcing on the application side for now


client/src/shared/models/CareModel.ts
```diff
@@ -1,6 +1,6 @@
 export enum CareModel {
-  inPerson = 'In-person',
-  hybrid = 'Hybrid',
-  distance = 'Distance',
-  unknown = 'Unknown',
+  InPerson = 'In-person',
```
jlhogan, line 2: upper cased for consistency

client/src/shared/models/BirthCertificateType.ts
```diff
@@ -1,5 +1,5 @@
 export enum BirthCertificateType {
   US = 'US birth certificate',
-  nonUS = 'Non-US birth certificate',
-  unavailable = 'Birth certificate unavailable',
+  NonUS = 'Non-US birth certificate',
```
jlhogan, line 3: upper cased for consistency

bamader (approved):
LGTM

mmazanec22 (approved):

PR #720, opened 2020-11-11
291 withdrawn

### LINKED ISSUE #291: Withdrawn enrollments: I can navigate from my roster to a view of all withdrawn enrollments.
# [Design in Abstract](https://share.goabstract.com/1545741f-51be-4382-a6aa-76bd918553eb)

– I can click on past enrollments on my roster view to see withdrawn enrollments.
Screen Shot 2020-10-19 at 9 36 33 AM

– I see withdrawn enrollments in a roster view sorted by the most recent withdrawal date and grouped by age group.
Screen Shot 2020-10-19 at 9 43 44 AM

- I can return to my current roster by clicking the link under the withdrawn enrollments icon label

– When I withdraw an individual child record it is shown in the withdrawn roster. 

– Do not show withdrawn records in the active roster.

mmazanec22:
Closes #291 

Depends on https://github.com/ctoec/component-library/pull/93


client/src/containers/Roster/hooks/useOrgSiteProps.ts
```diff
@@ -28,7 +25,11 @@ export const useOrgSiteProps = (isLoading: boolean, childCount: number) => {
       // Remove site param if clickedId !== current orgId
       if (clickedId !== query.organization) delete query.site;
       history.push({
-        search: stringify({ ...query, organization: clickedId }),
+        search: stringify({
+          ...query,
+          organization: clickedId,
+          site: undefined,
```
jlhogan, line 31: why is `site: undefined` needed? doesn't the `delete query.site` above accomplish the same thing?

client/src/containers/Roster/hooks/useUpdateRosterParams.ts
```diff
@@ -10,11 +11,7 @@ export const useUpdateRosterParams = () => {
   const isMultiOrgUser = user?.organizations || [];
 
   const history = useHistory();
-  const query = parse(history.location.search.slice(1)) as {
-    organization: string;
-    site: string;
-    month: string;
-  };
+  const query = parse(history.location.search.slice(1)) as RosterQueryParams;
```
jlhogan, line 14: if we update this file to use `query-string` instead of `querystring` cn get rid of this slice

client/src/containers/Roster/rosterUtils.tsx
```diff
@@ -33,21 +33,25 @@ export const getQueryMonthFormat = (month?: Moment) => {
...
-
   if (site) {
     filteredChildren = filteredChildren.filter(
       (child) => getCurrentEnrollment(child)?.site?.id.toString() === site
     );
   }
-
+  if (showOnlyWithdrawnEnrollments) {
+    filteredChildren = filteredChildren.filter(
+      (c) => c.enrollments?.findIndex((e) => !e.exit) === -1
```
jlhogan, line 52: i think more symantically clear: `c.enrollments?.every((e) => e.exit))`

jlhogan (approved):
does this remove withdrawn children from the normal roster? I believe that's a desired behavior. Which then makes me think, should this just be filtering we do on the backend?  I think we probably should be doing all date filtering in the server, really.

mmazanec22:
@RyanHansz the design shows that there isn't a column for missing info icon for withdrawn enrollments.  Will users be able to/need to correct that data?  I'm a little afraid that the inconsistency between the regular and withdrawn rosters will make people think that records are complete when they aren't

RyanHansz:
> @RyanHansz the design shows that there isn't a column for missing info icon for withdrawn enrollments. Will users be able to/need to correct that data? I'm a little afraid that the inconsistency between the regular and withdrawn rosters will make people think that records are complete when they aren't

Yes, the withdrawn table should be the same as the active roster. I'll reflect that in the design momentarily

RyanHansz:
Added https://github.com/ctoec/data-collection/issues/740 which edits alert text when a user batch uploads withdrawn records with missing info.

bamader (approved):
LGTM!

PR #741, opened 2020-11-11
Defeat tricky funding parsing

### LINKED ISSUE #540: Make funding reporting periods accept more date formats.

### LINKED ISSUE #717: Remove "+" from SHS funding spaces
Excel turns them into a formula :( 

Additional full-day/full-year | "9" or "FD/FY"
-- | --
Additional school-day/school-year | "10" or "SD/SY"
Additional school-day/school-year/extended-year | "11" or "SD/SY/EY"

bamader:
Expands types of formats that can be parsed as date values by the system. Closes #540 . Whoever left the comment in `mapUtils/funding/mapFunding` that the find query "wasn't working", thank you greatly!

Update: Throwing in changes to the SHS funding types cuz it's pressing delete three times. Closes #717 .

Non-essential details if you're interested: I see now why this was nefarious to solve, because it ended up blending the simple fix Julia thought with the SQL weirdness Melanie saw. For enrollment and funding _parsing_, simply adding to the array of acceptable format strings perfectly solved the problem. Everything gets correctly parsed as a moment. The problem for *fundings* on the Roster (which is the image Ryan linked) was in the retrieval. Our conditions required an exact match on date values (we couldn't use `.id` like we did in the rest of the system because we were doing a surface-level values check), which was leading to lots of funding periods not being returned. Instead, just matching on the month and the year of a funding period to return the customary format solved the problem.


src/controllers/enrollmentReports/mapUtils/funding.ts
```diff
@@ -62,15 +62,29 @@ export const mapFunding = async (
   // TODO: Cache ReportingPeriods, as they'll be reused a lot
   let firstReportingPeriod: ReportingPeriod,
     lastReportingPeriod: ReportingPeriod;
+
+  // Use some string information from the relevant moments to perform
+  // "fuzzy" matching based on SQL's `LIKE` expression (match a pattern)
+  // In this case, pure moments are stored as `Moment<YYYY-MM-DD>`
   if (source.firstReportingPeriod) {
+    // .month() counts January as 0, so increment by 1
+    let monthString = (source.firstReportingPeriod.month() + 1).toString();
```
jlhogan, line 71: I dont understand why fuzzy matching is necessary -- the `period` strings in the DB are all `YYYY-MM-01`; shouldn't we be able to enforce the moment we use here matches that (i.e. is the first), and then the moment transformer will format it as the YYYY-MM-DD string for DB querying ?

bamader: Hmm that's actually a good point. If we expect the dates in the DB to all be the first, then we shouldn't be running into this problem (I thought that reporting periods could all have different days, but I might have been looking at the arbitrary stuff in the fake data generators). I'll investigate more.

bamader: Aaaaand no sooner do I post that than I see that one of the formats (`MM/YYY`) is getting parsed as having a non-first date (I used 03/2020 in my date testing sheet and this got parsed as 03/20/2020).

jlhogan: we control the creation of reporting periods, so we can be confident that the `period` field will always be a 1st :) other date fields (periodStart, periodEnd, dueAt) will have various dates

bamader: I figured it out (it's real dumb lol). The moment format parser does two things that are buried in the docs: 1) it ignores separator characters in all given formats, and 2) it stops at the _first_ valid format it finds. The ordering of our formats had MM-DD-YY before MM/YYYY, so without separators, the middle 20 was being parsed as a day of the month rather than part of the year because both expected 6 characters 🙄

bamader: So this also means that since we're supporting `MM/DD/YYYY` formats for funding reporting periods, if the user uploads a date in this format, should we just hard change the `DD` to be `01`?

jlhogan: oooo good find!! that is indeed tricky! 
> What's the desired behavior if a user uploads a reporting period date in MM/DD/YYYY format where the DD isn't 01?

ignore the DD

src/controllers/enrollmentReports/mapUtils/funding.ts
```diff
@@ -64,7 +64,6 @@ export const mapFunding = async (
     lastReportingPeriod: ReportingPeriod;
   if (source.firstReportingPeriod) {
     firstReportingPeriod = await transaction.findOne(ReportingPeriod, {
-      // TODO: this comparison doesn't work-- need fuzzier matching
```
bamader, line 67: Since we control the creation of reporting periods and can set the day of the month to always be the first, we can just ignore day values in parsed dates and not need fuzzy matching

src/template/decorators/ColumnMetadata.ts
```diff
@@ -4,18 +4,20 @@ import { ColumnMetadata as ColumnMetadataInterface } from '../../../client/src/s
...
+  'MM/YY',
+  'MMM/YY',
+  'MM/YYYY',
+  'MMM/YYYY',
   'MM/DD/YYYY',
-  'MM-DD-YYYY',
   'MM/DD/YY',
-  'MM-DD-YY',
-  'YYYY-MM-DD',
+  'YYYY/MM/DD',
```
bamader, line 13: Moment parser ignores separator characters, so we don't need two copies of things with `/` and `-`

src/controllers/enrollmentReports/parse.ts
```diff
@@ -170,7 +170,18 @@ function parseEnrollmentReportRow(
 
     // Parse dates
     if (dateProperties.includes(prop)) {
-      rawEnrollment[prop] = getDate(value);
+      let parsedDate = getDate(value);
+      // Override the day value to the first of the month
+      // for funding periods
+      if (parsedDate && prop.toLowerCase().includes('period')) {
+        // .month() counts Jan as 0, so increment by 1
+        let month = (parsedDate.month() + 1).toString();
```
jlhogan, line 178: does
```
if (parsedDate && prop.toLowerCase().includes('period')) {
  parsedDate.startOf('month')
}
``` 
work?
should modify the moment in place, so i'd advocate for passing the prop name into the `getDate` util, and doing this conditional check in that func

bamader: Indeed it does work! Didn't know about that function, it def makes things simpler. And yeah I can push the check into the util, too.

src/template/decorators/ColumnMetadata.ts
mmazanec22: A comment in the code about this would be helpful for future devs debugging date issues-- esp if you link to the place in the moment docs where it says that!

mmazanec22 (approved):

src/template/decorators/ColumnMetadata.ts
bamader: V fair--will put in a comment now!

src/template/decorators/ColumnMetadata.ts
```diff
@@ -3,6 +3,11 @@ import { ColumnMetadata as ColumnMetadataInterface } from '../../../client/src/s
 
 // Formats
 export const BOOLEAN_FORMAT = 'Yes, Y, No, N';
+// Note: The Moment parser ignores separator characters in format strings, so don't
```
mmazanec22, line 6: Yay!

mmazanec22 (approved):

jlhogan (approved):

PR #745, opened 2020-11-11
682 e2e ci

mmazanec22:


jlhogan (approved):
woooooo

e2e-tests/tests/deleteChild.js
```diff
@@ -4,6 +4,14 @@ const {
   clickOnFirstChildInRoster,
 } = require('../utils/clickOnFirstChildInRoster');
 
+/**
+ * TODO: Deleting a record doesn't create an alert even in the most
+ * up to date local version of the app. So what we need to do is find
+ * the number of children on the  roster before the delete, then
+ * compare it to the number of children on the roster after the delete.
+ * If the diff is 1, we gucci.
```
mmazanec22, line 12: There's a ticket open to have an alert, so we should probably just comment this test out and add a todo in that ticket

bamader: Sounds good. I just pushed some incomplete work here because I felt like I wasn't making any productive progress after a lot of head-to-wall banging so  I needed to take a break and make some kind of actual contribution (on  the help  page branch). I'm actually thinking most of the tests are still good, there are just a few bugs that are preventing them from running correctly, so I think post bug-bash is a good time for me to revisit this.

e2e-tests/tests/changeChild.js
```diff
@@ -28,11 +28,15 @@ module.exports = {
...
-    // Then navigate to roster and see if that text is on the roster
     await navigateToRoster(browser);
-    const tableArgs = ['xpath', '//*/table[1]'];
-    await scrollToElement(browser, tableArgs);
-    browser.assert.containsText('table', newFirstNameText);
+
+    // We've changed/expanded the table header, so access the
+    // changed child and check if the form field has the
+    // correct value isntead
+    await clickOnFirstChildInRoster(browser);
```
mmazanec22, line 36: How are kids ordered in the roster by default?  Will that have any effect on this?

jlhogan: they're sorted alphabetically by last name

bamader: They're ordered by age group (alphabetically I think), which I think we've defined explicitly _somewhere_ in the app so that uses have a consistent layout every time. And nope, that shouldn't affect this at all because the same `clickOnFirstChild` function gets used for both the change and the check in the test, with no enrollments or additional records happening in between--no other child can usurp them!

bamader (approved):

e2e-tests/tests/changeChild.js
jlhogan: the data comes back sorted by child id, is then separated into age gruop sections, and sorted by last name (which is defined in the table instantiation)

PR #752, opened 2020-11-12
Enforce -1 capacity for SHS spaces

### LINKED ISSUE #684: Allow SHS spaces to be entered without an overall capacity #.
Any user with SHS funding should be able to choose SHS funding options for an enrollment.

On the roster, SHS should not show # enrolled/# capacity, just # enrolled.

mmazanec22:
Closes #684


jlhogan (approved):
nice! Altho, I don't think we do (or necessarily _should_) run validations for "metadata" like funding spaces so not sure validating actually does anything, but it's any easy way to document what that means so i have no objection

bamader (approved):

JordanGuinn (approved):

PR #776, opened 2020-11-13
Handle nulls/undefineds

### LINKED ISSUE #767: After trying to add a new child record, if the record doesn't save, refreshing on the roster page makes all records disappear
```
{"message":"TypeError: Cannot read property 'incomeDeterminations' of null"}
```

### LINKED ISSUE #768: BatchEdit says all records are complete even if an alert on the roster page says enrollments are missing information
```
{"message":"TypeError: Cannot read property 'incomeDeterminations' of null"}
```

### LINKED ISSUE #770: make soft delete filter handle null/undefined

bamader:
Makes soft delete filtering handle if `family`  is undefined, as well as  if  the rows to filter on are  undefined. Closes #770 .
closes #768 
closes #767

### DIFF
--- a/src/utils/filterSoftRemoved.ts
+++ b/src/utils/filterSoftRemoved.ts
@@ -6,8 +6,7 @@ import { Child, Enrollment } from '../entity';
  * soft-deleted.
  * @param rows
  */
-export const removeDeletedElements = (rows: any[]) => {
-  if (!rows) return [];
+export const removeDeletedElements = (rows?: any[]) => {
   return rows.filter((row: any) => !row.deletedDate);
 };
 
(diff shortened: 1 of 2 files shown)


src/utils/filterSoftRemoved.ts
```diff
@@ -1,13 +1,13 @@
...
 /**
  * Given an array of 'rows' (assumed to be a property of some
  * entity), filter out any objects in those rows that have been
  * soft-deleted.
- * @param entity
- * @param property
+ * @param rows
  */
 export const removeDeletedElements = (rows: any[]) => {
+  if (!rows) return [];
```
jlhogan, line 10: this is incompat with the function signature, but i think it makes sense to offload this check to here. I would update this to be 
```
export const removeDeletedElements = (rows?: any[]) => {}
```
and then you can remove the `|| []` from all the calls below

jlhogan (approved):

src/utils/filterSoftRemoved.ts
```diff
@@ -1,13 +1,12 @@
...
 /**
  * Given an array of 'rows' (assumed to be a property of some
  * entity), filter out any objects in those rows that have been
  * soft-deleted.
- * @param entity
- * @param property
+ * @param rows
  */
-export const removeDeletedElements = (rows: any[]) => {
+export const removeDeletedElements = (rows?: any[]) => {
```
jlhogan, line 9: sorry if i wasn't clear. By making this undefined, you now make the function signature align with the reality that you may be passing in undefined here. Still absolutely need to do the undefined check tho ( `if (!rows) return[]; `)

jlhogan (approved):

src/utils/filterSoftRemoved.ts
bamader: Oh that's what you meant about the signature. Gotcha, fixing now.

jlhogan (approved):

PR #796, opened 2020-11-17
Disable submit when not in prod

### LINKED ISSUE #634: Disable submit: Hide the ability to submit data until we are ready to start accepting it.
Since we are providing access to programs in December, but the data collection doesn't start until January, we need to disable the submit to OEC button on the roster. 
- disable in prod only
- Blocked until production environment is deployed

### LINKED ISSUE #779: add org id checks to all object get/create/update

bamader:
Disables submitting to OEC while the app isn't live, and also adds some org ID checks to object creation / updates / gets.
Closes #634 
Closes #779 

I might be missing some of the context for the Org ID check ticket because where I was looking, it seems like we already mostly have that in place (i.e. the child controllers do org checks, the funding controller does checks, most of the routes already filter, etc.). I just wound up updating families, income determinations, and enrollments (although that might've been the point in hindsight).

### DIFF
--- a/client/src/containers/Roster/Roster.tsx
+++ b/client/src/containers/Roster/Roster.tsx
@@ -210,7 +210,11 @@ const Roster: React.FC = () => {
                 : 'Send to OEC'
             }
             onClick={submitToOEC}
-            disabled={!query.organization}
+            // TODO: Submit button must be disabled until the app is in a prod environment
+            disabled={
+              !query.organization ||
+              process.env.REACT_APP_STAGE !== 'production'
+            }
           />
         )}
       </FixedBottomBar>
--- a/client/src/containers/Roster/Roster.tsx
+++ b/client/src/containers/Roster/Roster.tsx
@@ -210,11 +210,9 @@ const Roster: React.FC = () => {
                 : 'Send to OEC'
             }
             onClick={submitToOEC}
-            // TODO: Submit button must be disabled until the app is in a prod environment
-            disabled={
-              !query.organization ||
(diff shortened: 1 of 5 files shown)


client/src/containers/Roster/Roster.tsx
```diff
@@ -210,7 +210,11 @@ const Roster: React.FC = () => {
                 : 'Send to OEC'
             }
             onClick={submitToOEC}
-            disabled={!query.organization}
+            // TODO: Submit button must be disabled until the app is in a prod environment
+            disabled={
+              !query.organization ||
+              process.env.REACT_APP_STAGE !== 'production'
```
jlhogan, line 216: I think the AC was unclear, but in prod we want submission disabled until Jan, so we just want to make this button be always disabled, and later we can push a new version that has the button functioning normally.

jlhogan (commented):
the orgId filtering updates look good, but button needs to always be disabled in prod env (until jan) -- i would switch the condition to disable the button for only the prod env (we will likely want to test submission functionality in pre-prod envs ahead of enabling submission for users in jan)!

client/src/containers/Roster/Roster.tsx
bamader: Mmm that makes sense. I'll post a full disable then but leave the original code commented out, and we can just flip it later.

mmazanec22 (approved):

jlhogan (approved):
yup! that was indeed the point (at least in my mind!) :)

PR #809, opened 2020-11-18
Roster buttons and e2e fixes

### LINKED ISSUE #593: For single-site users and views: Remove ratios, just display # of children per funding type
If I am viewing a single site's data, then I should not see the ratio of funding spaces filled (example: 15/15 CDC Full-Time spaces filled)

Instead, I should see: 15 CDC Full-Time spaces filled

### LINKED ISSUE #634: Disable submit: Hide the ability to submit data until we are ready to start accepting it.

### LINKED ISSUE #682: run e2e tests in CI
against staging after deploy
Someone who is not Jordan needs to do this!

bamader:
Fast follows to fix some bugs with the roster button hiding (better matches AC now), single site view capacity (hides the ratio), and e2e tests (adds retries to combat their flakey nature).

Closes #682 
Closes #634 
Closes #593

### DIFF
--- a/e2e-tests/package.json
+++ b/e2e-tests/package.json
@@ -4,8 +4,8 @@
   "description": "End to end tests for OEC data collection tool",
   "main": "login.js",
   "scripts": {
-    "test": "node_modules/nightwatch/bin/nightwatch -e default",
-    "test-edge": "node_modules/nightwatch/bin/nightwatch -e edge --skiptags skip-edge"
+    "test": "node_modules/nightwatch/bin/nightwatch -e default --retries 3 --suiteRetries 3",
+    "test-edge": "node_modules/nightwatch/bin/nightwatch -e edge --skiptags skip-edge --retries 3 --suiteRetries 3"
   },
   "license": "MIT",
   "dependencies": {
--- a/client/src/containers/Roster/Roster.tsx
+++ b/client/src/containers/Roster/Roster.tsx
@@ -11,7 +11,9 @@ import {
 import { stringify, parse } from 'query-string';
 import moment from 'moment';
 import UserContext from '../../contexts/UserContext/UserContext';
-import { FixedBottomBar } from '../../components/FixedBottomBar/FixedBottomBar';
+// TODO: Uncomment this import when you want to reactivate the
+// bottom bar with buttons
+// import { FixedBottomBar } from '../../components/FixedBottomBar/FixedBottomBar';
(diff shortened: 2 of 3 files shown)


jlhogan (approved):

mmazanec22 (approved):

e2e-tests/package.json
```diff
@@ -4,8 +4,8 @@
   "description": "End to end tests for OEC data collection tool",
   "main": "login.js",
   "scripts": {
-    "test": "node_modules/nightwatch/bin/nightwatch -e default",
-    "test-edge": "node_modules/nightwatch/bin/nightwatch -e edge --skiptags skip-edge"
+    "test": "node_modules/nightwatch/bin/nightwatch -e default --retries 3 --suiteRetries 3",
```
JordanGuinn, line 7: Is there a ticket somewhere to follow-up on the e2e tests that warrant these retries?  I'm certainly missing some context here, but having to regularly re-run our test suite cuz certain tests are inconsistent seems like a problem

bamader: No ticket, but the various `retry` parameters only rerun the modules that fail, not necessarily all of them. For context, I'd been working on updating the e2e tests for a few days last week, and then was debugging some of them two days ago because I thought they were blocking some code from going to staging (which was completely wrong, but led me to look into repeated runs of our test suites). I saw a bunch of suuuuuper flaky behavior for literally _all_ of the tests. In one run, a test (ChangeChild, for example) would fail with an `element not found` error. So I'd watch the browserstack video, and there would be literally no unexpected behavior or error state happening in the app; it would get to some point and then the next step would just fail. Frequently, rerunning the test solved the issue and made the test succeed, even if nothing got changed. Julia and I saw flakiness for basically all of the tests, so a retry of tests that failed was our way of combatting tests that would seemingly for no reason just not work one run.

JordanGuinn: Ahhh, I see - I suppose that only re-running the modules that failed instead of the whole suite reduces the potential for false positives from cropping up.  Still probably worth revisiting when things aren't as hectic though, just cuz accepting inconsistent behavior from our test suite, for _whatever_ reason, isn't a great precedent to set

JordanGuinn (approved):

PR #811, opened 2020-11-19
Tweaks 11/18

### LINKED ISSUE #792: Error modal flags multiple enrollment rows with missing info
Uploading sample data shows:
Screen Shot 2020-11-16 at 5 57 10 PM

jlhogan:
1. Closes #792 
     - fixes bad default value in organization table for uniqueIdType column 
     - fixes is identifier match check by doing double-equals comparison of sasid/unique id fields to account for null/undefined 
     - Fixes up the update child stuff so that we don't try to save entities when we shouldn't and all entities that need Ids have them 

~1. Reduces the number of network requests the Roster page makes:
    - condenses the withdrawn/active enrollments request into one, since we always get both from the roster home page
    - makes the useOrgSiteProps hook return things as state variable, which for whatever reason drastically cut down on number of requests issued~ this turned into a melodrama, and we will figure out a more robust fix not at 11 pm

~1. Makes sure that on refetch of data, all pages are re-fetched:
    - adds some more weird logic to usePaginatedChildData hook~ same as above

2. Makes data fully recreate in deployed envs, and makes default test user that's created by app have single-org (Hogwarts) permissions

3. Makes enrollment reports controller clean up /tmp upload files afte rthey're parsed

### DIFF
--- a/src/controllers/enrollmentReports/mapUtils/child.ts
+++ b/src/controllers/enrollmentReports/mapUtils/child.ts
@@ -77,7 +77,8 @@ export const updateChild = async (
   organization: Organization,
   site: Site,
   child: Child,
-  user: User
+  user: User,
+  save: boolean
 ) => {
   let enrollment = getExistingEnrollmentOnChild(source, child);
   const modifyingExistingEnrollment = enrollment !== undefined;
@@ -87,13 +88,14 @@ export const updateChild = async (
     site,
     child,
     user,
-    enrollment === undefined
+    save && enrollment === undefined
   );
 
   // Apply any needed enrollment info updates before going to funding
   // TODO: this pattern is unnecessary because save updates an entity if it existed before-- we should refactor to just save
-  if (modifyingExistingEnrollment) {
+  if (save && modifyingExistingEnrollment) {
     await transaction.update(Enrollment, enrollment.id, enrollmentUpdate);
+    enrollmentUpdate.id = enrollment.id;
   }
 
   let funding = getExistingFundingForEnrollment(source, enrollment);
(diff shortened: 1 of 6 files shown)


src/controllers/enrollmentReports/mapUtils/child.ts
```diff
@@ -267,11 +269,12 @@ export const isIdentifierMatch = (
   if (child instanceof Child) {
     match =
       match &&
-      (child.sasid === other.sasidUniqueId ||
-        child.uniqueId === other.sasidUniqueId);
+      // NOTE: use == to evaluate undefined from 'other' and
+      // null from 'child' to be equivalent
+      (child.sasid == other.sasidUniqueId ||
+        child.uniqueId == other.sasidUniqueId);
```
bamader, line 275: I don't know whether to laugh or cry at it being a `==` instead of `===`. I feel like every other place in the system react complains about `==` 🙃

bamader (approved):
LGTM!

src/controllers/enrollmentReports/mapUtils/child.ts
jlhogan: react complaining about `==` is just a linting thing -- generally using strict checking (===) is best practice, however the cleanest / easiest way to check for null OR undefined in addition to comparing values is the loose equality check

mmazanec22 (approved):

src/routes/enrollmentReports.ts
```diff
@@ -69,6 +65,8 @@ enrollmentReportsRouter.post(
         throw new BadRequestError(
           'Your file isn’t in the correct format. Use the spreadsheet template without changing the headers.'
         );
+      } finally {
+        if (req.file && req.file.path) fs.unlinkSync(req.file.path);
```
JordanGuinn, line 69: ahhhhhh, good call

JordanGuinn (approved):

PR #817, opened 2020-11-20
Fix alert displays

### LINKED ISSUE #451: Batch upload confirmation: After I upload a spreadsheet to my roster, I see an alert confirming the upload.
Screen Shot 2020-11-11 at 5 04 02 PM
- Show a confirmation alert whenever a user successfully batch uploads a file.
- Show the missing info alert below the success alert if there are records with missing/incorrect info.

### LINKED ISSUE #725: Add confirmation alert after a record has been withdrawn or deleted.
Screen Shot 2020-11-12 at 12 40 58 PM

Screen Shot 2020-11-12 at 12 40 26 PM

bamader:
Improve alerts on the roster page by getting rid of clearing all alerts if there are no children with errors (this was causing the alerts for delete record and add record to not show up in the `history.push('/roster', { alert } )` calls). Also creates an alert that confirms a batch upload is successful and reports how many active and withdrawn records were parsed.

Closes #725 
Closes #451 

Still trying to figure out what's going on with the test failing


client/src/containers/Upload/Upload.tsx
```diff
@@ -83,10 +83,25 @@ const Upload: React.FC = () => {
         accessToken,
         rawBody: true,
       })
-        .then(() => {
+        // Response contains id of created enrollmentReport,
```
jlhogan, line 86: could make this a shared [payload](https://github.com/ctoec/data-collection/tree/base/client/src/shared/payloads) for some type checking in the future should this need to change

jlhogan (approved):

client/src/containers/Roster/hooks/useChildrenWithErrorsAlert.tsx
```diff
@@ -21,7 +21,7 @@ export const useChildrenWithErrorsAlert = (
   );
   useEffect(() => {
     if (isLoading || !childrenWithErrorsCount) {
-      setAlerts([]);
+      setAlerts([...alerts]);
```
mmazanec22, line 24: Why not just remove the set alerts here?

mmazanec22 (approved):

client/src/containers/Roster/hooks/useChildrenWithErrorsAlert.tsx
bamader: Initially I had it removed entirely but that was causing the roster test to fail (with that `jestfn()` calls expected 1 got 0 thing I mentioned in slack last Friday). I have no idea why not explicitly setting the alerts to what they already are would cause the apiMock to not get called. Jordan made an interesting point about the call potentially being made before a component render, so I thought this might make the error setting 'explicit' enough for the test. Which really doesn't make much sense (totally acknowledge that) but it seemed like a very small thing to keep spending cycles on. If that's the wrong call though I can def go back and try to debug it, was just not making any meaningful progress.

mmazanec22: Sounds like a weird one!  How about a TODO note then, since this needs to exist due to weirdness and should be re-investigated at some point in the future?

bamader: Yep yep good call! Sometimes I forget I can use comments to make note of weirdness instead of writing my usual overly long function pre-ambles 😂

PR #820, opened 2020-11-21
Fix error flashing

### LINKED ISSUE #561: Add enrollment forms are flashing error validations when moving between sections.
When I move between sections while adding a child, the forms get an error validation applied to them. This makes me think I have incorrectly entered data.

- At least happening when finishing the child info section in staging

mmazanec22:
Closes #561 I think??  I had a hard time reproducing the error

Edit: refactored a bit, depends on https://github.com/ctoec/component-library/pull/99


client/src/components/Forms/FamilyIncome/Form.tsx
```diff
@@ -93,6 +96,11 @@ export const FamilyIncomeForm: React.FC<FamilyIncomeFormProps> = ({
     );
   }
 
+  const { obj: child, setErrorsHidden } = useValidationErrors<Child>(
```
jlhogan, line 99: does this actually work for all forms? Because lookng at this hook, it only seems to set `validationErrors` to undefined on the top-level object (i.e. the child) and so i would expect any already-distributed-by-the-backend errors to still exist on the enrollments/fundings/family/income determinations. I wonder if there's not some simpler way to do this where we create a context that holds a flag like `shouldDisplayFormErrors` and then make all form fields use that flag to decide if they should display an error or not. 

Wouldn't consider that ^ blocking if this does indeed work, but if we do need to do deeper removal of `validationErrors` on these objects to effectively hide all errors in the front end, then we'd prob want to retain two copies of the object (one with errors, one without) and return the appropriate one given the flag as opposed to recreating the error-less object each time

mmazanec22: Yeah I guess it's time to actually refactor that-- will make it a context that passes back status funcs

mmazanec22:
@jlhogan I refactored this to be less hacky

client/package.json
```diff
@@ -43,7 +43,7 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen",
+    "test": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --silent",
```
jlhogan, line 46: does this hide all that awful to look at console log!!????

jlhogan (approved):

client/package.json
mmazanec22: Yes!  I was so sick of combing through those react dates warnings

bamader: I wish there were a tears of joy emoji

bamader (approved):

JordanGuinn (approved):

PR #826, opened 2020-11-23
Confidentiality agreement

mmazanec22:
Closes #824 

Note: I think I'll need to add an e2e util before merging

PR #827, opened 2020-11-23
825 one time confidentiality

### LINKED ISSUE #825: The confidentiality agreement is only shown when a user logs in and agrees for the first time.

mmazanec22:
Closes #825 

Fast follow for #826 and branched from that PR

### DIFF
--- a/src/routes/users.ts
+++ b/src/routes/users.ts
@@ -3,7 +3,7 @@ import { getManager } from 'typeorm';
 import { User } from '../entity';
 import { passAsyncError } from '../middleware/error/passAsyncError';
 import * as controller from '../controllers/users';
-import { BadRequestError } from 'src/middleware/error/errors';
+import { BadRequestError } from '../middleware/error/errors';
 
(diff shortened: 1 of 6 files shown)


src/routes/users.ts
```diff
@@ -15,3 +19,18 @@ usersRouter.get('/current', async (req, res) => {
   await controller.addDataToUser(user);
   res.send(user);
 });
+
+usersRouter.post(
+  '/:userId',
+  passAsyncError(async (req, res) => {
+    try {
+      const user = req.user;
```
jlhogan, line 27: since we don't user the `userId` param at all, maybe make this a post to `users/current` ?

jlhogan: alternative is to add a check that `userId` === `req.user.id` but i dont see that as conferring any benefit

jlhogan (approved):

bamader (approved):
Nice!

JordanGuinn (approved):

PR #832, opened 2020-11-24
Names where errors occur

### LINKED ISSUE #722: Add "name" or "row #" column to missing info modal when batch uploading.
Screen Shot 2020-11-23 at 9 18 49 AM
- Offset based on the template the user is using (excel or .csv)

### LINKED ISSUE #829: zipcodes are being parsed wrong again
01920 in spreadsheet becomes long negative # in staging

add a test!!

### LINKED ISSUE #831: Add back button on data template page

bamader:
Adds names of the child records where each type of error occurs. Closes #722 
Adds a back button to the data template page. Closes #831 
Fixes a capitalization typo that led to zip codes being parsed as negative longs. Closes #829


src/controllers/enrollmentReports/checkErrors.ts
```diff
@@ -13,15 +13,25 @@ import { ValidationError } from 'class-validator';
...
+  child: Child,
+  error: ValidationError,
+  errorDict: Object,
+  errorOccursIn: Object
+) => {
   // Base case: error is not in a nested field
   if (error.children.length === 0) {
     errorDict[error.property] += 1;
+    errorOccursIn[error.property].push(
+      (child.firstName || '') + ' ' + (child.lastName || '')
```
mmazanec22, line 26: `${child.firstName} ${child.lastName}`

mmazanec22 (approved):

e2e-tests/tests/zipCodeParse.js
```diff
@@ -0,0 +1,85 @@
+const { login } = require('../utils/login');
```
jlhogan, line 1: this is fine, but definitely does not _need_ to be tested as an e2e test. They take the longest to run, and so we want to use them for things that really require the full end to end app -- that is, things that are about how the front end and back end work together. I think for things like this, it makes more sense to add a  backend integration test, since we can fully test this functionality with only the server-side

src/controllers/enrollmentReports/parse.ts
```diff
@@ -174,8 +174,8 @@ function parseEnrollmentReportRow(
     }
 
     // Parse zipcodes
-    if (prop.match('/zipcode/i')) {
-      rawEnrollment[prop] = getZipcode(value);
+    if (prop === 'zipCode') {
```
jlhogan, line 177: so to write a non-e2e test for this, you would test the `parseEnrollmentReportRow` function and the `getZipCode` function in a test like the exiting (map tests)[https://github.com/ctoec/data-collection/blob/base/src/controllers/enrollmentReports/map.test.ts]

bamader: My b! I misunderstood test as a strictly e2e test. I'll work on moving this to just a back end test in `map.test.ts`. If I'm reading how to structure one of these right, I should just be able to define a fake `raw object` with a zip code of 01920, then call `parseEnrollmentReportRow` on it and expect `getZipCode` to be called once. I _think_ that should be easy to do (famous last words 😛 ).

jlhogan: yes! i think you've got all that right, except you'll make a new test file for parse tests!

jlhogan: why doesn't `prop.match(/zipcode/i)` work?

src/controllers/enrollmentReports/parse.test.ts
```diff
@@ -0,0 +1,21 @@
...
+import { EnrollmentReportRow } from '../../template';
+
+describe('controllers', () => {
+  describe('enrollmentReports', () => {
+    describe('parseZipCode', () => {
+      it('parses a zip code with a leading zero correctly, not as a negative long', () => {
+        const dateProperties = [];
+        const booleanProperties = [];
+        // Assume excel got its hands on the zipcode and killed off a 0
+        const rawObj = { zipCode: '1920' } as EnrollmentReportRow;
```
jlhogan, line 11: while here wouldn't hurt to add test for the other special case -- string > 5 (which i think we maybe actually don't want to enforce? because what if folks add the extra specific XXXXX-XXXX form of a zip code? So that's actually a product question, so nevermind)

jlhogan (approved):
approve but note to follow up on:
- why can't we use a regex match?
- what zip-code parsing/formatting should we actually do and enforce? especially RE long-format zip codes

src/controllers/enrollmentReports/parse.test.ts
bamader: I'll open a ticket to talk about the extra specific zip codes and decide what to do with them!

src/controllers/enrollmentReports/parse.ts
bamader: Mmmm it works without the `'`. Looking it up, it seems that putting quotes around the regex causes it to first parse as a string, which makes the matching behavior wrong because of the `/`'s.

client/src/containers/Upload/ErrorModal/ErrorObjectForTable.ts
```diff
@@ -7,4 +7,5 @@ export type ErrorObjectForTable = {
   property: string;
   count: number;
   formattedName: string;
+  occursIn: string[];
```
JordanGuinn, line 10: Nit-pick: This contains all the row numbers for which this particular error applies, correct?  If so, thoughts on potentially naming this `affectedRows` or something?  Just cuz (at least for me) `occursIn` isn't immediately indicative of that fact

JordanGuinn (approved):
Looks good overall!

client/src/containers/Upload/ErrorModal/ErrorObjectForTable.ts
bamader: Sure, can do! I am _always_ about variable clarity.

PR #836, opened 2020-11-25
Update Confidentiality Agreement Column

### LINKED ISSUE #833: user object tracks timestamp of confidentiality agreement acceptance
update user to have a timestamp for acceptance of confidentiality notice instead of a boolean

JordanGuinn:
Closes #833.

- Converts data point tracking confidentiality agreement to date from boolean.
- Adds dummy data to non-prod seeder, to ensure we catch any issues with migrations in environments with existing data.


client/src/containers/App/ConfidentialityAgreement.tsx
```diff
@@ -63,16 +64,20 @@ export const ConfidentialityAgreement: React.FC = ({ children }) => {
...
             />
           </div>
           <div>
             <Button text="Logout" href="/logout" appearance="outline" />
             <Button
               text="Continue to ECE Reporter"
-              disabled={!_confidentialityAgreed}
-              onClick={() => setConfidentialityAgreed(true)}
+              disabled={!_confidentialityAgreedDate}
+              onClick={() => setConfidentialityAgreedDate(new Date())}
```
jlhogan, line 80: so this is very nitpicky, but since we're never actually using the date value of `_confidentialityAgreedDate` local state var, that could be updated to just be a bool that controlls the disabled-ness of this `continue` button, which then sets the date that is actually posted back to the API. OR use  the local `_confidentialityAgreedDate` here in the `setConfidentialityAgreed()` call. I just dont like an unused var ;)

jlhogan (approved):
Approved with caveat that staging DB will need to be modified before this can be merged:
- delete migration from `migrations` table
- drop `confidentialityAgreed` column from `user` table

client/src/containers/App/ConfidentialityAgreement.tsx
bamader: Was going to suggest the same thing. Seems like a boolean would work perfectly well. But def not a blocker.

bamader (approved):

src/entity/User.ts
```diff
@@ -30,8 +32,8 @@ export class User implements UserInterface {
   @Column({ nullable: true })
   suffix?: string;
 
-  @Column({ nullable: true })
-  confidentialityAgreedDate?: Date;
+  @Column({ nullable: true, type: 'date', transformer: momentTransformer })
```
jlhogan, line 35: :( because i think we gonna go back in at some point and getting rid of all moment stuff but :) for consistency for now !!

jlhogan (approved):

client/src/contexts/UserContext/UserContext.tsx
```diff
@@ -8,12 +8,14 @@ import React, {
...
 import { apiGet, apiPost } from '../../utils/api';
+import { Moment as MomentInterface } from 'moment';
+import * as moment from 'moment';
 
 export type UserContextType = {
   user: User | null;
   loading: boolean;
   setConfidentialityAgreedDate: Dispatch<SetStateAction<Date>>;
-  confidentialityAgreedDate: Date | null;
+  confidentialityAgreedDate: MomentInterface | null;
```
jlhogan, line 18: why `MomentInterface` over just `Moment`?

JordanGuinn: Oh it can just be `Moment` I suppose, now that I'm exporting * from `moment` as `moment`, lowercase-d.

jlhogan: can just be `import moment, {Moment} from 'moment';

PR #841, opened 2020-12-01
830 triple boolean

### LINKED ISSUE #830: in create record, triple booleans default to false, and cannot select race option except not disclosed
Additionally, cannot select "not disclosed" value in triple booleans in create record or edit record flow

mmazanec22:
Closes #830


src/migration/1606268723397-RemoveDefaultsFromChild.ts
```diff
@@ -0,0 +1,49 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class RemoveDefaultsFromChild1606268723397
+  implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    const dropDefaultConstraint = async (tableName, columnName) => {
```
jlhogan, line 6: at this point should we maybe add a `utils` dir to `migration` and pop this in there?

src/controllers/export.ts
```diff
@@ -128,7 +128,12 @@ function formatProperty(value: any, propertyName: string) {
     return reportingPeriodToString(value);
   }
   if (typeof value === 'boolean') {
-    return value ? 'Yes' : 'No';
+    if (value === true) {
```
jlhogan, line 131: does some special processing need to be added for formatting out the `not disclosed` option of undefineable booleans?

jlhogan (approved):

src/controllers/export.ts
mmazanec22: No, because they are not actually booleans (and maybe need a better name), so AFAIK they will not hit this if statement

jlhogan: duhhhhhhh :)

client/src/containers/CreateRecord/CreateRecord.tsx
```diff
@@ -103,7 +103,10 @@ const CreateRecord: React.FC = () => {
 
   const commonFormProps = {
     child,
-    afterSaveSuccess: triggerRefetchChild,
+    afterSaveSuccess: () => {
+      triggerRefetchChild();
+      setAlerts([]);
```
bamader, line 108: Does setting alerts to `[]` still allow the "Child record added to roster" alert to show?

bamader (approved):

PR #850, opened 2020-12-02
565 inactivity timeout

### LINKED ISSUE #565: Timed logged out: After a given period of time, I am automatically logged out and taken to the login screen.
- Clear out all cache data and user
if refresh token has expired, then do logout stuff

mmazanec22:
Closes #565

### DIFF
--- a/client/src/contexts/AuthenticationContext/AuthenticationContext.tsx
+++ b/client/src/contexts/AuthenticationContext/AuthenticationContext.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import { useHistory } from 'react-router-dom';
 import {
   AuthorizationRequest,
@@ -47,7 +47,6 @@ const { Provider, Consumer } = AuthenticationContext;
  */
 const defaultProps = {
   loginEndpoint: '/login',
-  defaultOpenIdConnectUrl: null,
   redirectEndpoint: '/login/callback',
   logoutEndpoint: '/logout',
   responseType: AuthorizationRequest.RESPONSE_TYPE_CODE,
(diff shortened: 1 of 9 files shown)


client/src/contexts/AuthenticationContext/AuthenticationContext.tsx
```diff
@@ -305,11 +194,63 @@ const AuthenticationProvider: React.FC<AuthenticationProviderPropsType> = ({
...
+      .catch((e) => {
+        console.error('Could not get refresh token: ', e);
+        history.push(logoutEndpoint);
+      });
   }
+  useEffect(() => {
+    // Ensure token is always fresh
+    // (function exits early if token is still valid)
+    makeRefreshTokenRequest();
+  });
```
bamader, line 213: Does this hook need a dependency array? Maybe not because the function already exits early.

client/src/contexts/AuthenticationContext/useHandlers.ts
```diff
@@ -0,0 +1,101 @@
...
+   * Once a token is retrieved, the refresh token can be used to request subsquent
+   * tokens (until the refresh token expires)
+   */
+  async function makeAuthorizationCodeTokenRequest(
+    code: string,
+    verifier: string | undefined
+  ) {
+    if (!configuration) return;
+
+    let extras: StringMap | undefined = undefined;
```
bamader, line 75: Nit: any way we could make `extras` a clearer variable name? I'm finding these auth code files harder to read than most code because of the complexity, so every little bit of clarity helps.

client/src/contexts/AuthenticationContext/useOpenIdConnectUrl.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useState } from 'react';
+import { getConfig } from '../../config/getConfig';
+
+export const useOpenIdConnectUrl = (
```
bamader, line 4: Nit: maybe put a `/** */ ` level comment above this code about what it's doing. It's hard for me personally to follow some of the logic here so explicitly spelling things out would be helpful, at least to me. May not be a problem for others though, in which case feel free to ignore.

client/src/contexts/AuthenticationContext/usePath.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect, useState } from 'react';
+import { useLocation } from 'react-router-dom';
+import { AuthenticationProviderPropsType } from './AuthenticationContext';
+
+export const usePath = (authProps: AuthenticationProviderPropsType) => {
```
bamader, line 5: Same here about an overview comment

bamader (approved):
Woof this looks like it was tough to code up. Looks good though! Few overall comments:
- Nit: not sure if it's just me, but I found some of the logic in the `AuthenticationContext` file hard to follow. That might just be because I hadn't looked at it before and it's pretty involved, but I'd probably prefer a bit more comments throughout the code about the "big picture" that's happening in each part. There's plenty of good detail-oriented comments, but to me the complexity of the file is obscuring the broader logic flow. I realize this is super subjective though so if others feel like it's not an issue, can definitely ignore. Either way, also not a blocker.
- As far as the local storage goes, I'm a little unclear about why we need to store the access key in it (there's a comment about this in line 125 of `AuthenticationContext`). I get storing the expiration time there to retrieve and determine the difference remaining, but maybe we do need to revisit putting the full key there.

Otherwise, LGTM.

jlhogan:
big :clap: for reworking one of the first complex things we did in this app! I was hacking my way thru react when i contributed to this, and it is way clearer now. One question is why we need to store a timestamp for expiry in local storage. Shouldn't we be able to trigger the forced logout whenever we recieve an "expired JWT" token response? Or is that not restrictive enough and we want to log out other times as well ?

jlhogan (approved):
approved but want to confirm: we do not expect the user to encounter the expired JWT thing anymore beause we will fetch them a fresh one if they are navigating around the site OR we will log them out using the client-side logic otherwise

mmazanec22:
@jlhogan yes I believe that should be the case!

PR #851, opened 2020-12-02
Change warning to error

### LINKED ISSUE #496: Submitting with incomplete data: When I attempt to submit my incomplete data to OEC, I am blocked and see an error message.
INcomplete - error
– Block when data is incomplete or if there are 0 records

### LINKED ISSUE #751: When I attempt to submit data to OEC that is incomplete, I am scrolled to the top of the page and see an error alert.
Instead of disabling the submit button, it is always enabled. 

When a user hits the button with errors in their roster, the warning message with the "update your roster" header turns into an error.

Hide the bottom bar when the user has no data entered. 

Submit-error

bamader:
Block submits to OEC while a user has either no records in their roster, or one or more records contains errors. Users attempting to submit with errors are scrolled to the top of the roster page, where the yellow info alert to correct records with Batch Edit becomes a red error alert. Closes #751 . Closes #496 .

Note: I left the bottom bar itself commented out because I'm assuming it's still the plan to leave that out until January.


jlhogan:
in general let's prioritize bugs over new features that are blocked til jan

client/src/containers/Roster/hooks/useChildrenWithErrorsAlert.tsx
```diff
@@ -56,7 +58,8 @@ export const useChildrenWithErrorsAlert = (
 function getChildrenWithErrorsAlertProps(
   numberOfChildrenWithErrors: number,
   numberOfWithdrawnChildrenWithErrors: number,
-  organizationId?: string
+  organizationId?: string,
+  alertType?: 'warning' | 'error'
```
jlhogan, line 62: since below it appears `warning` is the default value, can force that here like 
```
alertType?: 'warning' | 'error' = 'warning'
```
and then get rid of the ternary on line 90

jlhogan (approved):

mmazanec22 (approved):

JordanGuinn (approved):

PR #853, opened 2020-12-02
Accept confidentiality agreement

bamader:
Adds a util into the login function that lets e2e tests accept the confidentiality agreement (if they need to for a particular run). Also, it looks like a *significant* portion of the e2e tests were failing due to the login identifying the wrong title, so I changed how that got parsed to find the landing page `h1`.

### DIFF
--- a/e2e-tests/README.md
+++ b/e2e-tests/README.md
@@ -20,12 +20,68 @@ $ yarn test --tag upload
 
 ## Gotchas:
 
-- Selenium/nightwatch/browserstack (not sure which) cares, for certain functions, about whether or not the element is currently on the screen. Nightwatch docs say that it scrolls the element into view for certain functions. This seems to be a lie. The `scrollToElement` util should actually do it though.
+As an overview, the nightwatch docs are a good starting resource, but should NOT be taken as true. There are lots of instances where the docs will say one thing but the referenced function will have completely different behavior or use parameter signatures differently. If something isn't working the way the docs suggest, google around or try one of the common fixes below before scrapping the idea.
 
-- Xpath arrays start at 1-- so if more than one el will be returned, use `[1]`, not `[0]`
-
-- We have to use CommonJS modules, even though VSCode will try to tell you that you can use ES Modules
+### About Selectors
 
 - Most nightwatch functions take a selection type `xpath` or `css selector`. The option to do this on a given function is usually but not always indicated in the nightwatch docs. It is totally unclear whether this is required, though there are funcs that switch the browser to using one selector type to another... so we default here to specifying it. There are also some nightwatch functions that specifically do not take a selector even though the docs seem to indicate that they do (`browser.assert.containsText`, for example, only works when a selector was not specified and a css selector was used).
 
-- Sometimes React changes take time to propagate, like when saving form data. `browser.pause()` is your friend.
+- xpath can be finnicky with its syntax, but here are some resources that can help you identify elements you need:
+
+* https://www.swtestacademy.com/xpath-selenium/#xpath-text
+* https://devhints.io/xpath
+
+### Nits, or Death By Details
+
+- We have to use CommonJS modules, even though VSCode will try to tell you that you can use ES Modules
+
+- Xpath arrays start at 1-- so if more than one el will be returned, use `[1]`, not `[0]`
+
+- Imports from utils or other files follow a different signature than JS. They look like `const { login } = require('../utils/login');` (note the `require` rather than `import`).
+
+- Defining a test uses the following syntax: 
+```
+module.exports = {
+  '@tags': ['tag_name_here'],
+  name_of_your_test_function: async function (browser) {
+    await browser.init();
+    // TODO: Your test code here
+    browser.end();
+  },
+};
+```
+
+### Notes on Visibility
+
+- Selenium/nightwatch/browserstack (not sure which) cares, for certain functions, about whether or not the element is currently on the screen. Nightwatch docs say that it scrolls the element into view for certain functions. This seems to be a lie. The `scrollToElement` util should actually do it though.
+
+- Nightwatch's definition of `visible` isn't the same as your definition. That's because nightwatch makes a distinction between three different types of 'states' of elements on a page, but doesn't explain how they're different or when to use them. Here's what we've figured out:
+
+* `present` elements are simply those that occur in a page's HTML. They can be visible, invisible, hidden, whatever; they're just there. *All* elements should be considered `present`.
+* `visible` elements are those that are `present`, and also can be seen by a viewer of the page (i.e. no hidden elements). What it means to be seen is somewhat murky, because things like disabled buttons or non-interactable form elements (see below) don't register as `visible`. The best working definition we have for `visible` is that an element is not hidden and is also 'active' on the page.
+* `interactable` elements are those that are `present`, `visible`, and also can receive data input or be manipulated in some way. Text inputs in form fields are `interactable` elements, as are radio buttons (and yet, the checkboxes in our component library are not...). `Non-interactable` elements are also not considered `visible`. An element must be considered `interactive` by nightwatch in order for a test to `click()` on it.
+
+### Sources of Frustration and Potential Solutions
+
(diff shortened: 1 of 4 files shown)


e2e-tests/utils/login.js
```diff
@@ -10,8 +11,50 @@ module.exports = {
...
+        ];
+        await browser.waitForElementVisible(...continueButtonArgs);
+        await browser.click(...continueButtonArgs);
+        console.log('Accepted confidentiality agreement');
+      } else {
+        console.log('Confidentiality agreement already accepted');
+      }
+    });
+
+    // Check to see if we logged in and the title has been changed
```
mmazanec22, line 53: The title element of the page found in the body element should also be "Let's get started" though-- I suspect this is an issue of needing to wait longer for the title el itself to change, rather than catching the wrong element

e2e-tests/utils/scrollToElement.js
```diff
@@ -1,10 +1,16 @@
 // Nightwatch docs say that the library handles this for certain functions (like click) but it seems to not :(
 
 module.exports = {
-  scrollToElement: async function (browser, elSelectorArgs) {
+  scrollToElement: async function (
+    browser,
+    elSelectorArgs,
+    waitForVisible = true
```
mmazanec22, line 7: Why wouldn't we want to wait for an element to be visible?

bamader: So this check is because "visible" to nightwatch doesn't mean what we intuitively think it does. Elements like disabled buttons and non-interactable JSX are clearly visible to the user, but nightwatch doesn't register them as `visible` from their own standpoint. They call that "present" instead (if you look at line  18 I had to use an  `elementNotPresent` check to measure a page transition because because the login button becomes `not visible` immediately after you click it and it disables).  And, because the input checkbox gets seen by nightwatch as "non-interactable" (see my comment on line 34), it doesn't get considered as a visible element (I think it's horrible weirdness and poor definition documenting on their part but it is what it is). So this conditional check was a way for  me to use the working code of getting a location and scrolling the screen to that point without duplication and without having the final `waitForElementVisible` check because that was causing the test to fail. As I'm writing this out, I'm thinking this may be worth sticking in a comment on the `scrollToElement` function in case we write more e2e tests in the future, because it def confused me a bunch this afternoon.

e2e-tests/utils/login.js
bamader: Is it true that the title in the body element should be the same? We never explicitly set the `title` attribute for either `body` or `main`, so nightwatch just starts at the root of whatever element we give it and then finds the highest-level string (or at least that's what their docs say). The only difference between the two elements is that `body` includes the header at the top with the ECE Reporter logo, which is I think why it gets  parsed as the title since  it occurs first, and main doesn't, so the highest-level thing there is the "Let's get started."

bamader: An easier way for us to write this check may actually just be to search for an `h1` with the "Let's get started" text, since we know the getting started landing page is the only one in the app that has that in an h1. When I was working with changeChild I was seeing a lot of flakey behavior from `assert` (which is why I changed the syntax there to use `expect`).

e2e-tests/utils/scrollToElement.js
mmazanec22: Yeah that sounds like a weirdness that took a lot of figuring out!  I think it's worth a note in this file and a long note in the "gotchas" section of the e2e readme

e2e-tests/utils/login.js
mmazanec22: The title of the page and the h1 should always be the same.  We have a hook that is running on every page to do that.  It's super obnoxious that the function doesn't do what it says... oh nightwatch.  I would agree that finding the h1 is fine.  We should probably write a "h1 text matches" helper func that ensures that a) the h1 is what we want it to be b) there is only one h1 on the page and c) the h1 and title match.  We can make a ticket for that though.

e2e-tests/utils/scrollToElement.js
bamader: I just put some serious BITE into the readme

e2e-tests/README.md
```diff
@@ -20,12 +20,68 @@ $ yarn test --tag upload
...
+* https://devhints.io/xpath
+
+### Nits, or Death By Details
 
 - We have to use CommonJS modules, even though VSCode will try to tell you that you can use ES Modules
 
-- Most nightwatch functions take a selection type `xpath` or `css selector`. The option to do this on a given function is usually but not always indicated in the nightwatch docs. It is totally unclear whether this is required, though there are funcs that switch the browser to using one selector type to another... so we default here to specifying it. There are also some nightwatch functions that specifically do not take a selector even though the docs seem to indicate that they do (`browser.assert.containsText`, for example, only works when a selector was not specified and a css selector was used).
+- Xpath arrays start at 1-- so if more than one el will be returned, use `[1]`, not `[0]`
+
+- Imports from utils or other files follow a different signature than JS. They look like `const { login } = require('../utils/login');` (note the `require` rather than `import`).
```
mmazanec22, line 40: This is what the note about commonjs modules means-- the two should prob be combined

e2e-tests/README.md
```diff
@@ -20,12 +20,68 @@ $ yarn test --tag upload
...
+  },
+};
+```
+
+### Notes on Visibility
 
-- Sometimes React changes take time to propagate, like when saving form data. `browser.pause()` is your friend.
+- Selenium/nightwatch/browserstack (not sure which) cares, for certain functions, about whether or not the element is currently on the screen. Nightwatch docs say that it scrolls the element into view for certain functions. This seems to be a lie. The `scrollToElement` util should actually do it though.
+
+- Nightwatch's definition of `visible` isn't the same as your definition. That's because nightwatch makes a distinction between three different types of 'states' of elements on a page, but doesn't explain how they're different or when to use them. Here's what we've figured out:
```
mmazanec22, line 58: Lol

e2e-tests/utils/login.js
```diff
@@ -10,8 +11,50 @@ module.exports = {
...
+    await browser.element(...confidentialityBoxArgs, async (result) => {
+      if (result.state === 'success') {
+        console.log('Need to accept agreement');
+        await scrollToElement(browser, confidentialityBoxArgs, false);
+
+        // For reasons I do not understand because selenium, the
+        // checkbox on the agreement page gets interpreted as
+        // "non-interactable", so need to use direct JS injection
+        // to click, like in scrollToElement
+        await browser.execute(
```
mmazanec22, line 38: I figured out what this was!  It's bc USWDS hides the actual checkbox off the screen, and replaces it with a ::before value.  You can click the label element instead (css selector `label[for=element-id]`)

mmazanec22: ...ergo the checkbox itself is never visible, and nightwatch is like "I couldn't possibly understand that things are hidden off screen"

bamader: I KNEW IT WAS THE LABEL. I tried messing with that first and was like why is this text messing with things so bad. And then I learned about `execute` from stackoverflow and was like thank you  for ending my suffering kind stranger 😭

bamader: Lol this readme is a catalogue of "we expect nightwatch to be able to gracefully handle this very normal thing. We have since discovered it cannot."

bamader: Also gonnna put this note about the checkbox and off-screen elements in the readme...

e2e-tests/README.md
mmazanec22: (Also, thank you for writing this much clearer explanation of what that means in actuality for writing tests!)

jlhogan (approved):
given how legit their landing page looks and the 177,539 weekly downloads on npm the failings of nightwatch are SURPRISING. we should be generous samaritans  and write a blog post... .one day.....

bamader:
Yeah, it kinda makes me sad that stuff doesn't perform the way it should. I feel like if they just corrected their docs or added clearer parameter signatures it would solve like 70-80% of the problems. Alas, earwax 👂

e2e-tests/utils/login.js
bamader: Wound up writing that util helper for this PR, actually (commit incoming). Once I realized how language chaining for the `expect` call works it was much easier than I thought!

mmazanec22 (approved):

PR #864, opened 2020-12-03
Fix header error

### LINKED ISSUE #835: download example bug
```
2020-11-25T10:22:01.945-05:00	Nov 25 15:22:01 ip-10-0-40-22 web: Unable to download example Error [ERR_HTTP_HEADERS_SENT]: Cannot remove headers after they are sent to the client
	2020-11-25T10:22:01.945-05:00	Nov 25 15:22:01 ip-10-0-40-22 web: at ServerResponse.removeHeader (_http_outgoing.js:578:11)
	2020-11-25T10:22:01.945-05:00	Nov 25 15:22:01 ip-10-0-40-22 web: at ServerResponse.send (/var/app/current/node_modules/express/lib/response.js:210:10)
	2020-11-25T10:22:01.945-05:00	Nov 25 15:22:01 ip-10-0-40-22 web: at /var/app/current/dist/src/routes/template.js:66:13
	2020-11-25T10:22:01.945-05:00	Nov 25 15:22:01 ip-10-0-40-22 web: at Generator.next (<anonymous>)
	2020-11-25T10:22:01.945-05:00	Nov 25 15:22:01 ip-10-0-40-22 web: at fulfilled (/var/app/current/dist/src/routes/template.js:5:58)
	2020-11-25T10:22:01.945-05:00	Nov 25 15:22:01 ip-10-0-40-22 web: at runMicrotasks (<anonymous>)
	2020-11-25T10:22:01.945-05:00	Nov 25 15:22:01 ip-10-0-40-22 web: at processTicksAndRejections (internal/process/task_queues.js:97:5) {
	2020-11-25T10:22:01.945-05:00	Nov 25 15:22:01 ip-10-0-40-22 web: code: 'ERR_HTTP_HEADERS_SENT'
	2020-11-25T10:22:01.945-05:00	Nov 25 15:22:01 ip-10-0-40-22 web: }
```

bamader:
Issue was caused by having two `res.send()` calls, resulting in the system trying to unexpectedly send a response after one response was already formatted and written (`streamTabularData` sets the content type and sends, and then the route was trying to send again with an empty body because `streamTabularData` doesn't return anything, which created the `cannot Remove headers` bug). Closes #835 .

### DIFF
--- a/src/routes/template.ts
+++ b/src/routes/template.ts
@@ -69,9 +69,7 @@ templateRouter.get(
       } else if (whichFakeChildren === 'missingOne') {
         fakeChildren = childrenAllMissingOneField;
       }
-      res.send(
-        await streamUploadedChildren(res, fakeChildren, fileType as BookType)
-      );
+      await streamUploadedChildren(res, fakeChildren, fileType as BookType);
     } catch (err) {
       console.error('Unable to download example', err);
       throw new InternalServerError('Could not download example file: ' + err);


jlhogan (approved):

src/routes/template.ts
```diff
@@ -69,9 +69,7 @@ templateRouter.get(
       } else if (whichFakeChildren === 'missingOne') {
         fakeChildren = childrenAllMissingOneField;
       }
-      res.send(
-        await streamUploadedChildren(res, fakeChildren, fileType as BookType)
-      );
+      await streamUploadedChildren(res, fakeChildren, fileType as BookType);
```
jlhogan, line 72: i wonder if this await is necessary ?

bamader: I _think_ it is (but could be wrong). Was doing some reading and saw that one of the biggest causes of response header errors were asynchronous functions trying to invoke `res.send()` before the response body was basically ready. Granted, this line is the last in its block, so it might not be an issue if there's nothing getting called after it.

jlhogan: i think the double header issue was def  a result of the `res.send` on top of the piping already happening, not an async issue. but doesn't really matter either way, was just postulating :)

mmazanec22 (approved):

PR #865, opened 2020-12-03
Api curl tests

jlhogan:


README.md
```diff
@@ -75,8 +75,28 @@ $ docker-compose exec client yarn test
...
+    To run these tests:
+    ```
+    $ cd client
+    $ yarn test-e2e
+
+    ```
+    Set `TEST_API_PATH` env var to run them against an environment other than the default `http://localhost:5001`
+
+2. **End-to-end Tests**: these [nightwatch](https://nightwatchjs.org/)-powered selemnium tests run in Browserstack. Thus, they can only be run against a deployed stack (for now, there are ways to set up Browserstack for local apps but we haven't done that). They live [here](e2e-tests), and there's more info about them [here](e2e-tests/README.md)
+
```
bamader, line 92: They have a nice desktop app for local testing but _oh my god_ trying to test a localhost with internet explorer is so slow

client/src/integrationTests/children.test.ts
```diff
@@ -0,0 +1,117 @@
...
+
+const TEST_OPTS: ApiOpts = {
+  headers: { 'X-Test-No-Authenticate': 'true' },
+  accessToken: '',
+};
+
+describe('integration', () => {
+  describe('api', () => {
+    beforeAll(() => {
+      disableFetchMocks();
```
bamader, line 18: What does this function do? Does this mean we're not mocking any of the api calls and instead are making actual calls?

bamader (approved):

mmazanec22 (approved):

client/src/integrationTests/children.test.ts
jlhogan: in `setupTests`, we call `enableFetchMocks` which does a bunch of stuff to mock `fetch`. Here, we want to actually do the real fetch-ing of real remote (or local, if you're running these locally) resources so we temporarily disable them for just these tests

README.md
jlhogan: :')

Issue #863, opened 2020-12-03
Enrollment and funding updates (remove unfunded)


RyanHansz:
- Remove unfunded option from funding types in add record flow and when changing enrollments in a child record
- Add the following alert at the top of the section:
<img width="867" alt="Screen Shot 2020-12-03 at 2 18 52 PM" src="https://user-images.githubusercontent.com/63253539/101077203-85aa3380-3572-11eb-9d9c-445cbe897bcd.png">
"Adding a child who has withdrawn, changed sites or age groups? Add their earliest enrollment then make changes from the child’s record in your roster."

mmazanec22:
Addressed in #866

PR #866, opened 2020-12-03
859 add funding bug

### LINKED ISSUE #859: Cannot save enrollment and funding to add child

mmazanec22:
Closes #859 and gets us to where we can at least add a child.  I made a few todo notes (that I will make into tickets) about some other bugs I ran into.

### DIFF
--- a/client/src/components/Forms/Enrollment/Fields/EnrollmentStartDate.tsx
+++ b/client/src/components/Forms/Enrollment/Fields/EnrollmentStartDate.tsx
@@ -23,12 +23,14 @@ export const EnrollmentStartDateField = <
 }: EnrollmentStartDateProps<T>) => {
   return (
     <FormField<T, DateInputProps, Moment | null>
-      getValue={(data) => enrollmentAccessor(data).at('entry')}
-      parseOnChangeEvent={(e: any) => e}
+      getValue={(data) => {
+        console.log(data)
+        return enrollmentAccessor(data).at('entry')
+      }}
+      parseOnChangeEvent={(e) => { console.log(e); return e as unknown as Moment }}
       inputComponent={DateInput}
       label="Enrollment start date"
       id="start-date"
-      defaultValue={null}
       status={(data, _, props) =>
         getValidationStatusForField(
           enrollmentAccessor(data),
--- a/client/src/components/Forms/Enrollment/Fields/EnrollmentStartDate.tsx
+++ b/client/src/components/Forms/Enrollment/Fields/EnrollmentStartDate.tsx
@@ -21,13 +21,18 @@ export const EnrollmentStartDateField = <
 >({
   enrollmentAccessor = (data) => data as TObjectDriller<Enrollment>,
 }: EnrollmentStartDateProps<T>) => {
+  // TODO: why does this not show up after save during create child?
+  // To reproduce: add child, save without completely filling out enrollment form (but include entry)
   return (
     <FormField<T, DateInputProps, Moment | null>
       getValue={(data) => {
-        console.log(data)
-        return enrollmentAccessor(data).at('entry')
+        console.log(data);
+        return enrollmentAccessor(data).at('entry');
       }}
-      parseOnChangeEvent={(e) => { console.log(e); return e as unknown as Moment }}
+      parseOnChangeEvent={(e) => {
+        console.log(e);
+        return (e as unknown) as Moment;
+      }}
       inputComponent={DateInput}
       label="Enrollment start date"
       id="start-date"
--- a/client/src/components/Forms/Enrollment/Fields/NewFunding.tsx
+++ b/client/src/components/Forms/Enrollment/Fields/NewFunding.tsx
@@ -18,8 +18,6 @@ import { stringify } from 'querystring';
 import { useAuthenticatedSWR } from '../../../../hooks/useAuthenticatedSWR';
 import { getValidationStatusForFields } from '../../../../utils/getValidationStatus';
 
-const UNFUNDED = 'Unfunded';
-
 type FundingFieldProps<T> = {
   fundingAccessor?: (_: TObjectDriller<T>) => TObjectDriller<Funding>;
   getEnrollment: (_: TObjectDriller<T>) => Enrollment;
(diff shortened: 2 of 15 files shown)


client/src/components/Forms/Enrollment/Fields/NewFunding.tsx
```diff
@@ -61,6 +63,51 @@ export const NewFundingField = <
...
+      ),
+    };
+  });
+
+  if (isEdit) {
+    options.splice(0, 0, {
+      text: UNFUNDED,
+      id: UNFUNDED,
+      value: UNFUNDED,
+      onChange: () => {},
```
jlhogan, line 107: this doesn't appear to actually do anything, so is it worth adding?

client/src/components/Forms/Enrollment/Fields/Site.tsx
```diff
@@ -30,6 +30,7 @@ export const SiteField = <T extends Enrollment | ChangeEnrollment>({
           e.target.value || null,
         value: `${site.id}`,
         id: `site-${site.id}`,
+        preprocessForDisplay: (data) => data === site.id,
```
jlhogan, line 33: what does this do?

client/src/containers/CreateRecord/CreateRecord.tsx
```diff
@@ -68,6 +68,7 @@ const CreateRecord: React.FC = () => {
             },
           ],
         });
+        return;
```
jlhogan, line 71: why is this return necessary? If i'm correctly matching my parens, the only other stuff in this func is in the else

jlhogan:
Thinking about this overnight, and my new request is instead of making the enrollment update cascade to funding, could we make the POST enrollment endpoint explicitly check for `fundings` object on the `enrollment` post body, and creating any fundings via separate db executions if they exist ?

client/src/components/Forms/Enrollment/Fields/NewFunding.tsx
mmazanec22: It's a required prop.  I would argue that it should be preserved in the component lib bc this is a pretty unusual use case.

client/src/components/Forms/Enrollment/Fields/Site.tsx
mmazanec22: Radio buttons and checkboxes expect a true or false value for display-- the "checked"v alue

client/src/containers/CreateRecord/CreateRecord.tsx
mmazanec22: Cruft from being totally bewildered about a bug!

mmazanec22:
What if a user tries to save without funding info, and then enters it and tries to save again?

client/src/components/Forms/Enrollment/Fields/NewFunding.tsx
jlhogan: oh i mean the whole option, not specifically the onChange prop. Maybe more of a design q but if choosing unfunded doesn't delete fundings (which i dont think this does?) then we shoulldn't let people think it will. To remove a funding, they'd hae to delete the funding directly

RyanHansz: #863 removes the unfunded option

RyanHansz: #863 removes the unfunded option

client/src/components/Forms/Enrollment/Fields/EnrollmentStartDate.tsx
```diff
@@ -21,14 +21,21 @@ export const EnrollmentStartDateField = <
...
   enrollmentAccessor = (data) => data as TObjectDriller<Enrollment>,
 }: EnrollmentStartDateProps<T>) => {
+  // TODO: why does this not show up after save during create child?
+  // To reproduce: add child, save without completely filling out enrollment form (but include entry)
   return (
     <FormField<T, DateInputProps, Moment | null>
-      getValue={(data) => enrollmentAccessor(data).at('entry')}
-      parseOnChangeEvent={(e: any) => e}
+      getValue={(data) => {
+        console.log(data);
```
jlhogan, line 29: nit: console.logs

client/src/components/Forms/Enrollment/Fields/EnrollmentStartDate.tsx
```diff
@@ -21,14 +21,21 @@ export const EnrollmentStartDateField = <
...
+        return enrollmentAccessor(data).at('entry');
+      }}
+      parseOnChangeEvent={(e) => {
+        console.log(e);
```
jlhogan, line 33: and here

client/src/components/Forms/FamilyIncome/Form.tsx
```diff
@@ -108,6 +108,8 @@ export const FamilyIncomeForm: React.FC<FamilyIncomeFormProps> = ({
     child?.family?.incomeDeterminations?.find(
       (d) => d.id === incomeDeterminationId
     ) || ({} as IncomeDetermination);
+  // TODO: if we're in the create record flow, just find the income det that exists
```
jlhogan, line 111: this existed at some point, and brandon merged a change to get rid of it b/c it was causing smething weird (presumably in the edit flow) so i think we'll need to distinguish between edit/create in this form

jlhogan (approved):

client/src/components/Forms/FamilyIncome/Form.tsx
bamader: Yeah Ryan wanted creating new dets to not prefill with the information from the existing det

bamader (approved):

client/src/components/Forms/FamilyIncome/Form.tsx
jlhogan: mhm mhm, so here if we're in create flow then look for existing det, otherwise empty det

bamader: Other way around, right? Create uses a blank and edit searches for existing?

PR #872, opened 2020-12-04
Conditional login test and other stability fixes

bamader:
Makes `login` util accept conditional parameter that specifies whether it's okay if we're already logged in during a test (this makes retries not totally useless). For all tests except login, this flag is true because we're not testing the login functionality. For the login test, this flag is false because the test must  be able to actually _do_ the log in. Also does a bit of refactoring to try to help keep our tests up to date (i.e. Upload was looking for the wrong number of sites).


e2e-tests/tests/changeChild.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
   '@tags': ['child', 'change'],
   changeChild: async function (browser) {
     await browser.init();
-    await login(browser);
```
mmazanec22, line 13: Maybe it should default to true and the flag can be optional?

e2e-tests/utils/acceptModal.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  acceptModal: async function (browser, buttonText, noModalText) {
```
mmazanec22, line 2: This is a nice util!

e2e-tests/utils/login.js
```diff
@@ -1,11 +1,19 @@
 const { scrollToElement } = require('../utils/scrollToElement');
 const { headerMatch } = require('../utils/headerMatch');
 module.exports = {
-  login: async function (browser) {
+  login: async function (browser, okayIfAlreadyLoggedIn) {
```
mmazanec22, line 4: This variable name is super clear

e2e-tests/utils/login.js
```diff
@@ -1,11 +1,19 @@
...
     // Click login from ECE Reporter
     await browser.click('xpath', "//*/a[contains(@href,'/login')]");
     await browser.waitForElementVisible('css selector', 'body');
-    browser.assert.title('IdentityServer4');
+    await browser.title(function (result) {
+      if (!okayIfAlreadyLoggedIn) {
+        browser.assert.title('IdentityServer4');
+      }
+      // If it's okay to be logged in and we are, test is done
+      else if (result.valule != 'IdentityServer4') {
```
mmazanec22, line 13: `valule` typo.  Also maybe we should check here for what the value should be instead?

bamader: Lol I don't know why but it reads like 'ukulele' to me 😂 . Yeah we can check for that instead, since already being logged in means we won't hit the confidentiality agreement and will always go straight to Getting Started.

mmazanec22 (approved):

mmazanec22 (approved):

e2e-tests/tests/upload.js
```diff
@@ -8,6 +9,10 @@ module.exports = {
   // uploadWrongFormat: async function (browser) {},
   // uploadMissingInfo: async function (browser) {},
   uploadCompleteInfo: async function (browser) {
+    // TODO: MODIFY THIS IF NUMBER OF SITES ON TEST USER CHANGES
+    // (Previously, we were checking for four, but then we switched
+    // something and went up to 8 but we didnn't change that in the test)
+    const EXPCTED_NUMBER_OF_SITES = 8;
```
jlhogan, line 15: this is fine, but i don't know that we actually need/want to be relying on this specific value. Could shorten the contains to just `100 children`  or something? but not a blocker

e2e-tests/tests/downloadFiles.js
```diff
@@ -13,9 +13,19 @@ async function testDownload(browser, opts) {
 
 const { templateExcel, templateCSV, exampleExcel, exampleCSV } = downloadOpts;
 
+/**
+ * NOTE: We skip this test right now because nightwatch doesn't allow
```
jlhogan, line 17: delete this file. version control means we have this around in old commits if we need it.

e2e-tests/tests/downloadExcelExample.js
```diff
@@ -0,0 +1,21 @@
...
+} = require('../utils/downloadTemplateOrExample');
+
+module.exports = {
+  // Need to skip this in edge and ie because testing download functionality with browserstack doesn't work in those browsers
+  // https://www.browserstack.com/docs/automate/selenium/test-file-download
+  '@tags': ['download-excel-example', 'skip-edge'],
+  downloadExcelExample: async function (browser) {
+    opts = {
+      buttonText: buttonTexts.exampleExcel,
+      expect: true,
```
jlhogan, line 14: do we still need expect? appears we're getting rid of the test case that looks for a nonexistant file (which i think makes sense, wasn't really conferring any additoinal confidence). So b/c we now always expect the file to exist we can remove this option and simplify the `downloadTemplateOrExample` util

jlhogan (approved):

PR #888, opened 2020-12-08
Home page updates

### LINKED ISSUE #871: Data definition for race
On the data definitions, For the race, below required, change the text to:
"At least one race option required"

Screen Shot 2020-12-04 at 11 03 30 AM

### LINKED ISSUE #873: Home page updates (pre-complete data)
When I log into ECE Reporter I see:
[This design](https://share.goabstract.com/42315efe-ac59-43d4-aea1-f47b0cbf4428?collectionLayerId=f14fea62-ba34-4153-9d2b-1885f9f3b99b&mode=design)

Screen Shot 2020-12-04 at 2 47 18 PM

- I can see the steps to upload data in ECE Reporter
- I can see the sites I manage in ECE Reporter
- I can see the how-to guide at: https://ece-reporter.documize.com/s/burrnubmbdja9sqnbh6g/ece-reporter-help/d/bv34kdhmdesjli7los1g/how-to-prepare-your-data-for-ece-reporter

### LINKED ISSUE #885: Batch Upload doesn't parse 'Y' and 'N' for undefinable booleans
There are a number of fields in the template that use undefinable booleans but allow users to enter in more than just the three ways Yes/No/Unknown (i.e. fields like dualLanguageLearner, hispanicOrLatinxEthnicity, foster, receivesDisabilityServices, etc. all have Yes, Y, No, N as valid inputs in the sheet instructions). Rows that are uploaded with anything but Yes or No in these columns are not parsed or saved as children to the roster.
* Just needs a `mapEnum` to be applied to each field (since UndefinableBoolean is an enum now) before instantiating the Child object and trying to save it to the DB

bamader:
Addresses some tweaks to the home page and race field definitions. Also fixes parsing of 'Y' and 'N' as valid options for fields stored as undefinable booleans.

Closes #885 
Closes #873 
Closes #871

### DIFF
--- a/client/src/containers/GettingStarted/GettingStarted.tsx
+++ b/client/src/containers/GettingStarted/GettingStarted.tsx
@@ -1,10 +1,5 @@
 import React, { useContext } from 'react';
-import {
-  Button,
-  Alert,
-  StepList,
-  TextWithIcon,
-} from '@ctoec/component-library';
+import { TextWithIcon } from '@ctoec/component-library';
 import { Link } from 'react-router-dom';
 import { getH1RefForTitle } from '../../utils/getH1RefForTitle';
 import { ReactComponent as Arrow } from '@ctoec/component-library/dist/assets/images/arrowRight.svg';
@@ -131,7 +126,6 @@ const GettingStarted: React.FC = () => {
           </Link>
         </p>
       </div>
-      {/* <div style={{width: '1em', height: 372, background: '#dcdee0'}}></div> */}
     </div>
   );
 };
--- a/client/src/containers/GettingStarted/GettingStarted.tsx
+++ b/client/src/containers/GettingStarted/GettingStarted.tsx
@@ -1,88 +1,137 @@
-import React from 'react';
-import { Button, Alert } from '@ctoec/component-library';
+import React, { useContext } from 'react';
+import {
+  Button,
+  Alert,
+  StepList,
+  TextWithIcon,
+} from '@ctoec/component-library';
 import { Link } from 'react-router-dom';
 import { getH1RefForTitle } from '../../utils/getH1RefForTitle';
-import { CSVExcelDownloadButton } from '../../components/CSVExcelDownloadButton';
+import { ReactComponent as Arrow } from '@ctoec/component-library/dist/assets/images/arrowRight.svg';
 import { ReactComponent as Image } from '../../images/PersonWithSpreadsheet.svg';
+import UserContext from '../../contexts/UserContext/UserContext';
 
 const GettingStarted: React.FC = () => {
+  const { user } = useContext(UserContext);
   const h1Ref = getH1RefForTitle();
+  const org = (user?.organizations || [])[0] || {};
+
   return (
     <div className="grid-container margin-top-4">
       <div className="grid-row grid-gap">
         <div className="tablet:grid-col-8">
-          <h1 ref={h1Ref}>Let's get started</h1>
+          <h1 ref={h1Ref}>Hello {user?.firstName}!</h1>
           <p>
-            This tool will allow you to share your state-funded enrollment data
-            with OEC.
+            ECE Reporter will allow you to share your state-funded enrollment
+            data with the Office of Early Childhood (OEC). OEC uses this data to
+            make informed program and policy decisions affecting young children
+            and families.
           </p>
-          <h2>Here's what to include</h2>
-          <h3>Enrollments in spaces funded by</h3>
+          <h2>Sites you manage at {org.providerName}</h2>
           <ul>
-            <li>Child Day Care (CDC)</li>
-            <li>Priority or Competitive School Readiness (PSR and CSR)</li>
-            <li>Smart Start (SS)</li>
-            <li>State Head Start (SHS)</li>
+            {(user?.sites || []).map((site) => (
+              <li>{site.siteName}</li>
+            ))}
           </ul>
-          <Alert
-            heading="Learn more about required data"
-            type="info"
-            text={
-              <span>
-                See all required data fields, definitions and rationale in{' '}
-                <Link
-                  className="usa-button usa-button--unstyled"
-                  to="/data-requirements"
-                >
-                  OEC's enrollment data requirements
-                </Link>
-              </span>
-            }
-          />
-          <h3>Data collection period</h3>
+          <h2>How to use ECE Reporter</h2>
           <p>
-            You must submit your data to OEC by 03/05/2021. The data collection
-            period begins 07/01/2020 and continues through at least 01/01/2021.
+            These instructions guide you through the file upload process to add
+            many records at once. If you'd like to add records manually,
+            <Link to="/roster/"> you can add them on your roster.</Link>
           </p>
-          <div>
-            <h2>Download the latest batch upload template</h2>
-            <CSVExcelDownloadButton
-              fileType="xlsx"
-              whichDownload="template"
-              className="margin-bottom-3"
-            />
-            <CSVExcelDownloadButton
-              fileType="csv"
-              whichDownload="template"
-              className="margin-bottom-3"
-            />
-          </div>
-          <div className="display-block">
-            <p>You can see a completed example template here:</p>
-            <CSVExcelDownloadButton
-              fileType="xlsx"
-              whichDownload="example"
-              className="margin-bottom-3"
-            />
-            <CSVExcelDownloadButton
-              fileType="csv"
-              whichDownload="example"
-              className="margin-bottom-3"
-            />
-          </div>
         </div>
         <div className="tablet:grid-col-4" role="presentation">
           <Image />
         </div>
       </div>
-      <div className="grid-row">
-        <Button
-          text="Go to file upload"
-          href="/upload"
-          className="margin-top-2"
-        />
-        <Button text="Go to roster" href="/roster" className="margin-top-2" />
+      <h3 className="display-flex font-body-lg height-5 line-height-body-6 margin-y-0">
+        <div className="text-white text-bold text-center bg-primary width-5 radius-pill margin-right-1">
+          1
+        </div>
+        Review the data requirements
+      </h3>
+      <div className="margin-left-6">
+        <p>
+          Learn more about the data elements ECE Reporter collects for each
+          enrollment.
+        </p>
+        <p>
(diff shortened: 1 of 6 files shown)


client/src/components/Forms/ChildInfo/Fields/Race.tsx
```diff
@@ -37,30 +43,50 @@ const raceOptions: { label: string; field: RaceField }[] = [
   },
 ];
 
+type RaceFieldProps = {
+  child: Child;
```
jlhogan, line 47: you shouldn't need to pass this in -- it should be accessible as `data` from the formContext, or via the `dataDriller.value` from form context

client/src/containers/GettingStarted/__snapshots__/GettingStarted.test.tsx.snap
```diff
@@ -12,122 +12,26 @@ exports[`GettingStarted matches snapshot 1`] = `
         class="tablet:grid-col-8"
       >
         <h1>
-          Let's get started
+          Hello !
```
jlhogan, line 15: should we update this test to be wrapped in a usercontext that returns a user ? so we get checks that we're correctly displaying name, org, sites moving forward

bamader: Yeah that sounds like a good idea. Incoming now.

client/src/components/Forms/ChildInfo/Fields/Race.tsx
```diff
@@ -70,20 +96,43 @@ export const RaceField: React.FC = () => {
...
+              updateData(
+                set(
+                  child,
+                  dataDriller.at(o.field).path,
+                  e.target.checked ? false : null
+                )
+              );
+            } else {
+              updateData(
+                set(child, dataDriller.at(o.field).path, e.target.checked)
```
mmazanec22, line 125: For [the birth cert bug](https://github.com/ctoec/data-collection/pull/890/files) I had to use produce in addition to set, the way the component library does, otherwise all of the other values were wiped from the child.  It didn't really make sense to me but it did fix the issue.  IDK if that will be necessary here.

client/src/components/Forms/ChildInfo/Form.tsx
```diff
@@ -94,7 +94,7 @@ export const ChildInfoForm = ({
         'hispanixOrLatinxEthnicity',
       ]) && (
         <>
-          <RaceField />
+          <RaceField child={child} />
```
mmazanec22, line 97: Should no longer be necessary to pass this

client/src/components/Forms/ChildInfo/Fields/Race.tsx
```diff
@@ -37,30 +43,50 @@ const raceOptions: { label: string; field: RaceField }[] = [
...
+ * Need to use a FormFieldSet that contains checkboxes based
+ * on a state variable instead of having a group of checkboxes,
+ * because we want the value of one of the checkboxes to control
+ * the state and checked status of other checkboxes. See BirthState
+ * and BirthTown form fields for a similar data pattern. An
+ * accessor is still used for safely manipulating each field in
+ * the child object.
  */
-export const RaceField: React.FC = () => {
+export const RaceField: React.FC<RaceFieldProps> = ({ child }) => {
```
mmazanec22, line 60: You can also get child from the form context

client/src/components/Forms/ChildInfo/Fields/Race.tsx
jlhogan: https://lodash.com/docs/#set -- set updates the actual object. Produce gives you back a copy of the obejct with the update applied (doesn't mutate the original).

client/src/containers/GettingStarted/GettingStarted.tsx
```diff
@@ -1,87 +1,130 @@
...
-          <h2>Here's what to include</h2>
-          <h3>Enrollments in spaces funded by</h3>
+          <h2>Sites you manage at {org.providerName}</h2>
           <ul>
-            <li>Child Day Care (CDC)</li>
-            <li>Priority or Competitive School Readiness (PSR and CSR)</li>
-            <li>Smart Start (SS)</li>
-            <li>State Head Start (SHS)</li>
+            {(user?.sites || []).map((site) => (
+              <li>{site.siteName}</li>
```
mmazanec22, line 28: This needs a key or React will give us console warnings

client/src/containers/GettingStarted/GettingStarted.tsx
```diff
@@ -1,87 +1,130 @@
...
+  const { user } = useContext(UserContext);
   const h1Ref = getH1RefForTitle();
+  const org = (user?.organizations || [])[0] || {};
+
   return (
     <div className="grid-container margin-top-4">
       <div className="grid-row grid-gap">
         <div className="tablet:grid-col-8">
-          <h1 ref={h1Ref}>Let's get started</h1>
+          <h1 ref={h1Ref}>Hello {user?.firstName}!</h1>
```
mmazanec22, line 18: This change will break e2e tests

client/src/components/Forms/ChildInfo/Fields/Race.tsx
mmazanec22: I wonder if the updateData that we get from the form context should just handle that stuff for us, and we can just pass in the path and value @jlhogan

jlhogan: yeah, definitely definitely a good idea

bamader: I actually tried to switch to produce when I saw that your PR used it, and for whatever reason the behavior of unchecking other race options when clicking `notDisclosed` went away. Just tested it and using set without produce for me doesn't clear any of the other data.

bamader: Also I don't think I understand--what's the idea here for how to change the race field? Or are y'all talking about a change to something in the component library (maybe a separate PR)?

mmazanec22: Yeah, separate PR for component library change.  For set and produce: about when you have entered other data in this section, and then go back and click on some of the race options?  That's the bug I was having with birth cert.  If that doesn't happen then it's probably fine the way it is!

client/src/containers/GettingStarted/GettingStarted.tsx
bamader: Updated the login util to check for the new header content (assuming our test user's name is still Voldemort)

client/src/components/Forms/ChildInfo/Fields/Race.tsx
bamader: Mmm good call, checking things in a different order. Apparently when I select other options _first_ and then go back and change the race, the de-selecting behavior also disappears. So there's some other weirdness going on here that requires further investigation.

client/src/containers/GettingStarted/__snapshots__/GettingStarted.test.tsx.snap
```diff
@@ -12,15 +12,25 @@ exports[`GettingStarted matches snapshot 1`] = `
...
         <h1>
-          Hello !
+          Hello Voldy!
         </h1>
         <p>
           ECE Reporter will allow you to share your state-funded enrollment data with the Office of Early Childhood (OEC). OEC uses this data to make informed program and policy decisions affecting young children and families.
         </p>
         <h2>
-          Sites you manage at 
+          Sites you manage at Mort School For Dark-Inclined Wizards
```
mmazanec22, line 21: Lol

mmazanec22 (approved):

client/src/containers/GettingStarted/GettingStarted.tsx
```diff
@@ -1,87 +1,130 @@
...
 import { getH1RefForTitle } from '../../utils/getH1RefForTitle';
-import { CSVExcelDownloadButton } from '../../components/CSVExcelDownloadButton';
+import { ReactComponent as Arrow } from '@ctoec/component-library/dist/assets/images/arrowRight.svg';
 import { ReactComponent as Image } from '../../images/PersonWithSpreadsheet.svg';
+import UserContext from '../../contexts/UserContext/UserContext';
 
 const GettingStarted: React.FC = () => {
+  const { user } = useContext(UserContext);
   const h1Ref = getH1RefForTitle();
+  const org = (user?.organizations || [])[0] || {};
```
jlhogan, line 12: this should probably not assume single-org, b/c donna will want to see all sites at all orgs, and we may need to come up with different display solution for multi-org lest the list get very long bu ti think they should just be a separate ticket

src/controllers/enrollmentReports/mapUtils/enum.ts
```diff
@@ -27,12 +28,20 @@ export const mapEnum = <T>(
...
+    | typeof UndefinableBoolean,
   value: string | undefined,
   opts: {
     isFundingSource?: boolean;
+    isUndefineableBoolean?: boolean;
   } = {}
 ) => {
+  // Handle special case of blanks in undefineable boolean columns
+  // right upfront
+  if ((!value || value === '') && opts.isUndefineableBoolean) {
```
jlhogan, line 41: empty string evaluates to falsy, so this `||` is unnecessary !

jlhogan (approved):

client/src/containers/GettingStarted/GettingStarted.tsx
bamader: I'll open a ticket for it in the backlog

jlhogan:
also, style-wise to match comps we'd want to:
- increase the spacing a little bit in the list of sites (can add `line-height-body-4` class to the li elements)
- add `:` after org name in the list title

and then from an accesibility standpoint, i dont know if this matters but just looking at the html and such i think all the new content you added could be inside the grid-col-8 div that includes the greeting and site list content (so tha tall the headers are together in the same heirarchy? not ssure it matters but it might?)

bamader:
@jlhogan Did you want me to push this to QA for styling review?

jlhogan (approved):

jlhogan:
@bamader nah! all good. Can run it locally :)

PR #889, opened 2020-12-08
Fix seemingly transient enrollment failure issue

### LINKED ISSUE #870: Unable to change enrollment in a child record
In staging:
Screen Shot 2020-12-04 at 7 25 16 AM

JordanGuinn:
Closes #870, as well as the issue with the last section of the create child workflow not working.

When the `changeEnrollment` route is invoked, the site ID supplied with the new enrollment was previously provided as a string (even though the actual Site id DB column is an auto-incrementing int).  This normally wouldn't be all that much of an issue but, _for whatever reason_, whenever that site ID exceeds 3 digits, **our API interprets it as a Moment instance instead of a number**.  So, for example, an enrollment site ID of `2077` supplied in the request body of the client is received by the `changeEnrollment` API as a Moment instance of **January 1st, 2077**.  Of course this breaks immediately when trying to use site ID in the where clause of a `findOne` query.

~There's still more work to do to figure out what's actually responsible for causing the string to Moment translation (I spent the entire day trying to track that down to no avail)~ We should certainly specify the format in which dates are expected client-side, but the quickest fix _for now_ is to just have the client specify `enrollment.site.id` as a number, not a string.

----------------------------------------------------------------------------------------------------------------------------

### Associated PR
- https://github.com/ctoec/component-library/pull/100

### DIFF
--- a/client/src/components/Forms/Enrollment/Fields/Site.tsx
+++ b/client/src/components/Forms/Enrollment/Fields/Site.tsx
@@ -28,7 +28,7 @@ export const SiteField = <T extends Enrollment | ChangeEnrollment>({
         getValue: (data) => enrollmentAccessor(data).at('site').at('id'),
         parseOnChangeEvent: (e: React.ChangeEvent<any>) =>
           e.target.value || null,
-        value: site.id,
+        value: `${site.id}`,
         id: `site-${site.id}`,
         preprocessForDisplay: (data) => data == site.id,
       }))}
--- a/client/src/components/Forms/Enrollment/Fields/Site.tsx
+++ b/client/src/components/Forms/Enrollment/Fields/Site.tsx
@@ -28,7 +28,7 @@ export const SiteField = <T extends Enrollment | ChangeEnrollment>({
         getValue: (data) => enrollmentAccessor(data).at('site').at('id'),
         parseOnChangeEvent: (e: React.ChangeEvent<any>) =>
           e.target.value || null,
-        value: `${site.id}`,
+        value: site.id,
         id: `site-${site.id}`,
         preprocessForDisplay: (data) => data == site.id,
       }))}
--- a/src/controllers/children/changeEnrollment.ts
+++ b/src/controllers/children/changeEnrollment.ts
@@ -158,7 +158,6 @@ async function createNewEnrollment(
     child,
     updateMetaData: { author: user } as UpdateMetaData,
   } as Enrollment;
-
   enrollment = tManager.create(Enrollment, enrollment);
   enrollment = await tManager.save(enrollment);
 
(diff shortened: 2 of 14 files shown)


client/src/components/Forms/Enrollment/Fields/Site.tsx
```diff
@@ -28,7 +28,7 @@ export const SiteField = <T extends Enrollment | ChangeEnrollment>({
         getValue: (data) => enrollmentAccessor(data).at('site').at('id'),
         parseOnChangeEvent: (e: React.ChangeEvent<any>) =>
           e.target.value || null,
-        value: `${site.id}`,
+        value: site.id,
```
bamader, line 31: Wait was _this_ the whole reason enrollment changes weren't going through? A type error that was messing up the search in `findOne`? RARGH it is the *worst* when smallest of little things completely break functionality 🤮

bamader (approved):

client/src/components/Forms/Enrollment/Fields/Site.tsx
JordanGuinn: oh it's _so much worse_ than it looks, heh

jlhogan:
for what it's worth, i think this is likely the culprit: https://github.com/ctoec/data-collection/blob/base/src/index.ts#L31
we should just specify the format we expect dates in, and pass those in to the moment instantiation here. then we avoid this weirdness!!!

mmazanec22 (approved):

JordanGuinn:
Agreed!  I think we still want to avoid stringifying `site.id` here too though, just to prevent any additional confusion when the request gets API-side.  We do that everywhere else from what I can see - in fact, the only reason that wasn't happening here was because the `site.id` value comes from a radio button group, which broadcasts string values.

src/controllers/children/changeEnrollment.ts
```diff
@@ -35,9 +35,12 @@ export const changeEnrollment = async (
   changeEnrollmentData: ChangeEnrollment,
   user: User
 ) => {
+  console.debug('changeEnrollment() invoked');
+
   const child = await getChildById(id, user);
 
   if (!child) throw new NotFoundError();
+  console.log(`Matching child found for ${id}.  Initiating enrollment change...`);
```
mmazanec22, line 43: Console logs btw

JordanGuinn: I left these here intentionally - I can probably drop down to `debug`, but they'll likely be helpful in the future for tracking down additional bugs (I know they were for me).

jlhogan: speaking of which, would be awesome to filter logs by level in cloudwatch. BUT prob not worth the investment right now until we know more about how long we expect this tool to live in this infra

jlhogan (approved):

PR #898, opened 2020-12-09
Add request logging

jlhogan:
adds request logging with user ids (i will be able to tell ashley how many people have successfully logged in!)

Might fix the moment weirdness (uses stricter moment parsing by not passing "undefined" as format -- instead, explictly requiring the ISO format that everything gets sent in already)

also updates all the server-side deps b/c might fix a vulnerability on base


package.json
```diff
@@ -12,6 +12,7 @@
     "@types/http-proxy": "^1.17.4",
     "@types/jest": "^26.0.15",
     "@types/jsonwebtoken": "^8.5.0",
+    "@types/morgan": "^1.9.2",
```
mmazanec22, line 15: I don't see this lib used in this PR?

mmazanec22 (approved):

package.json
jlhogan: ha i didnt check in the actual logger

bamader (approved):

PR #902, opened 2020-12-09
628 e2e delete

### LINKED ISSUE #628: Make deleted record e2e test more robust
Check that the child deleted was the right one

mmazanec22:
Yay they all pass!

Closes #628


jlhogan:
:'''''D

.circleci/config.yml
```diff
@@ -117,10 +117,6 @@ jobs:
           working_directory: e2e-tests
           name: Install e2e test dependencies
           command: yarn
-      - run:
-          working_directory: e2e-tests
-          name: Download example data
```
jlhogan, line 122: why not this anymore?

e2e-tests/tests/deleteChild.js
```diff
@@ -1,36 +1,45 @@
...
+        "//*/button[contains(., 'Yes, delete record')]"
+      );
+      // TODO: change if we change alert header level
+      await browser.waitForElementVisible(
+        'xpath',
+        "//*/h2[contains(., 'Record deleted')]"
+      );
+      await browser.verify.containsText(
+        { locateStrategy: 'css selector', selector: '.usa-alert' },
+        lastName
```
bamader, line 33: Nice! Way cleaner than the selector I would've tried.

.circleci/config.yml
mmazanec22: I don't think we need it anymore since the upload function now downloads the example file to upload

bamader: Looks like the functionality has been moved into an `uploadFile` util that downloads a file for any test that needs it. Probably makes the test suites more robust / independent that way?

mmazanec22: (I didn't implement that-- @bamader was that you?  Is that a correct understanding?)

bamader (approved):
This is the most beautiful thing I've ever seen 😭

.circleci/config.yml
bamader: Oh lol wait I did do that 😂 . Sorry. Yes, you're right though that we don't need it anymore. I moved the code over but didn't delete the stuff in the `config` file.

bamader: I just got too caught up in the row of green checks 😍

jlhogan: awesome yes makes sense!!! i didn't know we could upload a file from the local-to-test machine

jlhogan (approved):

.circleci/config.yml
mmazanec22: I didn't think it was possible!  Huge kudos to @bamader for figuring that one out

jlhogan: woooooooooooooooOOO!!! goooooooo @bamader :partying_face:

PR #907, opened 2020-12-10
Fix income determinations

### LINKED ISSUE #867: Going back to edit income det in create child shows empty fields
We're not passing an id in to the family income form for create child flow, so when a user goes back to edit, they don't see the info for the income det they already entered.  One way to solve this might be to useLocation to see if we're in the create record flow and then grab the only income det off the enrollment if there is one.

bamader:
Separates how we retrieve existing determinations into processes for create flow and edit record. Also fixes broken `find` query in income det's `PUT` to use nested alias filtering on the org ID. Closes #867 .


client/src/components/Forms/FamilyIncome/Form.tsx
```diff
@@ -104,12 +110,25 @@ export const FamilyIncomeForm: React.FC<FamilyIncomeFormProps> = ({
...
+  // once in add record
+  if (inCreateFlow) {
+    const dets = child?.family?.incomeDeterminations || [];
+    if (dets.length > 1) {
+      throw new Error(
+        'Child in add record flow cannot contain more than one income determination'
+      );
+    }
+    determination = dets[0] || ({} as IncomeDetermination);
+    console.log(determination);
```
jlhogan, line 125: console log

client/src/components/Forms/FamilyIncome/Form.tsx
```diff
@@ -104,12 +110,25 @@ export const FamilyIncomeForm: React.FC<FamilyIncomeFormProps> = ({
...
-  // TODO: if we're in the create record flow, just find the income det that exists
-  // If more than one exists in the create record flow, throw an error
+  let determination: IncomeDetermination;
+  // If we're in the create flow, want to restrict a child to only
+  // having a maximum of one determination--can't make two at
+  // once in add record
+  if (inCreateFlow) {
+    const dets = child?.family?.incomeDeterminations || [];
+    if (dets.length > 1) {
+      throw new Error(
```
jlhogan, line 120: i dont know that we actually gain anything, graceful degradation-wise by having this check & error throw. IMO it's fine to just say IF in create flow, THEN grab first det if exists or empty object ELSE attempt to grab the det with this id, or empty object
very possible i'm missing something about how this check could help, but like, it doesnt do anything to prevent us from getting to the bad state, just prevents the user from moving forward if they somehow did end up in a weird (and not really possible) data state.

src/routes/families.ts
```diff
@@ -47,16 +47,22 @@ familyRouter.put(
...
+          qb.andWhere(
+            'IncomeDetermination__family__organization.id IN (:...readOrgIds)',
+            { readOrgIds }
+          );
         },
-        {
-          where: { family: { organization: { id: In(readOrgIds) } } },
-        }
-      );
+        loadEagerRelations: true,
```
jlhogan, line 62: why is this needed?

src/routes/families.ts
```diff
@@ -47,16 +47,22 @@ familyRouter.put(
...
       const readOrgIds = await getReadAccessibleOrgIds(req.user);
 
-      const detToModify = await getManager().findOne(
-        IncomeDetermination,
-        {
-          id: detId,
-          familyId: famId,
+      // Need to build out the options dict because we want to do property
+      // filtering on a nested relation alias
+      const opts: FindOneOptions<IncomeDetermination> = {
```
jlhogan, line 52: an option here is to do what we do in the enrollment/funding controller: 
first try to get the appropriate family
If the family with given orgId doesn't exist, throw not found
Then if the family exists, grab the income determination
 If determination with given family id doesn't exist throw not found

An extra query, but perhaps worth it for readability since you can then just use vanilla typeorm and not have to do the alias-based manual wheres

client/src/components/Forms/FamilyIncome/Form.tsx
bamader: No that's totally fair. I just put the check in because there was a comment about wanting to throw that error but I have zero attachment to it.

src/routes/families.ts
bamader: I find the alias-based where to be a huge pain to read so if the extra query is okay I would def prefer to have that in there

mmazanec22 (approved):

src/routes/families.ts
```diff
@@ -47,14 +47,22 @@ familyRouter.put(
       const detId = parseInt(req.params['determinationId']);
       const readOrgIds = await getReadAccessibleOrgIds(req.user);
 
+      // Verify the user has access to the determination via
+      // family lookup--as in enrollment/funding controller
```
jlhogan, line 51: can we wrap this in a function and reuse for the determination delete logic as well

bamader: Yeah of course. Incoming.

mmazanec22 (approved):

src/routes/families.ts
```diff
@@ -46,18 +47,11 @@ familyRouter.put(
       const famId = parseInt(req.params['familyId']);
       const detId = parseInt(req.params['determinationId']);
       const readOrgIds = await getReadAccessibleOrgIds(req.user);
```
jlhogan, line 49: nit: Can push this into the controller helper func

jlhogan (commented):
can you add a test for this to the `integrationTests` in client? The way I wrote the child/enrollment ones rely on the fact that they happen after e2e tests and so data will already exist in the DB. I could NOT get file upload to work from the node runtime, even with our new browser-automagic-free method of writing the file data + headers for posting multipart form data. writing income determination tests you could:

bamader:
Looks like Melanie's working on the income determination integration tests, so I'm guessing that'll be a separate branch/PR from this. Do we want to wait to merge this in until those tests are done, or can this merge before that one lands?

jlhogan (approved):
nope i think we should merge this for sure!

src/middleware/error/passAsyncError.ts
```diff
@@ -12,9 +12,7 @@ import { RequestHandler, Request, Response, NextFunction } from 'express';
  * @param func The request handler function
  */
 export function passAsyncError(func: RequestHandler) {
-  return function (req: Request, res: Response, next: NextFunction) {
-    // @ts-ignore
-    // This works the way it needs to so I'm not sure why typescript is complaining
-    func(req, res, next).catch(next);
+  return async function (req: Request, res: Response, next: NextFunction) {
```
mmazanec22: Woo yay!

PR #912, opened 2020-12-10
771 reorder nav

### LINKED ISSUE #771: Reorder menu tabs: Rename getting started to home. move file upload before Roster

### LINKED ISSUE #878: Change all "batch upload" labels to "file upload"

mmazanec22:
Closes #771  ...maybe? Was there more to this AC than is written in the ticket, or did we give it two points by mistake?

Closes #878 too

### DIFF
--- a/client/src/containers/App/App.tsx
+++ b/client/src/containers/App/App.tsx
@@ -31,13 +31,13 @@ const App: React.FC = () => {
             },
             {
               type: 'primary',
-              text: 'Roster',
-              path: '/roster',
+              text: 'File upload',
+              path: '/upload',
             },
             {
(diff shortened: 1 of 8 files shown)


bamader (approved):
If I had to guess, I think the two points in #771 comes from changing the names of the pages in the different routes. Ryan said that the feedback we got implied that we should change the name of the page we're calling "Getting Started" to be Home, and that we should change the name of the page we're calling "Home" to Landing. I think the 2 points came from making all the changes associated with those.

jlhogan (approved):
heh, yeah, who can say. Probably a case of "I clicked two even tho we all agreed on 1"

PR #914, opened 2020-12-11
Totally empty roster

### LINKED ISSUE #877: Roster empty state updates
When I go to my roster before any data is added to it:
[Design](https://share.goabstract.com/42315efe-ac59-43d4-aea1-f47b0cbf4428?collectionLayerId=53aac75b-10a9-4a94-b18f-ed35d4e3c47f&mode=design)

Screen Shot 2020-12-07 at 1 52 29 PM

- The views, actions, bottom-button bar and tab-nav are hidden
- The info alert text is updated to let users know about both file upload and adding records individually

When I go to a site roster with no records, I see:
Screen Shot 2020-12-07 at 1 54 04 PM

bamader:
Implements an empty roster card that appears for users who have no recorded enrollments at all and who have no enrollments at a particular site. Also modifies a few buttons so that their icons have the same vertically centered alignment as their text (Add Record was really bugging me). Closes #877 .


client/src/components/AddRecordButton.tsx
```diff
@@ -25,22 +30,31 @@ export const AddRecordButton: React.FC<AddRecordButtonProps> = ({
...
-      >
-        <TextWithIcon Icon={PlusCircle} text="Add a record" />
-      </Link>
+      // Wrap in button so that icon is vertically centered with
+      // the same alignment the text is
+      <Button
+        appearance="unstyled"
+        className={formattedClassName}
+        text={
+          <Link
```
mmazanec22, line 39: We could totally use the href attribute on our button component instead-- may need to add the possibility of state for that.  Another alternative might be to put it in a paragraph or span or something that will have the same effect??  I think right now it might be a lil weird that it's <button><a></a></button> under the hood

client/src/components/UploadFileButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export const UploadFileButton: React.FC<UploadFileButtonProps> = ({
+  className,
+}) => {
+  // Use block display to ensure correct inline format
+  // (e.g. spacing with other buttons and text)
+  const formattedClassName = cx('display-block', className);
+  return (
+    <Button
```
mmazanec22, line 17: Same comment here

client/src/components/AddRecordButton.tsx
bamader: Just want to clarify so I understand the ask here: what should be put in the paragraph or span tag / what kind of formatting were you thinking? Using a link seemed like a pretty natural extension to me of a pattern I've seen us use in a couple places (Buttons wrapping a TextWithIcon), but if it's too weird I can change it up. I've tried cutting out the Link and just using the href option with putting `<p>` and `<span>` tags around the text prop, though, and it's giving really weird text alignment where the button text is like half a font size higher than the surrounding words 😕

RyanHansz:
@bamader can we see this in QA before it gets merged?

bamader:
@RyanHansz Yeah definitely! I'll wait until we clear up anything that needs changed and I get approvals, and then I'll let you know when I push it to QA.

client/src/components/AddRecordButton.tsx
mmazanec22: The html this puts out is a button wrapping a link. That would be confusing for screen readers since it's two interactive elements nested. We should not use a button element here (because it goes somewhere, so it should be a link), but you can use our button component as a link (by adding an href attribute) if you want to take advantage of the consistent styling.

bamader: Ah  I see. So either: a) get rid of the button altogether and just use a  link, because the button is bad accessibility design  in nesting interactive elements, or b) keep the button but use the `href` attribute. I'll try approach  a to see what kinds of  styling  we get. Maybe without so many layers of nested elements it'll match the alignment of the rest of the text.

jlhogan (approved):

client/src/components/AddRecordButton.tsx
bamader: I have no idea why the styling of this was giving me trouble on Friday. Your first suggestion of just using a Link with the text with Icon appears exactly as it needs to 🤷‍♂️

bamader: Wrapping the href button in a div element with an inline block formatting seems to have done the trick. For the life of me, I could not figure out styling elements that would situate the icons on the same vertical plane as the text while also keeping the TextWithIcon in the same vertical plane as the rest of the message.

mmazanec22 (approved):

RyanHansz:
Looks great to me!

PR #921, opened 2020-12-14
549 backend error messages

### LINKED ISSUE #484: Add date validations
For dates that are in the future change the error message to say:
- [Field name] cannot be in the future. 
- Make sure dates are in MM/DD/YYYY format
-- for example: 030/02/2020 should throw an error
- Other date validations:
- Funding period can't begin before an enrollment

### LINKED ISSUE #549: use backend validation error  messages
- Validation messages displayed in the application should come from the messages in the api responses
- error status should be displayed for hidden forms (i.e. enrollment, funding, family income det) when they are collapsed
  - right now, we only show the `!` if the field is missing. need to update this check to show the `!` if the field has a validation error.

mmazanec22:
Closes #549 closes #484 

Some of our funding validations weren't working because I misunderstood the nested validations when I wrote them.  I changed them to follow the same pattern we're using for child/family.

### DIFF
--- a/client/src/containers/BatchEdit/EnrollmentFunding/__snapshots__/Form.test.tsx.snap
+++ b/client/src/containers/BatchEdit/EnrollmentFunding/__snapshots__/Form.test.tsx.snap
@@ -86,7 +86,7 @@ exports[`BatchEdit EnrollmentFundingForm matches snapshot 1`] = `
         id="status-firstReportingPeriod"
         role="alert"
       >
-         First reporting period is required for OEC reporting.
+         Field is required for OEC reporting.
       </span>
       <select
         aria-describedby="status-firstReportingPeriod"
--- a/client/src/utils/formatters/sentenceCase.ts
+++ b/client/src/utils/formatters/sentenceCase.ts
@@ -1,0 +1,3 @@
+export function capitalizeFirstLetter(input: string) {
+  return input.charAt(0).toUpperCase() + input.slice(1);
+}
--- a/client/src/utils/getValidationStatus.ts
+++ b/client/src/utils/getValidationStatus.ts
@@ -1,8 +1,36 @@
-import React from 'react';
 import { FormStatusProps, TObjectDriller } from '@ctoec/component-library';
 import { ValidationError } from 'class-validator';
 import { ReactNode } from 'react';
-import { ObjectWithValidationErrors } from '../shared/models/ObjectWithValidationErrors';
+import {
+  EnrichedValidationError,
+  ObjectWithValidationErrors,
+} from '../shared/models/ObjectWithValidationErrors';
+import { capitalizeFirstLetter } from './formatters/sentenceCase';
+
+const getIsRequiredMessage = (fieldName: string) =>
+  `${fieldName} is required for OEC reporting.`;
+
+function getValidationErrorMessage(validationError: EnrichedValidationError) {
+  // Constraints i.e. { isNotEmpty: 'Error message from the backend about the field being empty' }
+  const { constraints } = validationError || {};
+  const fieldName = capitalizeFirstLetter(
+    validationError?.metadata?.formattedName || 'Field'
+  );
+  // Use the is required error message as a default to make typescript happy
+  // There is probably not acctually a case where there is an error but no constraints
+  // unless someone writes a custom validation and forgets
+  if (!constraints) return getIsRequiredMessage(fieldName);
+
+  return Object.keys(constraints).reduce((prev, current) => {
+    // Returns a string of sentences, one for each validation error
+    let appendMessage = constraints[current];
+    if (current === 'isNotEmpty') {
+      appendMessage = getIsRequiredMessage(fieldName);
+    }
+    if (!prev) return appendMessage;
+    return `${prev} ${appendMessage}`;
+  }, '');
+}
 
 /**
  * Wrapper function around getValidationStatusForField that passes in option
@@ -16,6 +44,8 @@ export function getValidationStatusForFieldInFieldset<
   path: string,
   fieldProps: any & { label: string }
 ): FormStatusProps | undefined {
+  // TODO: do we actually want to do this?
+  // Are there any cases where we want to show the error for each field in a fieldset?
   return getValidationStatusForField(objectDriller, path, fieldProps, {
     message: undefined,
   });
@@ -26,15 +56,6 @@ export type ValidationStatusOptions = {
   id?: string;
   message?: string;
 };
-
-export function drillReactNodeForText(inputNode: ReactNode | string): string {
-  if (typeof inputNode === 'string') {
-    return inputNode;
-  } else if (React.isValidElement(inputNode)) {
-    return drillReactNodeForText(inputNode.props.children);
-  }
-  return '';
-}
 
 /**
  * The form field passes object driller and path to the validation function,
@@ -51,7 +72,7 @@ export function getValidationStatusForField<
 >(
   objectDriller: TObjectDriller<NonNullable<T>>,
   path: string,
-  fieldProps: any & { label: string | ReactNode },
+  _: any & { label: string | ReactNode },
   options?: ValidationStatusOptions
 ): FormStatusProps | undefined {
   const splitPath = path.split('.');
@@ -72,9 +93,7 @@ export function getValidationStatusForField<
   return {
     type: 'error',
     id: `status-${field}`,
-    message: `${drillReactNodeForText(
-      fieldProps.label || fieldProps.legend
-    )} is required for OEC reporting.`,
+    message: getValidationErrorMessage(validationError),
     ...options,
   };
 }
@@ -111,7 +130,14 @@ export function getValidationStatusForFields<
     fields.includes(v.property)
   );
   if (!validationErrors.length) return;
-  const message = 'This information is required for OEC reporting.';
+  const message = validationErrors.reduce((prev, current) => {
+    // Result from get validation error message is a string of sentences.
+    // This chains those together.
+    // Override this default with a custom message in options if you don't want this.
+    const newMessage = getValidationErrorMessage(current);
+    if (!prev) return newMessage;
+    return `${prev} ${newMessage}`;
+  }, '');
 
   return {
     type: 'error',
--- a/src/entity/Child.ts
+++ b/src/entity/Child.ts
@@ -46,14 +46,14 @@ export class Child implements ChildInterface {
   uniqueId?: string;
 
   @Column({ nullable: true })
-  @IsNotEmpty({ message: 'First name is required' })
+  @IsNotEmpty()
   firstName?: string;
 
   @Column({ nullable: true })
   middleName?: string;
 
   @Column({ nullable: true })
-  @IsNotEmpty({ message: 'Last name is required' })
+  @IsNotEmpty()
   lastName?: string;
 
   @Column({ nullable: true })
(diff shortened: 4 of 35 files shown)


client/src/containers/EditRecord/EditRecord.tsx
```diff
@@ -95,6 +95,7 @@ const EditRecord: React.FC = () => {
     afterSaveSuccess: () => setTriggerRefetchCounter((r) => r + 1),
     setAlerts,
   };
+  console.log(child);
```
belucid, line 98: @mmazanec22 did you mean to keep this in?

mmazanec22: Nope!

client/src/utils/formatters/sentenceCase.ts
```diff
@@ -0,0 +1,3 @@
+export function sentenceCase(input: string) {
+  return input.charAt(0).toUpperCase() + input.toLowerCase().slice(1);
```
belucid, line 2: Do we want to lower case the rest of the characters? Won't this affect intentional casing in the rest of the sentence for names, proper nouns, acronyms, etc.?

mmazanec22: AFAIK there are no names or proper nouns in any of our labels

mmazanec22: Ugh but there is an abbreviation-- will handle that edge case in the error message func

belucid:
@mmazanec22 given that there aren't automated test changes to cover these error message changes, and also I know you went through some manual QA on this (we discussed it on the standup), it's good practice to include the manual tests steps you performed in the PR description. This helps the reviewer a lot, for a lot of reasons.

1) The reviewers knows manual QA has been done (a PR submitter is quite unlikely to document test steps they didn't perform)
2) It's easier to review the QA thoughtfully when it's written down... "does this cover every case... anything missing"
3) Reviewers should really be repeating any manual QA, and it's much easier and saves a lot of time if there are steps they can just follow. Repeating the manual QA helps find more issues (everyone's env. is just a little different, they perform steps with minor variations), and the going through the QA steps themselves almost always helps the reviewer with #2 above.

Ideally we'd also be using a template for PR descriptions, but that's something we can discuss as a team on the next retro.

client/src/containers/BatchEdit/EnrollmentFunding/__snapshots__/Form.test.tsx.snap
```diff
@@ -86,7 +86,7 @@ exports[`BatchEdit EnrollmentFundingForm matches snapshot 1`] = `
         id="status-firstReportingPeriod"
         role="alert"
       >
-         First reporting period is required for OEC reporting.
+         Field is required for OEC reporting.
```
jlhogan, line 89: does this change mean we lose the ability to drill down and get display name from html label?

client/src/utils/getValidationStatus.ts
```diff
@@ -1,8 +1,31 @@
...
+  EnrichedValidationError,
+  ObjectWithValidationErrors,
+} from '../shared/models/ObjectWithValidationErrors';
+import { sentenceCase } from './formatters/sentenceCase';
+
+const getIsRequiredMessage = (fieldName: string) =>
+  `${fieldName} is required for OEC reporting.`;
+
+function getValidationErrorMessage(validationError: EnrichedValidationError) {
+  const { constraints } = validationError || {};
```
jlhogan, line 14: can you throw in some comments here. What is the constraint, and why does its absence mean we use the is required message?

client/src/utils/getValidationStatus.ts
```diff
@@ -16,6 +39,7 @@ export function getValidationStatusForFieldInFieldset<
   path: string,
   fieldProps: any & { label: string }
 ): FormStatusProps | undefined {
+  // TODO: do we actually want to do this?
```
jlhogan, line 42: why not? what's up ?

client/src/utils/getValidationStatus.ts
```diff
@@ -111,7 +124,11 @@ export function getValidationStatusForFields<
     fields.includes(v.property)
   );
   if (!validationErrors.length) return;
-  const message = 'This information is required for OEC reporting.';
+  const message = validationErrors.reduce((prev, current) => {
```
jlhogan, line 127: this is chaining multiple messages for multiple fields? When do we want to do this? what does the end result look like? String of sentences?

src/entity/Child.ts
```diff
@@ -46,14 +46,14 @@ export class Child implements ChildInterface {
...
-  @IsNotEmpty({ message: 'First name is required' })
+  @IsNotEmpty()
   firstName?: string;
 
   @Column({ nullable: true })
   middleName?: string;
 
   @Column({ nullable: true })
-  @IsNotEmpty({ message: 'Last name is required' })
+  @IsNotEmpty()
```
jlhogan, line 56: Ah I see so we're going with: if no message from backend, then assume it is a 'missing but required' thing and that's why no constraint in my comment above means use that error message :) not sure where, but should capture that somewhere. Maybe all the comments can livein the FE utils?

src/entity/Child.ts
```diff
@@ -163,7 +163,8 @@ export class Child implements ChildInterface {
 
   @ValidateNested()
   @ValidateIf((child) => {
-    if (child.family) child.family.childIsFoster = child.foster;
+    if (child.family)
+      child.family.childIsFoster = child.foster === UndefinableBoolean.Yes;
```
jlhogan, line 167: ooooo

src/entity/Enrollment.ts
```diff
@@ -63,9 +64,17 @@ export class Enrollment implements EnrollmentInterface {
 
   @Column({ type: 'date', nullable: true, transformer: momentTransformer })
   @IsNotEmpty()
+  @MomentComparison({
+    compareFunc: (entry: Moment) => entry.isBefore(moment()),
+    message: 'Entry cannot be in the future.',
```
jlhogan, line 69: Did this come from some specific AC somewhere, or just something that seems logical?

src/entity/Enrollment.ts
```diff
@@ -80,9 +89,22 @@ export class Enrollment implements EnrollmentInterface {
...
   @ValidateNested({ each: true })
-  @FundingDoesNotOverlap()
   @FundingAgeGroupMatchesEnrollment()
   @IsNotEmpty()
+  @ValidateIf((enrollment) => {
+    // This is used by each funding and then deleted
+    // See similar pattern between child and family
+    enrollment.fundings.forEach((funding) => {
+      // Fundings must be undefined or we have circular json
+      funding.enrollment = { ...enrollment, fundings: undefined };
```
jlhogan, line 99: does funding specifically need enrollment for some validation? or does it just need the other fundings from the enrollment ?

src/entity/decorators/Funding/fundingBeginsAfterEnrollmentEntry.ts
```diff
@@ -0,0 +1,28 @@
...
+    registerDecorator({
+      name: 'fundingStart',
+      target: object.constructor,
+      propertyName,
+      options: { message, ...validationOptions },
+      constraints: [],
+      validator: {
+        validate(reportingPeriod: ReportingPeriod, { object: funding }) {
+          const { entry } = (funding as Funding)?.enrollment;
+          // TODO: should this be based on period start or period end???
```
jlhogan, line 19: the end of the first reporting period must be after the enrollment start

src/utils/validateObject.ts
```diff
@@ -1,15 +1,22 @@
...
-  const validationErrors = await validate(object, {
+  let validationErrors = await validate(object, {
     validationError: { value: false, target: false },
   });
-  return distributeValidationErrorsToSubObjects(object, validationErrors);
+  const metadata = getAllColumnMetadata();
+  return distributeValidationErrorsToSubObjects(
+    object,
+    validationErrors,
+    metadata
```
jlhogan, line 18: So we're adding the metadata just to get the formatted name? Can we no longer get that from the react node in the UI?

src/entity/Enrollment.ts
mmazanec22: AC is first bullet here: https://github.com/ctoec/data-collection/issues/484

client/src/utils/getValidationStatus.ts
mmazanec22: When we use this func we're not showing granular error messages for each field in the fieldset.  That makes sense for something like a date (maybe), but could be annoyingly vague in cases where a particular value is a blocker

client/src/utils/getValidationStatus.ts
mmazanec22: Yep, string of sentences

src/entity/Enrollment.ts
mmazanec22: Yep, for `FundingBeginsAfterEnrollmentEntry`

src/utils/validateObject.ts
mmazanec22: We can.  I made this change bc it seemed more stable to treat metadata as the source of truth so that there would be consistency in what the user sees across the app, between data requirements, the template, and our forms.  Was wondering even if we should consider using metadata as the default for field labels, and eventually for help text if we add that in.  As it stands there are a bunch of places we need to update (front end, metdata, error messages created in the backend) when terminology changes.  Can revert this if we're worried about label/error message consistency though!

jlhogan: no that concerned. I think that's a smart idea -- that this should be the source of truth for everything (obviously, not to implement in full right now, but to set up for for thefuture). Given that, I  wonder if shoving _ALL_ the metadata in each validation error is necessary (theres a lot of text in there). Might be worth just popping the formatted and property names on here as opposed to the whole object?

mmazanec22: That's fair!  Given the other comments about capitalization I wonder if we should change the metadata to default to not capitalized, then capitalize it where we use it if necessary (i.e. title casing in the template if we want that and sentence casing in the front end).  Would that break anything if I snuck it into this PR?

jlhogan: If the names get title-cased in the template, then that'll need to be accounted for here: https://github.com/ctoec/data-collection/blob/base/src/controllers/enrollmentReports/parse.ts#L289 (prob just make the comparison of headers case-insensitive by lower-casing the ones from the template?)

client/src/utils/getValidationStatus.ts
jlhogan: well i think when we want to display the full error on the field, just use getValidationStatusForField, right? like this is specifically for when we want the message-less validation status (and maybe should be renamed as such?)

src/template/constants.ts
```diff
@@ -4,7 +4,7 @@ import moment from 'moment';
 // TODO: add pre-merge hook that updates these values (probably put them in a diff file)
 // automatically when this file gets updated
 export const TEMPLATE_VERSION = 1;
-export const TEMPLATE_LAST_UPDATED = moment.utc('11-24-2020', ['MM-DD-YYYY']);
+export const TEMPLATE_LAST_UPDATED = moment.utc('12-16-2020', ['MM-DD-YYYY']);
```
jlhogan, line 7: :raised_hands:

jlhogan (approved):

bamader (approved):
Wow that's comprehensive. Looks great!

PR #929, opened 2020-12-16
908: normalize strings better

### LINKED ISSUE #908: accept user input with punctuation
Just saw a user losing data in file upload due to `'` in site name and `.` in `U.S.` (as opposed to `US`)

Probably, we should strip all punctuation and maybe also whitespace when comparing user inputs to constant values 

i think this regex already exists somewhere in the mapping functionality; make sure we use it!

jlhogan:
closes #908


src/controllers/enrollmentReports/mapUtils/enum.ts
```diff
@@ -60,11 +60,9 @@ export const mapEnum = <T>(
...
-      let normalizedReferences = normalizeString(refString, true);
-      if (!Array.isArray(normalizedReferences)) {
-        // Make typescript happy
-        normalizedReferences = [normalizedReferences];
-      }
+      let normalizedReferences = refString
+        .split('-')
+        .map((stringPart) => normalizeString(stringPart));
       normalizedReferences.forEach((_normalizedReference) => {
         if (processedStringsMatch(normalizedInput, _normalizedReference)) {
```
mmazanec22, line 67: I kind of wonder if we should rely more on the string match func so we can employ that elsewhere (including header match!) with a lower lift.  Like, maybe it should just normalize the strings and compare them so we don't have to do that each time.

mmazanec22 (approved):

src/controllers/enrollmentReports/mapUtils/enum.ts
jlhogan: I don't know that we want to do the specific match that we do here in other places (specifically if str1 starts with str2 or str2 starts with str1). If yes, then absolutely we can pull it out into a util. But it wasn't obvious to me that that would be the appropriate comparison outside of the enum case

bamader: Not sure about the others, but with `headerMatch` in the e2e tests it's a strict equality comparison. In practice checking the starts with's probably wouldn't hurt anything (unless something unexpected crept into a header), but my pref would be to leave them separate. Not a strong feeling, though, so happy to go either way.

bamader (approved):

src/controllers/enrollmentReports/mapUtils/enum.ts
jlhogan: I dont think we're talking about leveraging this in the e2e tests so much as elsewhere within the server logic

PR #940, opened 2020-12-18
918 income det tests

### LINKED ISSUE #918: Income determination tests
integration tests in client/src/integrationTests that cover all `family` and `family/id/income-determination` routes

mmazanec22:
Closes #918

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -44,7 +44,7 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --silent --testPathIgnorePatterns ./*/integrationTests",
-    "test-integration": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --testPathPattern ./*/integrationTests",
+    "test-integration": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --silent --testPathPattern ./*/integrationTests",
     "test-local": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen",
     "eject": "react-scripts eject"
   },
--- a/client/package.json
+++ b/client/package.json
@@ -44,7 +44,7 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --silent --testPathIgnorePatterns ./*/integrationTests",
-    "test-integration": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --silent --testPathPattern ./*/integrationTests",
+    "test-integration": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --testPathPattern ./*/integrationTests",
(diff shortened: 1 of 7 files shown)


client/src/integrationTests/incomeDetermination.test.ts
```diff
@@ -0,0 +1,126 @@
...
+          determinationDate: moment(),
+          family: familyToUpdate,
+        } as IncomeDetermination;
+
+        const res = await apiPost(
+          `families/${familyToUpdate.id}/income-determinations`,
+          newDet,
+          TEST_OPTS
+        );
+        expect(res.id).toBeGreaterThan(detToUpdate?.id);
```
bamader, line 91: Oooo that's clever

JordanGuinn (approved):

src/routes/families.ts
```diff
@@ -44,7 +43,22 @@ familyRouter.put(
...
+        {
+          id: detId,
+          familyId: famId,
+        },
+        {
+          where: { family: { organization: { id: In(readOrgIds) } } },
+        }
+      );
+      if (!detToModify) throw new NotFoundError();
+
```
bamader, line 61: Is there a reason to not use the controller func? We made that switch to follow the same access patterns as for Enrollments and Fundings because nesting the property that far down in the `where` clause was causing NotFound problems that the controller func fixed (by checking family access then determination existence).

bamader (commented):
The test itself looks great, just need some clarity on `src/routes/families.ts`

Edit: Here was the PR where we switched to a controller func https://github.com/ctoec/data-collection/pull/907

client/package.json
```diff
@@ -44,7 +44,7 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --silent --testPathIgnorePatterns ./*/integrationTests",
-    "test-integration": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --silent --testPathPattern ./*/integrationTests",
+    "test-integration": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --testPathPattern ./*/integrationTests",
```
jlhogan, line 47: FWIW, i added the local command to run all tests with console logs; you can use it to run only integration tests with console logs by:
```
$ yarn test-local --testPathPattern ./*/integrationTests
```

client/src/integrationTests/incomeDetermination.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+        const updatedChild: Child = await apiGet(
+          `children/${childToUpdate.id}`,
+          '',
+          TEST_OPTS
+        );
+        const createdDet = updatedChild?.family?.incomeDeterminations?.find(
+          (d) => d.income === newDet.income
```
jlhogan, line 99: nit: you could do the comparison on id, since you have it in res above

src/routes/families.ts
jlhogan: yeah i'm interested that this works, because we previously found that nested where clauses like this did not in fact work?

client/package.json
mmazanec22: Ohh didn't even see that there-- will undo this

src/routes/families.ts
mmazanec22: I have no idea how or why I made this change, will def revert it

bamader (approved):

PR #953, opened 2020-12-21
Split upload

### LINKED ISSUE #629: Make file upload tests more robust
After upload bugs are fixed, split upload tests into one that tests complete data upload and one that tests incomplete data upload.

bamader:
Splits the file upload e2e test into tests for the two kinds of missing data (the missing-one and missing-some conditions). Serves as the broad test mentioned in #629 (not doing more specific testing yet cuz the income not disclosed stuff seems higher priority). Closes #629 .


e2e-tests/tests/uploadMissingSome.js
```diff
@@ -0,0 +1,13 @@
+const { login } = require('../utils/login');
+const { uploadFile } = require('../utils/uploadFile');
+
+module.exports = {
+  '@tags': ['upload', 'missing-some'],
+  uploadMissingOne: async function (browser) {
```
mmazanec22, line 6: Looks like you forgot to rename this func

mmazanec22 (approved):

e2e-tests/tests/uploadMissingSome.js
bamader: Oh whoops. Thank you!

e2e-tests/tests/uploadComplete.js
```diff
@@ -2,9 +2,7 @@ const { login } = require('../utils/login');
 const { uploadFile } = require('../utils/uploadFile');
 
 module.exports = {
-  '@tags': ['upload'],
-  // uploadWrongFormat: async function (browser) {},
-  // uploadMissingInfo: async function (browser) {},
+  '@tags': ['upload', 'complete'],
```
jlhogan, line 5: this is a nit, but i understand tags to be used to group tests that we might want to run as functional groups, so a tag that only pertains to a single test isn't really necessary because to run a single test, can just pass the path to the test file. So, `upload` is likely sufficient for all of these guys.

jlhogan (approved):

PR #963, opened 2020-12-28
Income not disclosed

### LINKED ISSUE #938: Income not disclosed option
WHY THESE CHANGES
- After talking with programs, we discovered that OEC's income determination requirements are unclear, and there will be child records without income. To accommodate this, we will add an "income not disclosed" field to income determinations rather than trying to pin down specific scenarios where it is required vs not. 

**ADD A NEW DATA FIELD**
Income not disclosed (Optional)
Use this field if income is not collected for any reason

Definition
Indicate that the child's record does not have a documented income determination. Reasons may include children living with foster families, receiving special education/disability services, or other circumstances.

Reason for collecting
Indicates records where income is not collected 

Format
Yes
Y

**UPDATE INCOME DETERMINATION FIELDS REQUIRED STATUS (Household size, Annual household income, Determination date)**
Required if 
Required unless the "income not disclosed" field has a value of "yes"

**UPDATE FAMILY INCOME SECTION DEFINITION**
This data is collected to assess eligibility for public funding. Income determinations must be updated at least once a year. 
**For the July 1 to Dec 31 data collection:** If a child has had multiple income determinations, you only need to enter the most recent determination. 

**ADDING AND VIEWING CHILD RECORDS**
- Remove the info alert on the income determination page for children living with foster families. 
- Add an income not disclosed radio button to the income determination form when adding a record and adding an income determination on a child record: 
- Separate the income not disclosed element from the rest of the form with a divider.
- On checking "income not disclosed" all other fields become disabled
Screen Shot 2020-12-18 at 11 33 39 AM
Screen Shot 2020-12-18 at 11 33 14 AM

- Saving an "income not disclosed" income determination adds another income determination card with text of "Income not disclosed"
- Editing an income not disclosed income will open the income determination form as usual (except that the income not disclosed radio button will be checked)

Screen Shot 2020-12-18 at 11 34 55 AM

bamader:
Add the `incomeNotDisclosed` field to the data model. Change the edit record income form to accommodate this.
Closes #938 .

Update: Depends on https://github.com/ctoec/component-library/pull/109 to get the disabled styling the way we want it.


client/src/components/Forms/FamilyIncome/Fields/AnnualHouseholdIncome.tsx
```diff
@@ -1,23 +1,47 @@
...
  * Component that receives the annual household income of a given
  * family and records that as part of the income determination.
  */
 export const AnnualHouseholdIncomeField: React.FC = () => {
+  const {
+    data: determination,
+    dataDriller,
+    updateData,
+  } = useGenericContext<IncomeDetermination>(FormContext);
+  console.log(determination);
```
jlhogan, line 23: console log

client/src/components/Forms/FamilyIncome/Fields/DeterminationDate.tsx
```diff
@@ -12,15 +20,33 @@ import { getValidationStatusForFieldInFieldset } from '../../../../utils/getVali
...
+  const {
+    data: determination,
+    dataDriller,
+    updateData,
+  } = useGenericContext<IncomeDetermination>(FormContext);
   return (
     <FormField<IncomeDetermination, DateInputProps, Moment | null>
-      getValue={(data) => data.at('determinationDate')}
+      getValue={() => {
+        console.log(dataDriller.at('determinationDate'));
```
jlhogan, line 33: console log

client/src/components/Forms/FamilyIncome/Form.tsx
```diff
@@ -127,12 +100,14 @@ export const FamilyIncomeForm: React.FC<FamilyIncomeFormProps> = ({
       { accessToken }
     );
 
-  const updateDetermination = async (updatedData: IncomeDetermination) =>
+  const updateDetermination = async (updatedData: IncomeDetermination) => {
+    console.log('updated det is ', updatedData);
```
jlhogan, line 104: console log

src/routes/enrollmentReports.ts
```diff
@@ -36,18 +36,32 @@ enrollmentReportsRouter.post(
     return getManager().transaction(async (tManager) => {
       try {
         const reportRows = controller.parseUploadedTemplate(req.file);
+
+        console.log('parsed uploaded template');
```
jlhogan, line 40: console logs

jlhogan (approved):

bamader:
Oops, sorry about all the logging 🤣

client/src/components/Forms/FamilyIncome/Fields/AnnualHouseholdIncome.tsx
```diff
@@ -1,23 +1,47 @@
...
   return (
-    <FormField<IncomeDetermination, TextInputProps, number | null>
-      getValue={(data) => data.at('income')}
+    <TextInput
```
mmazanec22, line 25: Why not use formfield anymore?

client/src/components/Forms/FamilyIncome/Fields/DeterminationDate.tsx
```diff
@@ -12,15 +20,33 @@ import { getValidationStatusForFieldInFieldset } from '../../../../utils/getVali
...
+        return dataDriller.at('determinationDate');
+      }}
       defaultValue={null}
-      parseOnChangeEvent={(e: any) => e}
+      parseOnChangeEvent={(e: any) => {
```
mmazanec22, line 37: Why update data manually here?

client/src/components/Forms/FamilyIncome/Fields/DeterminationDate.tsx
```diff
@@ -12,15 +20,33 @@ import { getValidationStatusForFieldInFieldset } from '../../../../utils/getVali
  * a moment but the function expects a simple date.
  */
 export const DeterminationDateField: React.FC = () => {
+  // Use a driller to keep clearing state synched between data
+  // and the 'income not disclosed' box
+  const {
```
mmazanec22, line 25: I can see that you need the data driller for the disabled attr-- why does the rest of this break the pattern that the rest of the form fields use?

client/src/components/Forms/FamilyIncome/Fields/HouseholdSize.tsx
```diff
@@ -1,22 +1,47 @@
...
+    data: determination,
+    dataDriller,
+    updateData,
+  } = useGenericContext<IncomeDetermination>(FormContext);
   return (
-    <FormField<IncomeDetermination, TextInputProps, number | null>
-      getValue={(data) => data.at('numberOfPeople')}
-      parseOnChangeEvent={(e) => parseInt(e.target.value, 10) || null}
-      inputComponent={TextInput}
+    <TextInput
```
mmazanec22, line 25: Why not a form field here?

client/src/components/Forms/FamilyIncome/Fields/NotDisclosed.tsx
```diff
@@ -0,0 +1,41 @@
...
+export const NotDisclosedField: React.FC = () => {
+  // Need to track this as if it were state, so use a driller to
+  // keep everything properly updated when manipulating the object
+  const {
+    data: determination,
+    dataDriller,
+    updateData,
+  } = useGenericContext<IncomeDetermination>(FormContext);
+  return (
+    <Checkbox
```
mmazanec22, line 20: Same Q here-- why not a form field?

src/entity/Family.ts
```diff
@@ -62,12 +62,6 @@ export class Family implements FamilyInterface {
   @OneToMany(() => IncomeDetermination, (det) => det.family)
   @ValidateNested({ each: true })
   @ArrayMinSize(1)
-  @ValidateIf((family) => {
```
mmazanec22, line 65: There are some other places that need to be updated for this change-- off the top of my head, I think the fake data creation (maybe?) and a front end func that lives with the family income form and determines if it has errors.  Also the family income forms that skip past if the child is foster.

mmazanec22 (commented):
A few q's about form field vs manually using form context.

bamader:
Re: FormField vs. direct input component (TextInput, etc.): (lengthy post incoming cuz this was _super_ frustrating to me during development cuz I didn't know what was going on). I don't know why this is the case (but I have a suspicion...), but using FormFields doesn't seem to allow _other_ form fields or input component to make changes to a value (i.e. form field B can't change form field A's value). I remember having trouble with this *A TON* when I was first doing the not disclosed options for race, birth state, and birth town, because we wanted checking the box to clear the text values as well as disable them, but with FormFields nothing ever got cleared (they also wound up storing their string values still because the field wasn't empty). For this PR, I decided to investigate why this difference existed; I tried every permutation of attributes and props I could for each of FormField and TextInput, but still couldn't get things to clear properly. I've come to believe it's in the difference between two sets of props: `value` vs. `getValue` and `onChange` vs. `parseOnChangeEvent`. For the former two, they just return actual values (strings or numbers), whereas the latter two are functions that drill into the object and access the value indirectly. So my hypothesis is that when I try to have a FormField trigger a change in another field's value, that change doesn't register because form fields can only `get` instead of having direct access to. It's a bit shaky, I'll admit, but this is the same scenario that forced us to use direct inputs for raceNotDisclosed and birthState/Town.

src/entity/Family.ts
bamader: I think I caught the part of the front-end func that determines errors, as well as not skipping the income form in the create flow for a foster child. I'm not seeing anything in the fake data creation that breaks though. Which part do you think the bug lives in?

bamader: Ah I see what you mean in the create section--we still want it to show as complete and not exempt. Fixed it this time!

mmazanec22 (approved):

mmazanec22:
Weird!  We should maybe make a ticket to explore that

RyanHansz:
- We will need to update file template changed date
- Question for everyone: Since the field is optional, do we just need "yes" as an option in the file template? 
  - No reason for a user to put "no" on an optional field
- Since its optional, Income not disclosed should not show in the error modal for rows missing income information
Screen Shot 2020-12-30 at 1 06 07 PM
- Forms are missing disabled styling when income not disclosed is checked
Screen Shot 2020-12-30 at 1 54 18 PM

RyanHansz:
Adding a new income gets flagged as missing info when "income not disclosed" is not checked
Screen Shot 2020-12-30 at 1 57 07 PM

bamader:
@RyanHansz Working on changing the styling of the disabled boxes, but it turns out that the fields _were_ using disabled styling after all--the disabled gray background color for text boxes is just waaaaaay lighter than the disabled color for checkboxes 😑

PR #973, opened 2020-12-30
Set alerts to empty array if there are no children with errors

### LINKED ISSUE #969: Batch edit alert should not appear when returning to the roster after completing all records.
https://user-images.githubusercontent.com/63253539/103300235-6dd89a80-49cc-11eb-9e1e-5ea3554a9d6a.mp4

mmazanec22:
Closes #969 

There's still a little delay where the alert shows, I think bc of how caching is working (i.e. for the same reason that we see stale children values for a mo' when the roster loads).

### DIFF
--- a/client/src/containers/Roster/hooks/useChildrenWithErrorsAlert.tsx
+++ b/client/src/containers/Roster/hooks/useChildrenWithErrorsAlert.tsx
@@ -30,11 +30,6 @@ export const useChildrenWithErrorsAlert = (
       return;
     }
 
-    if (!childrenWithErrorsCount) {
-      setAlerts([]);
-      return;
-    }
-
     // set alert if:
     // - no existing matching alert OR
     // - existing matching alert has different number (i.e. was displaying content for a different organization)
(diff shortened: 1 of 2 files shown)


client/src/containers/Roster/hooks/useChildrenWithErrorsAlert.tsx
```diff
@@ -48,6 +53,7 @@ export const useChildrenWithErrorsAlert = (
     }
   }, [childrenWithErrorsCount, isLoading, alertType]);
 
+  if (isLoading || !childrenWithErrorsCount) return {};
```
JordanGuinn, line 56: This is probably something I'm just flat-out ignorant of, but why return an empty object here?

JordanGuinn (approved):

client/src/containers/Roster/hooks/useChildrenWithErrorsAlert.tsx
bamader: Same question, actually, since both of these cases seem to already be covered by the conditionals above?

mmazanec22: Above, if it's still loading, it will return the old alerts.  We should preserve the old alerts and not blow them away in that conditional in case they were set by the previous page, but we prob don't want to display them in case they actually are stale and will be reset or replaced by the current page.

jlhogan (approved):

PR #974, opened 2020-12-30
Only show one alert for missing funding

### LINKED ISSUE #967: Only show one alert when funding is missing
When funding information is missing, only the second alert needs to be shown:
Screen Shot 2020-12-29 at 11 12 54 AM

mmazanec22:
Closes #967 

@RyanHansz I took the AC to mean that there should only be one alert showing when funding info is missing-- even if there is other info missing.  Is that what you meant?

### DIFF
--- a/client/src/containers/EditRecord/EditRecord.tsx
+++ b/client/src/containers/EditRecord/EditRecord.tsx
@@ -70,7 +70,7 @@ const EditRecord: React.FC = () => {
           updatedChild
         );
         if (missingFundedEnrollmentAlertProps) {
-          newAlerts.push(missingFundedEnrollmentAlertProps);
+          newAlerts.splice(0, 1, missingFundedEnrollmentAlertProps);
         }
 
         // Only set success alert on a GET that happens after an update (refetch count > 0)


JordanGuinn (approved):

bamader (approved):

RyanHansz:
> Closes #967
> 
> @RyanHansz I took the AC to mean that there should only be one alert showing when funding info is missing-- even if there is other info missing. Is that what you meant?

Yes that's right!

client/src/containers/EditRecord/EditRecord.tsx
```diff
@@ -70,7 +70,7 @@ const EditRecord: React.FC = () => {
           updatedChild
         );
         if (missingFundedEnrollmentAlertProps) {
-          newAlerts.push(missingFundedEnrollmentAlertProps);
+          newAlerts.splice(0, 1, missingFundedEnrollmentAlertProps);
```
jlhogan, line 73: alternatively put this check first, and just return an array containing this error to exit early. Then do the missing info alert check and return an array with that error if exists. Then no need for the local array var or the splicing.

jlhogan (approved):

client/src/containers/EditRecord/EditRecord.tsx
mmazanec22: We still need to be able to show the success alert in addition to a missing info alert

PR #986, opened 2021-01-04
Mutates roster cache directly, b/c useSWR mutate was not playing nice

jlhogan:


bamader (approved):
Sucks that the mutate function on its own couldn't do what we needed  😕

client/src/contexts/RosterContext/RosterContext.tsx
```diff
@@ -0,0 +1,86 @@
...
+            })}`
+          ) &&
+          // filter out cache values we don't want (useSWR stores errors and isValidating state in cache as well)
+          !key.includes('validating') &&
+          !key.includes('err')
+      );
+
+      // If cache value exists, update it (immutably, with cloneDeep)
+      if (key) {
+        const paginatedChildren: Child[][] = cloneDeep(cache.get(key) || []);
```
JordanGuinn, line 59: **heh**, yeah this looks about right - _oh the joys of caching JS objects_ 😭

client/src/contexts/RosterContext/RosterContext.tsx
```diff
@@ -0,0 +1,86 @@
...
+  setRosterQuery: () => {},
+});
+
+const { Provider } = RosterContext;
+
+const RosterProvider: React.FC = ({ children }) => {
+  const [rosterQuery, setRosterQuery] = useState<RosterQueryParams>();
+
+  /**
+   * useSWR provides a `mutate` callback to update cache values, but I cannot
```
JordanGuinn, line 30: Bit of a broader question here, and _certainly_ not specific to this PR, but long-term should we potentially consider a different caching library?  Especially if we have to actively work **against** some of the fundamental behavior of `swr` just to _update the cache_.

JordanGuinn (approved):

client/src/contexts/RosterContext/RosterContext.tsx
jlhogan: tldr; YUP
https://react-query.tanstack.com/ seems like the big alternative

PR #987, opened 2021-01-04
Use name formatter

### LINKED ISSUE #977: When a child with a missing first or last name is deleted, we should not include "null" in the record deleted alert text
Screenshot from 2020-12-30 12-04-56

Currently the alert shows "null LastName" when the first name is missing.

Whoever takes this ticket should prob check that we're not doing this elsewhere/are consistently using a name formatting util anywhere we display a name

mmazanec22:
Closes #977

### DIFF
--- a/client/src/utils/formatters/name.ts
+++ b/client/src/utils/formatters/name.ts
@@ -1,7 +1,6 @@
 import { Child } from '../../shared/models';
 
-export const firstNameMissingText = '(first name missing)';
-export const lastNameMissingText = '(last name missing)';
+export const firstNameMissingText = '(name missing)';
 
 export const nameFormatter = (
   child?: Child,
--- a/client/src/utils/formatters/name.ts
+++ b/client/src/utils/formatters/name.ts
@@ -1,15 +1,50 @@
 import { Child } from '../../shared/models';
 
+export const firstNameMissingText = '(first name missing)';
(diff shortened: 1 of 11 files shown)


jlhogan (approved):
niiiiiiiiiiiiice!

JordanGuinn (approved):

client/src/utils/formatters/name.ts
```diff
@@ -1,15 +1,49 @@
...
-        child.suffix ? `, ${child.suffix}` : ''
-      }`
-    : `${child.firstName || ''}${
-        child.middleName ? ` ${child.middleName}` : ''
-      } ${child.lastName || ''}${child.suffix ? `, ${child.suffix}` : ''}`;
+  const { lastName, firstName, middleName, suffix } = child || {};
+  const {
+    lastNameFirst,
+    firstOnly,
+    lastOnly,
```
JordanGuinn, line 20: Are the `lastOnly`, `missingFirstAndLastText`, and `missingLastNameText` used anywhere, or going to be used anywhere?

mmazanec22: Not ATM-- there was no copy AC here so I was kinda anticipating changes to how we handle missing names.  Will ping for design review now and take out these opts if they remain unused.

bamader (approved):

PR #1002, opened 2021-01-07
Adds values for existing income dets

jlhogan:
wondering how others feel about the merits of having validation errors around "incomeNotDisclosed" ? I don't know that it actually is useful, but maybe it is? happy to put it back, since this change to the DML prevents the bad state

### DIFF
--- a/src/migration/1608582587797-AddIncomeNotDisclosedToFamily.ts
+++ b/src/migration/1608582587797-AddIncomeNotDisclosedToFamily.ts
@@ -1,4 +1,5 @@
 import { MigrationInterface, QueryRunner } from 'typeorm';
+import { dropDefaultConstraint } from './utils/dropDefaultConstraint';
 
 export class AddIncomeNotDisclosedToFamily1608582587797
   implements MigrationInterface {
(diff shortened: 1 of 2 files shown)


src/migration/1608582587797-AddIncomeNotDisclosedToFamily.ts
```diff
@@ -1,16 +1,22 @@
...
+import { dropDefaultConstraint } from './utils/dropDefaultConstraint';
 
 export class AddIncomeNotDisclosedToFamily1608582587797
   implements MigrationInterface {
   name = 'AddIncomeNotDisclosedToFamily1608582587797';
 
   public async up(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(
-      `ALTER TABLE "income_determination" ADD "incomeNotDisclosed" bit`
+      `ALTER TABLE "income_determination" ADD "incomeNotDisclosed" bit DEFAULT(0) WITH VALUES`
```
JordanGuinn, line 10: Should we default to `false` only here (for historical records), or in the entity as well (for future records)?

JordanGuinn (approved):
I think I agree - the value of `incomeNotDisclosed` seems to come when paired with the presence/absence of an actual income, not on its own.

bamader (approved):
Jordan said it perfectly - the field on its own doesn't help much, it's the pairing with other income fields that's informative.

mmazanec22:
Instead of a validation we could have a test that validates that the other fields are cleared when it is checked

src/migration/1608582587797-AddIncomeNotDisclosedToFamily.ts
jlhogan: i dont think there should be a default for future records -- if they're created wthout the explicit value, then the user didn't add the specific value and we need to prompt them to add it or explicitly say "not disclosed" (at least that's the pattern we've gone with for other fields that are potentially not collected)

PR #1008, opened 2021-01-07
1015- large file uploads

### LINKED ISSUE #1015: Improve file upload speed
the `check` route succeeds quickly enough to handle >2000 rows; the actual upload route times out for anything more than approx 1000 rows. The difference is that in the upload route, we do writes to the DB. 

- instead of saving each row individually, do these in batches (going to have to figure out how to batch create families, then batch create income dets with those references, then batch create children, then enrollments with those references,etc OR set up the ORM to create entire object graph)

jlhogan:
closes #1015 

- bumps up max file size  in nginx config
- massively re-does file row mapping. I attempted to test it robustly manually, but would appreciate any additional stuff you can throw at it or suggestions for tests to add. brain too tired right now :) 
- also redoes how we're doing the cascading soft deletes, because they were only working for enrollment -> funding before. 
- it's still pretty SLOW so i'll spend a little more time timing things and seeing where we can get back some time; but might be worth adding some language (if we detect a large file, or always) saying "this may take a minute" @RyanHansz 

successfully uploaded 1300 records, altho got a timeout for 1800. still playing around, but should work for donna :pray:


jlhogan:
Should I pull the fix for cascade deletes out into separate changeset so we can push taht live ASAP ?
as I write that, it becomes obvious to me that that's a good call. So i'm gonna do that.

src/controllers/enrollmentReports/map/map.ts
```diff
@@ -0,0 +1,275 @@
...
+  mapFunding,
+} from './mapUtils';
+import { EnrollmentReportRow } from '../../../template';
+import { BadRequestError, ApiError } from '../../../middleware/error/errors';
+import { getReadAccessibleOrgIds } from '../../../utils/getReadAccessibleOrgIds';
+
+/**
+ * Can use optional save parameter to decide whether to persist
+ * the mapped results to the DB or not. If we don't persist,
+ * mappnig is still performed to get the data in the right format.
```
mmazanec22, line 32: Teeny tiny nit: mappnig typo

mmazanec22 (approved):

bamader:
Checking this branch out now to do any extra testing on it I can think of

src/controllers/enrollmentReports/map/mapUtils/child.ts
```diff
@@ -0,0 +1,129 @@
...
+) => {
+  return (
+    childA.firstName === childB.firstName &&
+    childA.lastName === childB.lastName &&
+    childA.birthdate &&
+    childA.birthdate?.format('MM/DD/YYYY') ===
+      childB.birthdate?.format('MM/DD/YYYY') &&
+    childA.sasidUniqueId === childB.sasidUniqueId
+  );
+};
```
bamader, line 129: Based on my testing, this function isn't ever returning true. I think this is because the `childA` parameter is an `EnrollmentReportRow` in the function signature, but when it gets called in `map` on line 200, we pass in a `child` object. A `Child` doesn't have the `.sasidUniqueId` field so the last `&&` is always evaluating to false. I altered the function to be 

```
export const isIdentifierMatch = (
  childA: Child,
  childB: EnrollmentReportRow
) => {
  let existingId: string;
  if (childA.sasid) existingId = childA.sasid;
  else if (childA.uniqueId) existingId = childA.uniqueId;

  const idMatch = (!existingId) || (existingId === childB.sasidUniqueId);
  const demoMatch =
    (childA.firstName === childB.firstName &&
    childA.lastName === childB.lastName &&
    childA.birthdate &&
    childA.birthdate?.format('MM/DD/YYYY') ===
      childB.birthdate?.format('MM/DD/YYYY'));
  return idMatch && demoMatch;
};
```

in my local during tests, and it completely solved the historical enrollment parsing problems I mention in #1 in my comment (it just uses the appropriate ID type on the child, if it exists, and doesn't penalize a record if there's no ID type from the org).

bamader (approved):
Looks good! I was able to get ~1400 records uploaded with my own manual testing, so I feel like load-wise we should be well-covered on that front. Also tested all the combinations of missing data I could find, and validations/error messages are behaving as expected. As far as parsing goes, there are just two small things I noticed (not sure we even had these features pre-this-PR, but I know we've talked about them before):

1) fields that we expect to be blank in historical enrollment rows (for example, gender, family address info, income det stuff) are getting flagged in the `checkErrors` modal as missing information; in the withdrawn records tab on the Roster page, these historical enrollments are also missing all the information for these fields from their affiliated child. I left a fix that worked for me as a comment on `/map/mapUtils/child`.
2) for `raceNotDisclosed` and `incomeNotDisclosed`, we're parsing them as true if the user didn't supply any race or income information, even if the notDisclosed field itself isn't filled in. This isn't something I feel strongly about, but I believe in the past, Ryan had said that he'd prefer a user has to actively mark 'Yes' in the respective notDisclosed column for us to parse it that way, rather than us just assuming that missing stuff should be not disclosed (as a way to make users be deliberate about what data should be unknown rather than flagged for them).

The second of these is imo not a blocker to this getting merged, but the first should probably incorporate either the fix I propose or another that makes historical enrollments work before we merge it in.

jlhogan:
@JordanGuinn i know this is approved, but would love your :eyes:  + :heavy_check_mark:  before merge. Hoping to get this deployed before EOD tomorrow so donna can try her big upload ahead of monday time with ryan

src/controllers/enrollmentReports/map/map.ts
```diff
@@ -0,0 +1,275 @@
...
+ * that'll end up in the query (couldn't find any better way to do this
+ * from typeORM), chunk the inserts into batches that will have <2000
+ * parameters per batch.
+ * @param transaction
+ * @param entities
+ */
+async function doBatchedInsert<T>(transaction: EntityManager, entities: T[]) {
+  const parametersPerEntity = Object.keys(entities[0]).length;
+
+  const batchSize = Math.floor(2000 / parametersPerEntity);
```
JordanGuinn, line 265: 2,000 seems like a decent max batch size anyway 👍🏽 Was the parameters per query limit the only real thing that we had to account for here?  Wondering if there's an implicit bulk insert limit w/SQL Server (or if the `save` method even performs a bulk insert behind the scenes)

src/routes/enrollmentReports.ts
```diff
@@ -127,9 +118,12 @@ enrollmentReportsRouter.post(
           req.user,
           { save: true }
         );
-        const report = await tManager.save(
-          tManager.create(EnrollmentReport, { children: reportChildren })
-        );
+
+        // This insert was failing due to too many params, and we don't use it
```
JordanGuinn, line 122: Is this API just... broken?  Do we need to create a ticket for that?  Or just remove it as a route entirely?

src/controllers/enrollmentReports/map/map.ts
```diff
@@ -0,0 +1,275 @@
...
+
+  const createdData: any[] = [];
+  for (let b = 0; b < entities.length; b += batchSize) {
+    const batch = entities.slice(b, b + batchSize);
+    const batchEntities = await transaction.save<T>(batch);
```
JordanGuinn, line 270: Bit of a broader question here: Would it be worthwhile to perform any explicit handling of rows that we fail to insert?  I ask cuz it looks like we're performing a potentially significant number of operations under this transaction, but if we need for this to be all-or-nothing then I guess that doesn't really matter

JordanGuinn (approved):

.platform/nginx/conf.d/https.conf
```diff
@@ -11,12 +11,17 @@ server {
...
+
     location / {
         proxy_set_header X-Real-IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header Host $http_host;
         proxy_set_header X-NginX-Proxy true;
         proxy_pass http://127.0.0.1:8081;
         proxy_redirect off;
+
+				fastcgi_read_timeout 300;
```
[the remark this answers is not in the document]
jlhogan: yes, and also these changes didnt seem to do anything from what i could see. Something somewhere else is sending the 504

src/controllers/enrollmentReports/map/map.ts
jlhogan: parameters per query is the only thing we came across. 
save does indeed result in a "bulk" insert if by bulk insert we just mean insert a bunch of rows in a single `INSERT` statement:
```
server_1          | [Node] INSERT INTO "funding"("enrollmentId", "deletedDate", "fundingSpaceId", "firstReportingPeriodId", "lastReportingPeriodId", "authorId", "updatedAt") OUTPUT INSERTED."id", INSERTED."deletedDate", INSERTED."updatedAt" INTO @OutputTable VALUES (@0, DEFAULT, @1, @2, DEFAULT, @3, DEFAULT), (@4, DEFAULT, @5, @6, @7, @8, DEFAULT), (@9, DEFAULT, @10, @11, DEFAULT, @12, DEFAULT), (@13, DEFAULT, @14, @15, @16, @17, DEFAULT), (@18, DEFAULT, @19, @20, DEFAULT, @21, DEFAULT), (@22, DEFAULT, @23, @24, DEFAULT, @25, DEFAULT), (@26, DEFAULT, @27, @28, DEFAULT, @29, DEFAULT), (@30, DEFAULT, @31, @32, @33, @34, DEFAULT), (@35, DEFAULT, @36, @37, DEFAULT, @38, DEFAULT), (@39, DEFAULT, @40, @41, @42, @43, DEFAULT), (@44, DEFAULT, @45, @46, DEFAULT, @47, DEFAULT), (@48, DEFAULT, @49, @50, DEFAULT, @51, DEFAULT), (@52, DEFAULT, @53, @54, DEFAULT, @55, DEFAULT), (@56, DEFAULT, @57, @58, @59, @60, DEFAULT), (@61, DEFAULT, @62, @63, DEFAULT, @64, DEFAULT), (@65, DEFAULT, @66, @67, DEFAULT, @68, DEFAULT), (@69, DEFAULT, @70, @71, DEFAULT, @72, DEFAULT), (@73, DEFAULT, @74, @75, DEFAULT, @76, DEFAULT), (@77, DEFAULT, @78, @79, @80, @81, DEFAULT), (@82, DEFAULT, @83, @84, DEFAULT, @85, DEFAULT), (@86, DEFAULT, @87, @88, DEFAULT, @89, DEFAULT), (@90, DEFAULT, @91, @92, DEFAULT, @93, DEFAULT), (@94, DEFAULT, @95, @96, DEFAULT, @97, DEFAULT), (@98, DEFAULT, @99, @100, @101, @102, DEFAULT), (@103, DEFAULT, @104, @105, DEFAULT, @106, DEFAULT), (@107, DEFAULT, @108, @109, @110, @111, DEFAULT), (@112, DEFAULT, @113, @114, DEFAULT, @115, DEFAULT), (@116, DEFAULT, @117, @118, @119, @120, DEFAULT), (@121, DEFAULT, @122, @123, DEFAULT, @124, DEFAULT), (@125, DEFAULT, @126, @127, @128, @129, DEFAULT);
```

i think the "implict" limit to this is just the allowed number of params.

src/routes/enrollmentReports.ts
jlhogan: no no this route is very much useful and functional; it is how the enrollment reports supplied by users are uploaded into the system. we just don't use the resulting `enrollmentreport` entity or it's returned id in any way. 

it's possble it could become useful in the future when we're dealing with reconciling uploaded reports with existing data, and want to show something about that reconciliation to the user, but it doesn't feel beneficial to solve for that right now given that that's just a possibility. EnrollmentReport really just shouldn't exist.

src/controllers/enrollmentReports/map/map.ts
jlhogan: b/c the insert is a single query, if inserting one entity fails then inserting all of them fails; so by leveraging the "bulk" insert we're kind of locked in to an all or nothing approach here. 

We could be more defnesive and rigorous in our checks of each entity as we're creating them to ensure they'll be create-able, and then save aside a list of those that we don't think will successfully be added to DB, not include them in the bulk upload, and report back on them in the response. but that's a problem for another day, IMO

JordanGuinn: That makes sense.  The same probably applies for the series of inserts made within any one of these transactions as well, since all the data points kinda depend on each other.  Like if the `Family` and `IncomeDetermination` inserts succeed, but then the `Child` insert fails, we prolly don't want to commit any of it.

jlhogan: yerp yerp. i think the thing to do is try harder to catch things that will cause write fails ahead of write

src/controllers/enrollmentReports/map/map.ts
JordanGuinn: Ah okay, I was confused.  SQL Server has a literal `BULK_INSERT` command, but it looks like that expects an actual file instead, so that wouldn't be what's used here.  So if it's just a big ol' `INSERT` thennnnnnn we're chillin

jlhogan: yup yup, i dont think we're at a point with max couple thou records where we need to get into file-based efficient inserts :D

JordanGuinn: HEH FULLY AGREED

PR #1025, opened 2021-01-12
1024: Fixes soft deletes

### LINKED ISSUE #1024: Soft deletes do not cascade

jlhogan:
closes #1024


bamader (approved):
Did we ever figure out why the soft deletes only worked for some of the entities (I think you'd said enrollments/fundings)?

mmazanec22 (approved):

jlhogan:
> Did we ever figure out why the soft deletes only worked for some of the entities (I think you'd said enrollments/fundings)?

because the enrollment -> funding relationship was defined with `cascade: true`, which covers the cascade of soft deletes. By defining `cascade: ['soft-remove']` on the other relationships (child -> enrollment and family -> income det) we enable them there as well

PR #1037, opened 2021-01-13
Moar home page

### LINKED ISSUE #981: Home page update - Show funding space types before a user has verified data
Value: Lets users verify that we've added the correct sites and funding space combinations

Screen Shot 2021-01-07 at 3 26 03 PM

- Add a slim alert with a link to a form created in #1005  for users to suggest revisions to their data
- Do not show the funding spaces for multi-org users
- Use funding space cards and formatting from #875 
- Hide SR Capacity (we're not as confident in those numbers and they change often.
- Change intro copy to 
  - "ECE Reporter will allow you to share your state-funded enrollment data with the Office of Early Childhood (OEC). Make sure the site names and funding spaces below are accurate before you start entering data.
- Add a header with font-size of 26px: "Your managed sites and funding spaces"
- Add a "sites" header: 22px 
- Add a "funding" header: 22px
- divider line between funding info and How to use ECE Reporter section. 2rem margin-top and bottom on the divider
- Increase "How to use ECE Reporter" to 26px

Note #1028 to hide SR space totals

### LINKED ISSUE #1005: Home page: Funding and Site revision form
I can suggest revisions to my site names and funding space types so that I can successfully submit data.
[Design here](https://share.goabstract.com/1e96b7f9-917a-4471-bbda-d42d46a63674)

- From the home page, I can click on a link in a slim info alert that takes me to this form to request changes to my ECE Reporter data.
- I can request site name changes
- I can request a site be removed from my management
- I can request a new site
- I can indicate additional funding space types my program should have access to (example: We have Part-time CDC in ECE Reporter but we also need full-time)
- I can indicate funding space types my program does not need access to

- Skylight can see the email and organization info of the user making the request to send follow ups to

Screen Shot 2021-01-07 at 5 55 31 PM

- After submitting a request, the user sees a confirmation message:
Screen Shot 2021-01-07 at 5 55 41 PM

### LINKED ISSUE #1036: Remove margin-top on funding cards

bamader:
Creates a funding spaces and sites revision form accessible from the pre-submit home page. Uses state to maintain a copy of the changes a user would like to make to their sites and/or funding spaces, and stores these changes across everal fields in a DB table specifically stood up to hold revision requests. Adds a snapshot test and integration tests for this form to ensure that the API works correctly. Also shows funding spaces to users before they have submitted their data.

Closes #1005 
Closes #981 
Closes #1036


client/src/containers/Home/RevisionRequest.tsx
```diff
@@ -0,0 +1,311 @@
...
+                <b>Registry ID: </b> {` ${s.registryId || 'None'} `}
+              </p>
+            </div>
+            <div className="tablet:grid-col-3 display-flex flex-align-center">
+              <TextInput
+                label=""
+                id={`${s.siteName}-rename`}
+                type="input"
+                onChange={(e) => {
+                  const changeReq = `CHANGE ${s.siteName} TO ${e.target.value}`;
```
jlhogan, line 91: can this be a json object with site id and new name like:
`{ ${s.id}: ${e.target.value} }`

client/src/containers/Home/RevisionRequest.tsx
```diff
@@ -0,0 +1,311 @@
...
+      </p>
+      <h2>Managed sites</h2>
+      {siteSection}
+      <div className="tablet:grid-col-6">
+        <h2>Request a new site</h2>
+        <p className="usa-hint">
+          Fill out this section if your program manages a site not included in
+          your ECE Reporter account.
+        </p>
+        <TextInput
```
jlhogan, line 192: what if a user wants to add more sites? let's maybe abstract this out into an object like 
```
{
  siteName: string;
  licenseId?: string;
  naeycId?: string;
  registryId?: string;
}
```
and then in state, store a list of these objects and in the api post body, send a list of these objects, and then in the backend service maybe we can have 2 tables one for site change requests and one for new site change requests? 

also I dont know that there's a point in having the checkboxes (license exempt, new - no naeyc id, etc) since we're not doing anything with that info. I imagine helper text that just says "leave empty if license exempt" or "leave empty if site does not have naeyc id because it is new, or because it is only head-start funded (I assume that's what that box means @RyanHansz  ? )

client/src/containers/Home/RevisionRequest.tsx
```diff
@@ -0,0 +1,311 @@
...
+          }}
+        />
+      </div>
+      <div className="tablet:grid-col-6 margin-top-4 margin-bottom-4">
+        <Divider />
+      </div>
+      <div className="tablet:grid-col-6">
+        <h2>Funding space types</h2>
+        <p className="usa-hint">
+          Request correction to the funding types received by your program in
```
jlhogan, line 281: it's not clear to me how the user is supposed to use this part of the form -- do they need to check all of the funding space types they have, including those which already correctly show up in the tool, or only those which they don't yet have? What if they want to get rid of a funding type? Not sure what the easiest way to move forward here is; maybe just two checkboxes next to each funding space type with headers above them that indicate one check box is to add and one is to remove? Or we could separate the funding spaces they have from the ones they dont, and allow them to indicate if they want to remove any of the ones they have, or add any of the ones they dont ? @RyanHansz

jlhogan (commented):
Can you please separate these into separate PRs for separate issues? that way, they won't block one another from going live. Now that we have active users, being able to push changes as soon as their ready is all the more important! and smaller PRs are easier to review. If working on multiple branches at the same time feels funky, grab some time with me and we can do some git workshopping to figure out workflows that feel good.

client/src/containers/Home/RevisionRequest.tsx
RyanHansz: A checkbox next to a funding type means the organization receives it, and any currently received funding type should have the box checked when the user opens the form. 

Let's update the helper text here to read:
"Check marks show the funding space types your program receives. Add checks to show other funding types recieved by your program. Uncheck items if your program does not receive that funding type."

jlhogan: ahhhh that makes mucho sense! @bamader may be easier to implement that if you pull the funding space type stuff from https://github.com/ctoec/data-collection/blob/base/client/src/shared/constants/FundingSourceTimes.ts as opposed to redefining them as plain string here

bamader:
@jlhogan Yes can definitely separate these into separate PRs! I had them grouped together originally because the UI changes to the Presubmit Home page included an alert that redirected to the change request form, but I think the UI change PR will be easier to land and so we can definitely do that one first. Should I just leave the Info alert that prompts users to the Revisions page off for now, and then add that in the PR with the form?

client/src/containers/Home/RevisionRequest.tsx
bamader: Ooooh I really like the suggestions in here. Two kinds of requests for two tables feels like a much cleaner way to handle storing these objects, and that still leaves the underlying data flow easy for us to manipulate on the front end. Also, personally a _big_ fan of getting rid of the checkboxes and just using helper text because they don't feel like they add additional information over just leaving the box blank.

jlhogan:
> Should I just leave the Info alert that prompts users to the Revisions page off for now, and then add that in the PR with the form?

yup!

client/src/containers/Home/RevisionRequest.tsx
bamader: @jlhogan @RyanHansz Question on the "what if a user wants to add more sites": how should we design the UI of the form to enable this?

jlhogan: my gut would be a button at the bottom that says "+ Add another site" which pops an additional site form below the existing one, and appends a new site object to the newSites array in page state

client/src/containers/Home/RevisionRequest.tsx
jlhogan: @RyanHansz your initial explanation/designs/copy seemed very clear to me -- with check boxes pre-checked for funding spaces you have. Then you just check/uncheck accordingly and send the resulting map to the backend. Honestly, I think that's even clearer than the separated one above. I just didnt think of it (thus why i am not a designer!!!!!!) :D :D

bamader: Oh yeah cuz if the spaces you already have are pre-checked, then whatever the state of the checks is after you're done, we can just send that away and it'll always mean "we should have these specific funding spaces"--won't even need to try to parse ADDs/REMOVEs, etc.

jlhogan: @bamader in interest of getting this out sooner rather than later, let's just go with that nad ryan can object later in QA/design review :D

bamader: Got it. Since it doesn't make sense to me to stick a `fundingSpaces` field in either of the request types we have so far (addNew and updateName), I think I'll make a fundingSpaces request type as well. Might be type overkill but I think it'll be the cleanest.

jlhogan: not overkill makes sense to me. Altho you can/should send all of the various change requests in a single API request, which you can then decompose and write into the various tables

client/src/containers/Home/RevisionRequest.tsx
```diff
@@ -0,0 +1,269 @@
...
+            const stringRep = ag + ' - ' + fst.displayName + ' - ' + time.value;
+            fsOptions.push({
+              fundingSpace: stringRep,
+              shouldHave: false,
+            } as ChangeFundingSpaceRequest);
+          });
+        }
+      });
+    });
+    setUserFundingSpaces((o) => fsOptions);
```
jlhogan, line 74: nit: if you're not using the function argument, you can just do the direct set call since it'll be functionally the same thing. that is:
```
setUserFundingSpaces(fsOptions);
```

client/src/containers/Home/RevisionRequest.tsx
```diff
@@ -0,0 +1,269 @@
...
+  }, []);
+
+  // Now fetch all the funding spaces that the user and their org
+  // have access to, and when we generate the funding space
+  // checkboxes on the page, mark all these as true
+  const [fundingCheckboxes, setFundingCheckboxes] = useState<JSX.Element[]>();
```
jlhogan, line 80: this shouldn't need to be it's own state var, right? can't you just use the result of this API call to update the funding space options obejct you created above to have `souldHave = true` for these options?

src/routes/revisions.ts
```diff
@@ -0,0 +1,135 @@
...
+    } catch (err) {
+      if (err instanceof ApiError) throw err;
+      console.error("Couldn't submit revision request: ", err);
+      throw new BadRequestError('Revision request information not saved');
+    }
+  })
+);
+
+/**
+ * GET /revision-request/update-sites/:organizationId
```
jlhogan, line 64: I dont think we should write these routes until we want to use them so that they don't end up being unused code --  i dont really imagine we ever will need to access this data via API when we can just grab it directly from the DB. make sure to leave this code in a commit in a branch somewhere so we can easily re-implement it if neeed, but IMO we should only be writing code we plan to use (i personally violate this rule all the time and have to talk myself back around to remembering it, but at the end of the day i think it's smart practice!)

jlhogan (commented):
Once we remove the unused GET routes for revision requests i think this is good to go!!

src/routes/revisions.ts
JordanGuinn: `we should only be writing code we plan to use` Fully agreed

client/src/containers/Home/RevisionRequest.tsx
bamader: Yeah, that's what the `.then` part of the call does. The problem I was encountering was that if I defined `fundingCheckboxes = getFundingSpaceCheckboxes(...)` outside this `useEffect`, the boxes were getting created but they weren't being pre-checked with spaces the user had. Having them dynamically update via state seemed like an easier solution than iterating through again and trying to modify the checkboxes after creating them.

src/routes/revisions.ts
bamader: That's fair. It'll be in the version history of these commits even after I merge this PR, so I'm fine to just delete this stuff and cut down the integration test.

jlhogan (approved):

JordanGuinn (approved):

client/src/containers/Home/RevisionRequest.tsx
```diff
@@ -0,0 +1,268 @@
...
+                match.shouldHave = true;
+                return o;
+              });
+            }
+          });
+          setFundingCheckboxes(
+            getFundingSpaceCheckboxes(userFundingSpaces, setUserFundingSpaces)
+          );
+        })
+        .catch((err) => {
```
JordanGuinn, line 105: minor nit - if all that we're doing with this caught error is re-throwing it, then should be safe omitting the catch block entirely

PR #1038, opened 2021-01-13
Adds system versioning to user, org, site, funding_space tables

### LINKED ISSUE #1039: Add create/update timestamps to app metadata objects

jlhogan:
this way, we have ability to restore data to old state (very good since process for fixing any of these DB entities is i manually modify em), and timestamps about data creation/update (very good for telling more stories about roll out and go live)

closes #1039


src/migration/1610564061121-AddTemporalTableSupportAppMetdata.ts
```diff
@@ -0,0 +1,37 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+import { addSystemVersioningColumnsToTableQuery } from './queries/addTemporalTableSupport';
+import { dropDefaultConstraint } from './utils/dropDefaultConstraint';
+
+export class AddTemporalTableSupportAppMetdata1610564061121
+  implements MigrationInterface {
+  name = 'AddTemporalTableSupportAppMetdata1610564061121';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
```
jlhogan, line 10: Leftover kruft was removed from models but no migration to clean up was created

src/migration/1610564061121-AddTemporalTableSupportAppMetdata.ts
```diff
@@ -0,0 +1,37 @@
...
+      `ALTER TABLE "child" DROP COLUMN "specialEducationServicesType"`
+    );
+    await dropDefaultConstraint(
```
jlhogan, line 13: Align DB schema with models -- we added this default to make sure we appropriately back-filled the `incomeNotDisclosed` column in prod. In reality, we don't want a false default there; creating the record will null value is a valid (but unacceptable) action. User/UI is responsible for ensuring true/false value is appropriately provided

src/migration/queries/addTemporalTableSupport.ts
```diff
@@ -6,7 +6,7 @@ export const addHistorySchemaQuery = () => `
 
 // from https://docs.microsoft.com/en-us/sql/relational-databases/tables/creating-a-system-versioned-temporal-table?view=sql-server-ver15#adding-versioning-to-non-temporal-tables
 export const addSystemVersioningColumnsToTableQuery = (tableName: string) => `
-	ALTER TABLE ${tableName} ADD
+	ALTER TABLE [${tableName}] ADD
```
jlhogan, line 9: needed for `user` table, b/c user is a protected word in sql server

JordanGuinn: _Of course it is_

src/migration/1610564061121-AddTemporalTableSupportAppMetdata.ts
```diff
@@ -0,0 +1,37 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+import { addSystemVersioningColumnsToTableQuery } from './queries/addTemporalTableSupport';
+import { dropDefaultConstraint } from './utils/dropDefaultConstraint';
+
+export class AddTemporalTableSupportAppMetdata1610564061121
```
JordanGuinn, line 5: nit - looks like there's an `a` missing in `AppMetdata`, both in the class name and the title

src/migration/1610564061121-AddTemporalTableSupportAppMetdata.ts
```diff
@@ -0,0 +1,37 @@
...
+    await queryRunner.query(
+      `ALTER TABLE "child" DROP COLUMN "specialEducationServicesType"`
+    );
+    await dropDefaultConstraint(
+      queryRunner,
+      'income_determination',
+      'incomeNotDisclosed'
+    );
+
+    await queryRunner.query(addSystemVersioningColumnsToTableQuery('user'));
```
JordanGuinn, line 19: Is there no `down` counterpart for the `addSystemVersioningColumnsToTableQuery` function?

src/migration/1610564061121-AddTemporalTableSupportAppMetdata.ts
JordanGuinn: +100000 for DB schema matching ORM relations 🔥

JordanGuinn (approved):

src/migration/1610564061121-AddTemporalTableSupportAppMetdata.ts
jlhogan: not at present. I wrote this and don't see us ever actually wanting to undo it. But if we feel there's a legit use case for having down migrations that actually run (ok, i see after writing that there obviously is), we can make it a todo to add one and update these down migrations.

src/migration/1610564061121-AddTemporalTableSupportAppMetdata.ts
jlhogan: yeahhhhh..... don't know that i care.....

src/migration/1610564061121-AddTemporalTableSupportAppMetdata.ts
bamader: _THIS_ 😎

bamader (approved):

src/migration/1610564061121-AddTemporalTableSupportAppMetdata.ts
JordanGuinn: We should absolutely add a TODO.  I def doubt we'll ever _want_ to undo it, more about having the option if we're ever forced to

jlhogan: done https://github.com/ctoec/data-collection/issues/1056

PR #1057, opened 2021-01-14
Preserve tab nav if no children

mmazanec22:
CVloses #1042


jlhogan (approved):
LGTM!!! -- clicking thru it, my eyes/brain would find it more consistent if the roster actions table was present for empty org roster, but that's a @RyanHansz question.

Also, very dig the env var for multi-org user solution! I would keep that in the docker-compose (contrary to your comment saying to comment it out), but just update the value to false so that it's obvious to others how to set that value.

RyanHansz:
> LGTM!!! -- clicking thru it, my eyes/brain would find it more consistent if the roster actions table was present for empty org roster, but that's a @RyanHansz question.
> 
> Also, very dig the env var for multi-org user solution! I would keep that in the docker-compose (contrary to your comment saying to remove it), but just update the value to false so that it's obvious to others how to set that value.

Yes, we can make the actions table show up

bamader (approved):

JordanGuinn (approved):

PR #1073, opened 2021-01-17
Makes api integration tests reliable

### LINKED ISSUE #945: validation integration tests
tests to ensure validations are all running as expected, implemented as integration tests to get DB in test surface
- [x] child
- [x] family
- [x] income det
- [ ] enrollment
- [ ] funding

### LINKED ISSUE #1072: More reliable API integration tests

jlhogan:
Also, rounds out enrollment and funding validation tests

closes #945
closes #1072

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -44,7 +44,7 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --silent --testPathIgnorePatterns ./*/integrationTests",
-    "test-integration": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --silent --runInBand --testPathPattern ./*/integrationTests",
+    "test-integration": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --silent --runInBand --testTimeout=50000 --testPathPattern ./*/integrationTests",
     "test-local": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen",
     "eject": "react-scripts eject"
   },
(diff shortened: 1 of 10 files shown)


client/src/integrationTests/children.test.ts
```diff
@@ -1,57 +1,64 @@
...
     beforeAll(async () => {
       disableFetchMocks();
       utilMock.getCurrentHost.mockReturnValue(
         process.env.API_TEST_HOST || 'http://localhost:5001'
       );
 
-      const thisUser: User = await apiGet('/users/current', '', TEST_OPTS);
-      ALLOWED_ORGS = thisUser.organizations as Organization[];
+      const user: User = await apiGet('/users/current', '', TEST_OPTS);
+      organization = user?.organizations?.shift() as Organization;
```
bamader, line 33: Any reason we're using `.shift()` here? I saw that in a couple places in the integration test and wasn't sure why we were using it over just accessing the `0th` element. In particular because I think shift removes the item from the array after it returns it, which might create problems for orgPermissions (I wonder if that contributed to the integration tests being flaky).

bamader (approved):

client/src/integrationTests/children.test.ts
jlhogan: for typing reasons -- you can't do `[]` access of an array that might be undefined

client/package.json
```diff
@@ -44,7 +44,7 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --silent --testPathIgnorePatterns ./*/integrationTests",
-    "test-integration": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --silent --runInBand --testPathPattern ./*/integrationTests",
+    "test-integration": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --silent --runInBand --testTimeout=50000 --testPathPattern ./*/integrationTests",
```
JordanGuinn, line 47: What's the timeout for?

JordanGuinn (approved):

client/package.json
jlhogan: saw a failure due to async callback not invoked within timeout, and couldn't track down any failures that would be associated with it / could explain it so figured it couldn't hurt to just bump the timeout, esp since these run against deployed env over real network etc etc

PR #1085, opened 2021-01-20
Remove acronyms from funding sources

### LINKED ISSUE #635: Remove the acronym for funding source in text displayed in the app.
Example: Child Day Care instead of CDC – Child Day Care

- When adding or editing enrollment and funding
- In the roster headers 
- Child row
- Remove acronym, only show spelled out funding type name (Child day care)
- Do not change accepted data from the file upload template, just how the text displays in the app

JordanGuinn:
Closes #635.

### DIFF
--- a/src/template/constants.ts
+++ b/src/template/constants.ts
@@ -4,8 +4,8 @@ import moment from 'moment';
 // UPDATE THESE IF YOU CHANGE THIS FILE!!
 // TODO: add pre-merge hook that updates these values (probably put them in a diff file)
 // automatically when this file gets updated
-export const TEMPLATE_VERSION = 2;
-export const TEMPLATE_LAST_UPDATED = moment.utc('01-19-2021', ['MM-DD-YYYY']);
+export const TEMPLATE_VERSION = 1;
+export const TEMPLATE_LAST_UPDATED = moment.utc('01-04-2021', ['MM-DD-YYYY']);
 
 // Formats
 export const BOOLEAN_FORMATS = ['Yes', 'Y', 'No', 'N', 'Not collected'];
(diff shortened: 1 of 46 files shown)


bamader (approved):

src/template/constants.ts
```diff
@@ -47,3 +48,12 @@ export const DEMOGRAPHIC_REPORTING_REASON = 'Used for demographic reporting';
 export const GEOGRAPHIC_REPORTING_REASON = 'Used for geographic reporting';
 export const UTILIZATION_REPORTING_REASON =
   'Used for reporting on the utilization of state funded spaces.';
+
+export const FUNDING_SOURCES_WITH_ACRONYMS: {[ key in keyof typeof FundingSource ]: string} = {
```
jlhogan, line 52: why this approach instead of using a formatter that trims everything before the `-` when displaying ?

JordanGuinn: It seems like the only place we want the acronym to actually exist in the funding source is in data requirements and the template data downloads, so I wanted the enum to be reflective of how it's seen most of the time.  That being said, this enum change I made is really screwing with the `mapEnum` utility, so I might just end up doing that anyway.

jlhogan: nah that makes a lot of sense to me -- might make sense to just write a special funding space specific mapping func instead of trying to shoehorn it into generic `mapEnum` with that yucky flag anyway

JordanGuinn:
Trying to do this the appropriate way has blown up in scope and broken all the things, so I'm gonna close this PR and just forcibly trim the funding sources where they need to be trimmed.  yolo

PR #1092, opened 2021-01-21
1039: Actually turn on temporal table support for metadata entities

jlhogan:
:( whoops


JordanGuinn (changes requested):
Looks good to me, once the `down` block gets filled in.

bamader (approved):

JordanGuinn (approved):

PR #1094, opened 2021-01-21
Fix upload parsing error

bamader:
Fixes a parsing bug where conditional validations would crash on enrollments that tried to parse fundings with errors. I can't be sure this is what caused the specific user error in the ticket, but the log output from the same problem that de-railed a qa e2e test makes it seem like this should do the trick.

As far as adding a timestamp to the file name, I tried changing that in a couple places and got `cannot change _file.name because it is a read-only property`, so I'm not sure where the best place to try to add a timestamp is. Persisting to disk is handled here, but any advice on where in the code is the right place to append a timestamp?

Associated Ticket: https://github.com/ctoec/component-library/issues/121

### DIFF
--- a/src/entity/Enrollment.ts
+++ b/src/entity/Enrollment.ts
@@ -93,7 +93,7 @@ export class Enrollment implements EnrollmentInterface {
   @ValidateIf((enrollment) => {
     // This is used by each funding
     // See similar pattern between child and family
-    enrollment.fundings.forEach((funding) => {
+    enrollment.fundings?.forEach((funding) => {
       // Fundings must be undefined or we have circular json
       funding.enrollment = { ...enrollment, fundings: undefined };
       // But we still need access to the other fundings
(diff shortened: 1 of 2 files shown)


jlhogan:
> As far as adding a timestamp to the file name

I believe this is something you'll do via the multer middleware configuration, by defining a storage option with filename func https://www.npmjs.com/package/multer#diskstorage

jlhogan (approved):

bamader:
Awesome, thanks Julia! Was trying to change the file `fieldname` so that's what was causing my problem.

src/entity/Enrollment.ts
```diff
@@ -93,7 +93,7 @@ export class Enrollment implements EnrollmentInterface {
   @ValidateIf((enrollment) => {
     // This is used by each funding
     // See similar pattern between child and family
-    enrollment.fundings.forEach((funding) => {
+    enrollment.fundings?.forEach((funding) => {
```
JordanGuinn, line 96: 😎👏🏽

JordanGuinn (approved):

PR #1098, opened 2021-01-22
1096: ensures enrollment report for site-level user will only overwrite their sites

jlhogan:
## Background
Bug revealed by user's data disappearing!

## Zenhub Issue
https://app.zenhub.com/workspaces/ece-dev-board-5ff506028d35f30012e0e937/issues/ctoec/data-collection/1096

## Test Coverage
- [ ] Unit tests have been added to cover this changeset. - help, not sure how to test this
- [ ] e2e tests have been added to cover this changeset. - help, not sure how to test this

## Manual Validation
on my local i:
- uploaded data for a bunch of orgs as an org-level user
- switched to site-level user with only subset of site permissions
- uploaded data, choosing to replace data for "all sites"
- switched back to org-level user and confirmed data for other sites (that site user could not access) was untouched

### DIFF
--- a/src/controllers/enrollmentReports/removeExisting.ts
+++ b/src/controllers/enrollmentReports/removeExisting.ts
@@ -32,16 +32,24 @@ export const removeExistingEnrollmentDataForUser = async (
   if (!allChildren || !allChildren.length) return;
 
   const childrenToDelete =
-    siteIdsToReplace && siteIdsToReplace.length
+    // If specific site ids are supplied in API call, remove data for those sites
+    siteIdsToReplace?.length
       ? allChildren.filter(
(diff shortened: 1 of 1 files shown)


JordanGuinn (approved):
LGTM - What does validation of this ticket look like?

bamader (approved):
As far as testing goes, I'm having trouble thinking of a way to test this without having a site level user account. We could try writing an integration test in which we define a site-level user and make sure they can only overwrite a subset of sites?

jlhogan:
Yeah, what i think this reveals is a larger hole in our testing strategy. May be time to bite the bullet and set up a DB for integration tests.

PR #1120, opened 2021-01-25
Parse blank as unavailable

bamader:
## Background
Users were experiencing errors when uploading birth certificate types that were blank even though the field is listed as optional.

## GitHub Issue
https://app.zenhub.com/workspaces/ece-dev-board-5ff506028d35f30012e0e937/issues/ctoec/data-collection/1090

## Validation Plan
Tested with the following file, which checks that every type of birth certificate option is successfully parsed (all valid enum choices + blank).
[Certificate_Type_Test.xlsx](https://github.com/ctoec/data-collection/files/5868747/Certificate_Type_Test.xlsx)

## Automated Testing
- [x] If applicable, unit tests have been added to cover this changeset (backend parsing test added to `src/controllers/enrollmentReport/map/map.test.ts`).

### DIFF
--- a/src/controllers/enrollmentReports/map/map.test.ts
+++ b/src/controllers/enrollmentReports/map/map.test.ts
@@ -133,11 +133,19 @@ describe('controllers', () => {
         (inputBirthCertificateType, expectedBirthCertificateType) => {
           const parsed = mapEnum(
             BirthCertificateType,
-            inputBirthCertificateType
+            inputBirthCertificateType,
+            { isBirthCertificateType: true }
           );
           expect(parsed).toEqual(expectedBirthCertificateType);
         }
       );
+
+      it('can parse a blank certificate type as Unavailable, *not* as Not  Collected', () => {
+        const parsed = mapEnum(BirthCertificateType, '', {
+          isBirthCertificateType: true,
+        });
+        expect(parsed).toBe('Unavailable');
+      });
 
       it.each([
         ...Object.values(Gender).map((gender) => [gender, gender]),
--- a/src/controllers/enrollmentReports/map/map.test.ts
+++ b/src/controllers/enrollmentReports/map/map.test.ts
@@ -1,5 +1,5 @@
 import { EnrollmentReportRow } from '../../../template';
-import { Organization, Site, Child } from '../../../entity';
+import { Organization, Site, Child, Family } from '../../../entity';
 import { BadRequestError } from '../../../middleware/error/errors';
 import {
   BirthCertificateType,
@@ -19,6 +19,7 @@ import {
   mapFundingTime,
   MISSING_PROVIDER_ERROR,
   isIdentifierMatch,
+  mapChild,
 } from './mapUtils';
 import moment from 'moment';
 
@@ -133,19 +134,11 @@ describe('controllers', () => {
         (inputBirthCertificateType, expectedBirthCertificateType) => {
           const parsed = mapEnum(
             BirthCertificateType,
-            inputBirthCertificateType,
-            { isBirthCertificateType: true }
+            inputBirthCertificateType
           );
           expect(parsed).toEqual(expectedBirthCertificateType);
         }
       );
-
-      it('can parse a blank certificate type as Unavailable, *not* as Not  Collected', () => {
-        const parsed = mapEnum(BirthCertificateType, '', {
-          isBirthCertificateType: true,
-        });
-        expect(parsed).toBe('Unavailable');
-      });
 
       it.each([
         ...Object.values(Gender).map((gender) => [gender, gender]),
--- a/src/controllers/enrollmentReports/map/map.test.ts
+++ b/src/controllers/enrollmentReports/map/map.test.ts
@@ -1,5 +1,5 @@
 import { EnrollmentReportRow } from '../../../template';
-import { Organization, Site, Child, Family } from '../../../entity';
+import { Organization, Site, Child } from '../../../entity';
 import { BadRequestError } from '../../../middleware/error/errors';
 import {
   BirthCertificateType,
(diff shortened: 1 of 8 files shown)


src/controllers/enrollmentReports/map/mapUtils/enum.ts
```diff
@@ -33,13 +33,19 @@ export const mapEnum = <T>(
   value: string | undefined,
   opts: {
     isFundingSource?: boolean;
+    isBirthCertificateType?: boolean;
```
jlhogan, line 36: i dont think this should really go in this util. instead i would put the logic in `mapUtils/child`, as above. Then the test would be of `mapChild` func instead of the `mapEnum` func

src/controllers/enrollmentReports/map/mapUtils/child.ts
```diff
@@ -25,7 +25,8 @@ export const mapChild = (
   // Birth certificate type
   const birthCertificateType: BirthCertificateType = mapEnum(
     BirthCertificateType,
-    source.birthCertificateType
+    source.birthCertificateType,
```
jlhogan, line 28: as mentioned below IMO this makes more sense like
```
const birthCertificateType: BirthCertificateType = !source.birthCertificateType
  ? BirthCertificateType.Unavailable
   : mapEnum(BirthCertificateType, source.birthCertificateType)
```

src/controllers/enrollmentReports/map/mapUtils/enum.ts
bamader: I moved the logic just fine, but testing mapChild is giving me a bit of trouble (getting a `ConnectionNotFoundError`). Currently we don't have a test for mapChild because it relies on what I think is a DB connection via its `return getManacer().[other stuf]...` call at the end. We don't have any other tests of code that uses a getManager() in the map.test file, so this one thing is causing a failure. Should we write like a separate test thing for birth certificates? Or do we have to mock it in a different way?

jlhogan: you'll need to mock the DB; seems you can mock the whole of typeorm and just supply an impl for `getConnection().create` since no other typeorm funcs are used

bamader: Were you looking at a specific resource or how-to guide for this? Maybe this is just me being dumb, but I've looked at a lot of github issues and help pages and mocking typeorm is proving difficult. Want to make sure I'm not overcomplicating it.

bamader:
I'm having a lot of trouble getting the typeorm mock to succeed 😕 . In the interest of timeboxing / prioritizing things, are y'all okay if I open a separate ticket about mocking the DB and don't worry about writing an automated test for certificate parsing, since it's a minor change and covered by manual testing at the moment? If this would be an instance of bad dev practices I completely understand and will keep cranking at this, but I'm trying to be more intentional with asking for clarifications.

JordanGuinn:
I'm personally okay with that in this circumstance, but I am curious to know what the issues are with mocking TypeORM objects and methods (as eventually we'll want to make that happen)

bamader:
As a preface, this could again just be me being dumb. But I'd say the biggest difficulty I'm experiencing is just mocking an implementation of _only_ the `getManager().create()` function. No matter the syntax I use, it seems like the test is trying to connect to an _actual_ typeorm database and then is running into all kind of weird errors that I haven't seen in any of our other tests (`EntityMetadataNotFound`, `type error: decorator is not a function`, etc.). I've tried mocking the connection, the getManager(), the create, combinations therein, using factory overrides for the typeorm object itself, and nothing seems to work. There are a few github issues pages and stack overflow questions that seem to be related, but I can't get their solutions to work for me.

jlhogan:
i think the intuition to timebox this is a smart one! let's definitely track the work to figure out how we can mock typeorm, but since we've manually tested and feel confident, i'm fine going forward with this!

src/controllers/enrollmentReports/map/mapUtils/enum.ts
```diff
@@ -37,7 +37,7 @@ export const mapEnum = <T>(
   } = {}
 ) => {
   // Handle special case of blanks in undefineable boolean columns
-  // right upfront
+  // and the birth certificate type right upfront
```
jlhogan, line 40: this comment is no longer true

src/controllers/enrollmentReports/map/map.test.ts
```diff
@@ -1,5 +1,5 @@
 import { EnrollmentReportRow } from '../../../template';
-import { Organization, Site, Child } from '../../../entity';
+import { Organization, Site, Child, Family } from '../../../entity';
```
jlhogan, line 2: nit: unused import

jlhogan (approved):

jlhogan:
can we squeeze this in here as well? User just surfaced this issue 
```	
     const model: CareModel = !source.model
		? CareModel.Unknown
		: mapEnum(CareModel, source.model);
```

src/controllers/enrollmentReports/map/mapUtils/enrollment.ts
```diff
@@ -16,7 +16,9 @@ export const mapEnrollment = (
   child: Child
 ) => {
   const ageGroup: AgeGroup = mapEnum(AgeGroup, source.ageGroup);
-  const model: CareModel = mapEnum(CareModel, source.model);
+  const model: CareModel = !source.model
```
JordanGuinn, line 19: I know it's super minor, but given this was a late add, was this explicitly tested?

JordanGuinn (approved):
LGTM, assuming unit/e2e tests are both green and both blank parsing situations have been tested 👍🏽

src/controllers/enrollmentReports/map/mapUtils/enrollment.ts
bamader: Yup! I changed up the test file to test each possible kind of care model including a blank field, same as I did with the certificate type.

Issue #1097, opened 2021-01-22
Funding spaces in home page should come from available org funding spaces


jlhogan:
NOT just from funding spaces that have children in them

PR #1125, opened 2021-01-26
1097 org funding

mmazanec22:
## Background
Refactors home page to get all funding spaces for all orgs a user has access to rather than just the ones that are currently used. Diff looks more expansive than it is because some files were changed by prettier.

## GitHub Issue
#1097 

## Associated PRs
None

## Validation Plan
Added typing where there was none so that we can be more confident about data returned from API. Manually verified in UI.

## Automated Testing

## Additional Context
None


client/src/components/Forms/Enrollment/Fields/NewFunding.tsx
```diff
@@ -105,7 +105,11 @@ export const NewFundingField = <
       // The radio buttons only really control what expansions are shown
       // They don't change any form data on their own
       id={`funding-source-${enrollment?.id}`}
-      hint={ !!enrollment?.ageGroup ? undefined : "Choose an age group to see available funding types"}
+      hint={
```
JordanGuinn, line 108: I'm definitely the reason for these prettier diffs 😅🤦🏽‍♂️ My bad

src/controllers/children/get.ts
```diff
@@ -95,80 +94,6 @@ export const getCount = async (user: User) => {
...
-};
-
-/**
- * Function that accumulates the distribution of how all children in
- * an organization are assigned to their various funding spaces
- * across age groups and enrollment times. Handles partitioning for
- * displaying so that the front end just has to spit this back out.
- * @param children
- */
-export const getFundingSpaceMap = async (children: Child[]) => {
```
JordanGuinn, line 120: 👏🏽👏🏽

JordanGuinn (approved):

src/routes/children.ts
```diff
@@ -49,10 +48,7 @@ childrenRouter.get(
 
     // Send back a nice pretty display structure for the user's home
     // page if they've submitted their data
-    if (fundingMap && fundingMap === 'true') {
-      const fundingSpacesMap = await controller.getFundingSpaceMap(children);
-      res.send({ fundingSpacesMap });
-    } else if (siteMap && siteMap === 'true') {
+    if (siteMap && siteMap === 'true') {
```
jlhogan, line 51: seeing this makes me think we should likely also display _all_ sites in post-submit home page, even those with 0 enrollments. if that feels too big to put into this PR, then can open another ticket.

jlhogan (approved):

Issue #487, opened 2020-10-02
Improved error state for child records with missing enrollments and/or incomes


RyanHansz:
## Background
The error state for child records with missing enrollments or income determinations is pretty vague, and doesn't supply the user with anything meaningful to take action.  We should provide more specific errors in these circumstances.

### When there are no enrollments (current or past) on a child record
- Add `At least one funded enrollment is required for each child record.` Error message and icon below the enrollment and funding header
- Expand add new enrollment form on opening the section
- Change button to say "add new enrollment"
- Remove the `Missing Enrollment` error banner alert, as this error will provide more specificity

### When there are enrollments (current or past) but no funding on a child record
-  Add `At least one funded enrollment is required for each child record.` Error message and icon below the enrollment and funding header
- Remove the `Missing Funding` error banner alert, as this error will provide more specificity
- Show the funding section of the card with error icons in each field
- Change the "start new funding" button link to "add funding info"

### When there are no income determinations (current or past) on a child record
- Add `A recorded income determination is required.` Error message and icon below the income section header
- Expand add new enrollment form on opening the section
- Change button to say "add new enrollment"

### When there are no enrollments (current or past) on a child record
![No Funded Enrollment.png](https://images.zenhubusercontent.com/5f6e38ea084e48ea4bed8071/f8fc39fe-dd42-43ee-ab71-17a067515ef3)
### When there are enrollments (current or past) but no funding on a child record
![No Funding.png](https://images.zenhubusercontent.com/5f6e38ea084e48ea4bed8071/1ad91199-6ff3-4e2d-ac5f-a625f8153839)

### When there are no income determinations (current or past) on a child record
![No Income.png](https://images.zenhubusercontent.com/5f6e38ea084e48ea4bed8071/bb330420-6dc2-4a32-85b6-c7f86c01e50a)

RyanHansz:
Look up accessibility requirements for error/alert/label

bamader:
Based on some design work and looking at our components, I changed two small things about how this ticket looks in implementation:

1) Based on talking to Victoria, I made the alert text for the case where there are enrollments that lack funding information read: `All enrollments must be associated with at least one funding.`
[...]

PR #1130, opened 2021-01-28
Improve enrollment and funding error states

bamader:
## Background
Improves the error states of a variety of validation issues associated with income determinations, enrollments, and fundings. Changes the banner texts for these, as well as adjusts the UI to be more intuitive to what's missing in particular cases. 

## GitHub Issue
#487 

## Validation Plan
To test these changes, I put together a spreadsheet of the possible test cases I could imagine that would be impacted by these form changes. They include records with:
- Normal income determinations that are disclosed
- Normal income determinations that are not disclosed
- No income determinations recorded
- Normal, funded enrollments documented
- No enrollment or funding information at all
- Enrollment information but no funding information
- Normal historical enrollments (all funded)
- Historical enrollments (some funded, some unfunded).
I verified that the expected pattern of form elements appeared in each of these situations, and also checked that once a user corrected each of these cases, the "normal"/expected appearance of the EditRecord page would return.

The sheet I created to test this is: 
[Enrollment_Funding_Test.xlsx](https://github.com/ctoec/data-collection/files/5883978/Enrollment_Funding_Test.xlsx)

## Automated Testing

## Additional Context
To make this work, I also made the pattern by which we create DB objects on the backend consistent (creating them even if the data in them is empty) because this is informative to the front-end about what kind of error state we're in (i.e. whether a user needs to edit or add). We were already doing this for most of the objects on the backend, so I just extended it to Fundings as well. I think all other context should be covered by comments I documented throughout the change set.


client/src/containers/EditRecord/EnrollmentFunding/Form.tsx
```diff
@@ -20,6 +21,19 @@ export const EnrollmentFundingForm: React.FC<RecordFormProps> = ({
   // Separate enrollments into current (no end date) and past
   // (with end date). Either may not exist
   const enrollments: Enrollment[] = child.enrollments || [];
+  const allEnrollmentsFunded = enrollments.every(
+    (e) => e.fundings !== undefined && e.fundings.length > 0
+  );
+
+  // We create enrollments in the backend even if there's no data
```
jlhogan, line 28: will this be true for enrollments created via the in-app create flow? Not sure this can _always_ be counted on to be true, unless we start creating the empty enrollment when we create a child. Why is an empty enrollment necessary, i.e. what does `(so that we can check validation errors)` mean?

client/src/containers/EditRecord/EnrollmentFunding/Form.tsx
```diff
@@ -33,43 +47,82 @@ export const EnrollmentFundingForm: React.FC<RecordFormProps> = ({
...
-          <ChangeFundingCard
-            {...commonProps}
-            enrollment={currentEnrollment}
-            orgId={child.organization.id}
-            // This heading should be nested under current enrollment
-            topHeadingLevel={getNextHeadingLevel(topHeadingLevel, 2)}
-          />
+          {currentEnrollment.fundings?.map(
+            (funding) =>
+              !noRecordedEnrollments && (
```
jlhogan, line 103: why is this check necessary? If there is a current enrollment withfundings, then there is also by definition a recorded enrollment, right?

jlhogan (commented):
I'm not sure I understand why creating empty placeholder objects is necessary. Shouldn't we be able to determine that an array-type property is expected and missing from a validation error on the owning entity? I.e. child will have an `enrollments` validationerror if enrollments is empty? if that's not the case, we should be able to fix up the validations to ensure that it is.

bamader:
@jlhogan Re: placeholder objects. There are three things going on: poor commenting on my part, something I saw while testing, and the ticket AC. There's a TLDR at the bottom if you wanna skip the text wall (but writing this out helps me be precise and organize my thoughts).

1. (the comments) Looking at the code before and after this changeset, it seems like I wrote my comments in the _least_ helpful way possible for this PR 🙄 . Prior to this changeset, the way we'd written the `mapIncomeDet` and `mapEnrollment` functions would have them create entities even if the user supplied minimal data (such as an enrollment having only an orgId and a site, or an income det just having a family). My comments referred to these as blank/data-less when they shouldn't have because these entities are never actually created empty, so that was sloppy on my part and I apologize for the confusion.
2. (the testing thing) I did some testing with object creation on the backend, and it looks like one of the reasons we create objects with incomplete data is for the checkErrors modal. Currently, without creating minimal-data fundings, the specific fields on a funding object (like `reportingPeriod` and `fundingSource`) don't show up in the modal as missing information. I think this is because the IsNotEmpty() validations on fields in the entity schema don't run if there's no entity to run them on. I also believe this is why fields on Enrollments or IncomeDets that are missing info do show in the modal.
3. The AC of the ticket asks for a funding card to be shown with missing info icons in every field for any enrollment that exists but isn't funded. To me, the best way to implement this seemed to just be to create a funding without any specific information, because then the card would show up in EditRecord and it would also get flagged in the modal (without having to show a blank funding card for a non-existent funding).

TLDR: I think I should do the following: delete my comments about empty entities, leave the enrollment and income det backends untouched, leave in the minimal data funding creation I wrote for this PR (lines 92-96 in `.../mapUtils/funding`). This should address the confusion, ensure validations are correct, and create the card like the ticket asks.

Sorry for the text wall 😞 📜

client/src/containers/EditRecord/EnrollmentFunding/Form.tsx
bamader: Not necessarily, no. Because our back end can create entities that have a minimal amount of data (such as an enrollment that has only an orgId, or something like that), it's possible to wind up with a minimal enrollment and a minimal funding that don't have meaningful information but still exist. See my longer comment below.

bamader:
Also, if instead of leaving things as is, we want to clean up the back end and how we're creating objects so that we only create things when data is supplied, we can do that. I just didn't here because that seemed to be tech debt work that was outside the ticket's scope and I think will involve checking a lot more things.

client/src/containers/EditRecord/EnrollmentFunding/Form.tsx
bamader: Good catch. If a user starts the create flow and exits either a) after making the child identifiers but before reaching the enrollment step, or b) after reaching but before saving the enrollment step, there are two other cases where `noRecordedEnrollments` has different conditions. I've modified the conditional and the "New Enrollment" display to account for these (though we might be overhauling stuff, depending on what we do with the backend below)

jlhogan:
I dont necessarily agree that _not_ creating  the empty fundings amounts to tech debt work, however if this makes sense to you and others, and addresses all the issues then i'm fine with it :)

bamader:
Okay comments have been clarified, more cases checked in the conditionals (with some of the logic moved to utils for readability), and objects in the back end are created consistently and acceptably

client/src/components/Forms/FamilyIncome/Form.tsx
```diff
@@ -137,17 +139,24 @@ export const FamilyIncomeForm: React.FC<FamilyIncomeFormProps> = ({
...
         id={`${id}-fieldset`}
         legend={legend}
-        status={(data) =>
-          getValidationStatusForFields(data, incomeDeterminationFields, {
-            message: 'Income determination is required for OEC reporting.',
-          })
-        }
+        status={(data) => {
+          // Surpress fieldset red warnings if we want to hide errors
+          if (errorsHidden) return undefined;
```
jlhogan, line 144: ternary for consistency?

client/src/containers/EditRecord/EnrollmentFunding/ChangeEnrollment/Card.tsx
```diff
@@ -45,6 +47,12 @@ export const ChangeEnrollmentCard: React.FC<ChangeEnrollmentCardProps> = ({
     if (closeCard) setCloseCard(false);
   });
 
+  // Child batch uploaded without meaningful enrollment
+  // Don't show this card because the form will directly open
```
jlhogan, line 51: this comment is confusing -- "Don't show this Change Enrollment card, because a form to add missing enrollment info is already open" ? or something ofthesort

src/controllers/enrollmentReports/map/mapUtils/funding.ts
```diff
@@ -69,7 +69,8 @@ export const mapFunding = (
     );
   }
 
-  // If the user supplied _any_ funding-related fields, create the funding.
+  // If the user supplied _any_ funding-related fields, create
+  // a propagated funding.
```
jlhogan, line 73: propagated? perhaps populated?

jlhogan (approved):
Thoughts, not for this PR just generally:
- should hide error be a context as opposed to passed prop?

JordanGuinn (approved):

Issue #1012, opened 2021-01-08
Large roster counts are inaccurate until all records have been loaded


RyanHansz:
## Background
Currently, when my roster has many records, the total enrollment numbers and number of records with missing info continually update until all records are loaded.

## Acceptance Criteria
- **If this isn't slow**, the enrollment list response includes a total count of records within and is displayed on the page
- **If the above is slow**, there is some sort of loading indicator blocking the record count until that information is available
- The totals persist instead of being reloaded every time I load my roster

## Designs
Current state:

https://user-images.githubusercontent.com/63253539/104027782-e2ae9000-5195-11eb-85c5-858417841700.mp4

jlhogan:
A nice-to-have i'd been thinking about would be updating the way the alerts hook works to not ALWAYS scroll to top of window when there are alerts https://github.com/ctoec/data-collection/blob/base/client/src/hooks/useAlerts.tsx#L23, as this results in the user beingpulled to top of page on every API page load

(not sure what the fix here should be, and if nothing ends up being obvious/easy, then this can abso be considered out of scope. But i imagine could be achieved by wrapping the provided `setAlerts` func in a wrapper func, moving the call to `window.scrollTo` into there, and maybe  taking an additional flag as a param in wrapper func to indicate when to skip the scroll? -- that's just an idea off top of my head. dunno.)

JordanGuinn:
Closing and holding on, in favor of the upcoming roster redesign work.

PR #1136, opened 2021-01-29
Fix incorrect child record totals on roster page

JordanGuinn:
## Background
The total child record numbers on the roster page are currently tied to the last page the user has visited, and not the total number of records returned from the underlying query.  This PR addresses those inaccuracies.

## GitHub Issue
- https://github.com/ctoec/data-collection/issues/1012

## Associated PRs
N/A

## Validation Plan
- [x] All funding spaces are still available in funding space dropdown when changing child's enrollment
- [x] Roster count is accurate for a roster with **less than 100 children**
- [x] Roster data is correct for a roster with **less than 100 children**
- [x] Roster count is accurate for a roster with **greater than 100 children**, both
   - [x] **Before** loading the next page
   - [x] **After** loading the next page
- [x] Roster export operates and includes the same data as it did previously
- [ ] Roster count is accurate when applying the **previous month filter**
- [ ] Roster data is correct when applying the **previous month filter**
- [x] Child count is accurate on the batch edit page for children with missing info
- [x] Child data is correct on the batch edit page for children with missing info

## Automated Testing


jlhogan (approved):

bamader (approved):

vonbearshark:
Checking in on the state of this. I might like to lift the pattern for the active/withdrawn validation errors count. Is this generally a working idea/PR, or a potential pitfall? Also, would you like a hand moving this along? I could even pick it up, code-wise, if there was something to be done still.

JordanGuinn:
@vonbearshark It's a working idea, there's just one outstanding issue with roster counts when applying the `Previous month` filter being inaccurate and I just haven't had the time to address that last piece 😕

And now that I think about it, I imagine is inherently getting addressed in the previous month filter ticket you're working on 😅

vonbearshark:
Ha. Yeah I think I can see how this fits together. I would be happy to just incorp these changes into my PR? Or push to this?

JordanGuinn:
Perf - let's incorporate these changes into your PR then 🙂

vonbearshark:
Closing out this PR, but stay tuned for the future ✨

Issue #1137, opened 2021-01-29
Funding overlap validation bug causing Roster infinite load


jlhogan:
## Background
User-reported roster is stuck in infinite load. Determined cause is 500 response to `children` query caused by:
```
[...]

PR #1138, opened 2021-01-29
Address Funding Space Overlap Bug

JordanGuinn:
## Background
Our funding space overlap validator currently breaks when comparing a given funding with all the other fundings within an organization and **at least one** of those fundings is missing a reporting period start date.  This PR addresses that.

## GitHub Issue
- #1137

## Associated PRs
N/A

## Validation Plan
- [x] Funding Space Overlap validator doesn't include fundings without a reporting period start date
- [x] Rosters containing enrollments with fundings **without** a starting reporting period can be loaded successfully
- [x] Funding Space Overlap validator still functions as expected

## Automated Testing

## Additional Context
N/A

### DIFF
--- a/client/src/containers/Roster/Roster.tsx
+++ b/client/src/containers/Roster/Roster.tsx
@@ -248,6 +248,7 @@ const Roster: React.FC = () => {
 
   return (
     <>
+    <ErrorBoundary alertProps={defaultErrorBoundaryProps}>
       <div className="Roster grid-container">
         <BackButton
           location={
@@ -316,6 +317,7 @@ const Roster: React.FC = () => {
           )}
         </FixedBottomBar>
       )}
+      </ErrorBoundary>
     </>
   );
 };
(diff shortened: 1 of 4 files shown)


jlhogan (approved):

client/src/containers/Roster/Roster.tsx
```diff
@@ -79,11 +79,17 @@ const Roster: React.FC = () => {
...
 
-  // TODO: handle fetching error
-  const {
+  let {
     children: allChildren,
     stillFetching: loading,
+    error: fetchRosterError
   } = usePaginatedChildData(query);
+
+  if (fetchRosterError) {
```
mmazanec22, line 88: We should probably just change usePaginatedChildData to set loading to false and throw this error

client/src/containers/Roster/Roster.tsx
```diff
@@ -242,6 +248,7 @@ const Roster: React.FC = () => {
 
   return (
     <>
+    <ErrorBoundary alertProps={defaultErrorBoundaryProps}>
```
mmazanec22, line 251: There's already an error boundary around each rendered container component (see App.tsx), so I'm not sure this one does much for us. Also, if we do keep this error boundary here, then we don't need the fragment on lines 250 and 320.

Issue #1128, opened 2021-01-27
Confirm optionality of fields in batch upload


jlhogan:
## Background
For the sake of a consistent user experience, we want to make sure that all of the optional fields within our batch upload UI are in fact optional at the API and/or data layer.

## Acceptance Criteria
- All fields marked as **OPTIONAL** in data requirements table can be left empty in upload without causing errors
  - If any fields are **conditionally** required based on the value of another field, those should be enumerated and confirmed as well
  - Birth certificate ID number, town of birth and birth state are required if birth certificate type is US
- e2e test(s) exist to confirm this for **each and every applicable field**

JordanGuinn:
@bamader Looks like the e2e suite is failing

bamader:
@JordanGuinn Oh crap this is the really dumb exit 10 failure where none of the tests actually run and we get no helpful log output 😑 . That means it's some super small nuance bug. Into the mines I go...

JordanGuinn:
Is there any more output when the `--verbose` flag is added to the nightwatch command running the tests?

bamader:
I managed to diagnose it. `exit code 10` has thus far referred to finicky module stuff (importing, exporting, ts vs js, etc.) so that was at least somewhat of a clue. I'm putting details in the Readme.

JordanGuinn:
@bamader Created a ticket for the bug we found earlier.  Lemme know if I'm missing anything:

https://app.zenhub.com/workspaces/ece-dev-board-5ff506028d35f30012e0e937/issues/ctoec/data-collection/1157

bamader:
@JordanGuinn I made a few updates just to give some extra details, but it looks good! Realistically I'll probably handle it first thing tonight/tomorrow once the audit PR lands so we won't have to wait long 😛

PR #1139, opened 2021-01-29
Audit optional and conditional fields with new tests

bamader:
## Background
We currently have a number of fields marked as either optional or conditional, and we don't have automated verification that these fields behave as expected when they're missing. This PR adds that confirmation by building export functionality for sets of records without optional/conditional fields and then expanding upload tests to include these new data sources.

## GitHub Issue
#1128 

## Validation Plan
Because there was some changing with parsings and validations, I made sure to use the `DevUtility` page to download every version of every file we support, upload them to the tool, and make sure children had the expected number of errors in the expected fields. More spreadsheet uploading and testing of creative cases by any reviewers would be appreciated just to make sure we cover as many upload cases as we can think of. I've added links to download data sets with records missing optional fields and records missing conditional fields to the Dev page for this purpose.

## Automated Testing
- [ ] If applicable, e2e tests have been added to cover this changeset (two new e2e tests were added based off the `upload` model; these tests verify expected behavior of optional fields [which, if they're missing from a record, should cause _no_ errors at all] and conditional fields [which, if they are expected to be present because their conditional is satisfied and they are not, _should_ cause errors]).

## Additional Context
In the process of this PR, I found that we had some inconsistencies in our backend parsing logic that had given rise to a few bugs, so I also tried to correct those here and keep us consistent/robust. They were:
1) `incomeNotDisclosed` wasn't parsing as a boolean because it was set to default to `undefined` rather than false, which means that none of the validations or error modal behavior was catching income issues
2) Our `export` controller wasn't accounting for children with multiple fundings for an enrollment; so in any export we did, if an enrollment had multiple fundings, only the first was appearing as a row
3) We didn't have a validation error for the case where a previous/historical enrollment either didn't' have an exit date, or the exit date was after a child had started a newer enrollment; that was causing wonkiness with conditional tests on `CHANGE_ENROLLMENT` so it's now a requirement that current enrollments must start on or after the exit dates of past enrollments.


JordanGuinn (approved):

src/controllers/export.ts
```diff
@@ -26,16 +26,28 @@ export async function streamUploadedChildren(
...
-        // Can just use 0th element of each array as the 'active'/'current'
-        // value because controller.getChildById does presorting for us
-        childStrings.push(
-          flattenChild(columnMetadatas, child, {
-            enrollment,
-            skipInfoForPastEnrollments: i !== 0,
-          })
-        )
-      );
+      child.enrollments?.forEach((enrollment, i) => {
```
jlhogan, line 29: prefer `enrollmentIdx` and `fundingIdx` over `i` and `j`

src/entity/decorators/Child/enrollmentDatesValidation.ts
```diff
@@ -0,0 +1,36 @@
...
+      propertyName: propertyName,
+      options: { message: enrollmentDatesCannotOverlap, ...validationOptions },
+      validator: {
+        async validate(_, { object: child }) {
+          const enrollments = await getManager().find(Enrollment, {
+            where: {
+              childId: (child as Child).id,
+            },
+          });
+          const currentEnrollment = enrollments.find((e) => !e.exit);
```
jlhogan, line 25: make sure this validation does not fail if child has no enrollments nor if enrollments are missing both/either entry and exit dates!

src/template/constants.ts
```diff
@@ -30,7 +30,8 @@ export const REPORTING_PERIOD_FORMATS = [
 ];
 
 // Conditional requirement strings
-export const REQUIRED_IF_US_BORN = 'Required if child was born in the U.S.';
+export const REQUIRED_IF_US_BORN =
+  'Required if the "Birth Certificate Type" field is "US Birth Certificate."';
```
jlhogan, line 34: why this change?

jlhogan (approved):

client/src/containers/DevUtility/DevUtility.tsx
```diff
@@ -62,6 +62,36 @@ const DevUtility: React.FC = () => {
             queryParamsAsObject={{ whichFakeChildren: 'missingOne' }}
             downloadText="Download CSV of all children missing one field"
           />
+          <h3>Children Set Missing All Optional Fields</h3>
+          <CSVExcelDownloadButton
+            fileType="xlsx"
+            whichDownload="example"
+            className="margin-bottom-3"
+            queryParamsAsObject={{ whichFakeChildren: 'missingOptional' }}
```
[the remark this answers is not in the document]
bamader: Sure, an enum seems like a natural choice to handle this

e2e-tests/tests/uploadMissingOptional.js
```diff
@@ -0,0 +1,14 @@
...
+const { uploadFile } = require('../utils/uploadFile');
+
+module.exports = {
+  '@tags': ['upload'],
+  uploadMissingOptional: async function (browser) {
+    await browser.init();
+    await browser.timeoutsImplicitWait(10000);
+    await login(browser);
+    await uploadFile(browser, 'missingOptional');
+    // TODO: expect missing info icons in roster
```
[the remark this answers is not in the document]
bamader: That, my friend, is a piece of kruft from an age long passed (more specifically, before we were just pulling the number of children with missing info from the batch edit banner alert--the comment is superfluous now)

src/template/constants.ts
bamader: To me it was just a slightly clearer choice of language. I thought there could be parents with children born in the US who may not have the birth certificate available, and so would still try to pick US Birth Certificate (which would then throw errors). That may be desired, though, and if this change doesn't actually clarify anything I can switch it back.

src/entity/decorators/Child/enrollmentDatesValidation.ts
bamader: Ooooo good catch, since we've def had issues with validations crashing like that before

Issue #657, opened 2020-11-03
Confirmation alert on adding all information to a single child batch edit.


RyanHansz:
When returning to the roster after adding all information to a single child in a batch edit flow (i.e. when you click on the Missing Info icon for a single record), display a confirmation alert that says:
"Changes saved
[Child's name]'s record now has all the required information."

JordanGuinn:
## Testing Strategy
### General
- [x] All unit tests are passing
- [ ] All e2e tests are passing

### Single Child Batch Edit Flow
- [ ] Batch edit can still be completed successfully
- [ ] On completion, a confirmation alert is displayed
- [ ] The confirmation alert contains the text specified
- [ ] Upon dismissal of the confirmation alert, the user is taken back to their roster

### Multi-Child Batch Edit Flow
- [ ] Batch edit can still be completed successfully
- [ ] On completion, the user is moved to the next record in the batch

JordanGuinn:
@bamader Looks good overall, just a couple things and this'll be ready to go:

1. The new single child confirmation alert is only shown when the user clicks on the `Missing Info` icon in the corresponding child row in the roster, and **not** when the user clicks on link in the `Add missing info to complete your enrollment data` text in the alert at the top of the roster view.  Is this desired?
2. If the child's first name is the field that's missing, the alert displayed reads `'s record now has all required information.`

bamader:
@JordanGuinn 
1. Yep this is the desired behavior. The "single record batch edit" flow is accessible by clicking the missing info icon in the corresponding row, because single record batch edit is supposed to simulate an Edit Record page that directly opens to the form and that's how we tell BatchEdit to differentiate an array of Children from a childId. Ticket confirms this in the parenthetical clause.
2. That is, however, _not_ desired behavior, so will fix that up!

Issue #273, opened 2020-09-04
Inactive fundings should have gray BG in enrollment/funding section


RyanHansz:
## Acceptance Criteria
- Change label to "inactive funding" and add light-gray background to funding cell BG
- Most recent/active funding should be on the top.

## Designs
<img width="1048" alt="Screen Shot 2020-09-04 at 12 23 34 PM" src="https://user-images.githubusercontent.com/63253539/92262666-7f3fe900-eea9-11ea-8ac4-1cf79689f3df.png">

JordanGuinn:
## Testing Strategy
### General
- [ ] All unit tests are passing
- [ ] All e2e tests are passing

### Current Enrollment
- [ ] Most recent funding for the **current** enrollment is at the top
- [ ] Most recent funding for the **current** enrollment is not greyed out
- [ ] All other fundings for the **current** enrollment are sorted by first reporting period, with most recent closest to the top
- [ ] All other fundings for the **current** enrollment are greyed out

### Previous Enrollments
- [ ] Most recent funding for the **all previous** enrollments are at the top
- [ ] Most recent funding for the **all previous** enrollments are not greyed out
- [ ] All other fundings for **all previous** enrollment are sorted by first reporting period, with most recent closest to the top
- [ ] All other fundings for **all previous** enrollment are greyed out

JordanGuinn:
@bamader Overall things are working as expected, just a couple things:

1. Don't know if this is intended or not, but the actual enrollment panel (containing site/model type/age group info) is _also_ greyed out for all previous enrollments.  I'll include a screenshot to clarify.
2. The border around funding cells is the exact same color as the gray background of inactive fundings, so it the border effectively vanishes for all inactive fundings.  I'll also make this clear in the screenshot.

![image.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/6bc950d8-576d-4e71-aa6f-2fc8e398c8f6)

bamader:
@JordanGuinn 
1. Yep, this is desired! We want all historical/withdrawn enrollment blocks (enrollments + their fundings) to be our `secondary` gray color.
[...]

PR #1143, opened 2021-02-02
Fix up batch edit and funding cards

bamader:
## Background
There are two quick fast follows fixes that involve tweaking up some UI aspects of the tool. This PR sorts funding cards so that the most current one is at the top, grays out the backgrounds of expired funding cards, and pushes an alert to the roster page if a user comes from a single record batch edit.

## GitHub Issue
#273 
#657 

## Validation Plan
I tested the new batch edit functionality with a variety of single record batch edits and confirmed that it works when using the traditional batch edit on multiple records. I got the correct alert state in each case.

## Automated Testing

### DIFF
--- a/client/src/containers/BatchEdit/BatchEditItemContent.tsx
+++ b/client/src/containers/BatchEdit/BatchEditItemContent.tsx
@@ -14,7 +14,7 @@ import { useAlerts } from '../../hooks/useAlerts';
 import { hasValidationErrorForField } from '../../utils/hasValidationError';
 import { apiGet } from '../../utils/api';
 import AuthenticationContext from '../../contexts/AuthenticationContext/AuthenticationContext';
-import { Link } from 'react-router-dom';
+import { Link, useHistory } from 'react-router-dom';
 import { getCurrentEnrollment } from '../../utils/models';
 import { mutateCallback } from 'swr/dist/types';
 import RosterContext from '../../contexts/RosterContext/RosterContext';
@@ -24,16 +24,19 @@ type BatchEditItemContentProps = {
   organizationId?: string;
   moveNextRecord: (_?: Child[]) => void;
   mutate?: (_: mutateCallback<Child[]>, __: boolean) => void;
+  isSingleRecord?: boolean;
 };
 
 export const BatchEditItemContent: React.FC<BatchEditItemContentProps> = ({
   childId,
   moveNextRecord,
   mutate,
+  isSingleRecord,
 }) => {
   const [child, setChild] = useState<Child>();
 
   const { setAlerts } = useAlerts();
+  const history = useHistory();
 
   const [steps, setSteps] = useState<StepProps<RecordFormProps>[]>();
   const [activeStepKey, setActiveStep] = useState<string>();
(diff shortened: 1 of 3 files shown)


JordanGuinn:
Nit: Moving forward, can you use the **GitHub** issue links in the PR templates instead of the **ZenHub** issue links?  For whatever reason, the PR is only associated with the issue in ZenHub when the _GitHub_ issue link is used in the PR (which makes infinite negative sense to me, but 🤷🏽‍♂️😅)

JordanGuinn (approved):

client/src/containers/BatchEdit/BatchEditItemContent.tsx
```diff
@@ -43,6 +46,23 @@ export const BatchEditItemContent: React.FC<BatchEditItemContentProps> = ({
 
   const { updateCurrentRosterCache } = useContext(RosterContext);
 
+  // Special function for single record batch edit because we use the child's
+  // name information in an alert
+  const singleRecordSuccess = () => {
```
jlhogan, line 51: why can't this just be passed in as the `moveNextRecord` function in `BatchEdit.tsx` and we don't need the `isSingleRecord` flag?

client/src/containers/EditRecord/EnrollmentFunding/Form.tsx
```diff
@@ -53,16 +53,26 @@ export const EnrollmentFundingForm: React.FC<RecordFormProps> = ({
...
-              {...commonProps}
-              key={funding.id}
-              isCurrent={true}
-              fundingId={funding.id}
-              enrollmentId={currentEnrollment.id}
-              topHeadingLevel={getNextHeadingLevel(topHeadingLevel, 2)}
-            />
-          ))}
+          {currentEnrollment.fundings
+            ?.sort((a, b) => {
```
jlhogan, line 57: pull this sort func out into a util

client/src/containers/BatchEdit/BatchEditItemContent.tsx
bamader: For a single record batch edit, the `BatchEdit` container doesn't have a copy of the child (none of the `fixedRecordsForDisplay` states or the related things get filled because we don't make an API call if `childId` is defined), so it doesn't have access to the child's name. I thought handing it down was better than adding an extra API call.

bamader:
@JordanGuinn Yeah, definitely can do. Didn't realize the linking would be weird. Sorry about that!

jlhogan (approved):

client/src/containers/BatchEdit/BatchEditItemContent.tsx
mmazanec22: Move next record takes the param records, so you could pass it the one record, and then have it get the child name that way

JordanGuinn:
No worries!  You'd think it'd be something that ZenHub could figure out, but alas 😅

client/src/containers/BatchEdit/BatchEditItemContent.tsx
jlhogan: @bamader is this possible? ^

bamader: Doesn't look like this works. For single record batch edit, we have no need to mutate the cache to update different children, so we never set the records param to be anything other than an empty list (`_records` is set within a conditional dependent on mutate). To me it's simpler to just add a flag for whether it's a single record batch edit rather than invoke cache mutation for one object. Thoughts?

jlhogan: word. fine by me

PR #1146, opened 2021-02-03
Carbon

jlhogan:
## Background
Pulls in updated component lib with carbon components + styles; tweaks things to remain _mostly_ consistent. Moving forward, we should be able to either replace existing components under the hood with carbon if necessary, or pull carbon components directly from the component lib. 

note: divider component is not working, investigating

## GitHub Issue
https://github.com/ctoec/component-library/issues/112

## Associated PRs
https://github.com/ctoec/component-library/pull/126

## Validation Plan
Manual visual confirmation that all pages looks acceptably similar to previous state

## Automated Testing

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -167,7 +167,7 @@ jobs:
       - run:
           working_directory: e2e-tests
           name: Run e2e tests (in chrome)
-          command: yarn test tests/withdrawChild.js
+          command: yarn test
       - run:
           working_directory: client
           name: Run api integration tests
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,7 +58,8 @@ jobs:
           aws_access_key_id: $AWS_ACCESS_KEY_ID
           aws_secret_access_key: $AWS_SECRET_ACCESS_KEY
     environment:
-      NODE_OPTIONS: --max-old-space-size=1536
+      # https://support.circleci.com/hc/en-us/articles/360009208393-How-can-I-increase-the-max-memory-for-Node-
+      NODE_OPTIONS: --max-old-space-size=2048
     steps:
       - attach_workspace:
           at: .
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,6 +64,9 @@ jobs:
       - attach_workspace:
           at: .
       - run:
+          name: print env
+          command: env
+      - run:
           name: Run npm audit for back-end dependencies
           command: yarn audit
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -167,7 +167,7 @@ jobs:
       - run:
           working_directory: e2e-tests
           name: Run e2e tests (in chrome)
-          command: yarn test
+          command: yarn test tests/withdrawChild.js
       - run:
           working_directory: client
           name: Run api integration tests
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,13 +59,11 @@ jobs:
           aws_secret_access_key: $AWS_SECRET_ACCESS_KEY
     environment:
       # https://support.circleci.com/hc/en-us/articles/360009208393-How-can-I-increase-the-max-memory-for-Node-
-      NODE_OPTIONS: --max-old-space-size=2048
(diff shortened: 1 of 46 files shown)


client/src/components/FixedBottomBar/_index.scss
```diff
@@ -2,6 +2,7 @@
   position: fixed;
   bottom: 0;
   width: 100%;
+  z-index: 999999;
```
bamader, line 5: lol what does this do? Seems like a dummy value that's just in there to be large 😂

bamader (approved):
Wasn't so scary to review once you realize that 85% of the changed lines are in snapshots 😝

client/src/components/FixedBottomBar/_index.scss
jlhogan: https://www.w3schools.com/cssref/pr_pos_z-index.asp 
"stack order" ensures this is always on "top" of all other elements, which we want for this overlay button bar. Wasn't an issue until something in the Carbon accordion ended up on top and this seemed like a reasonable fix to my mind

bamader: Yeah makes sense. Not like anything's gonna come in with a z value larger than that.

femmebot: Yeah, CSS is weird. It's likely to ensure the bottom bar remains the top-most element on the page.

RyanHansz:
Looks like a smooth transition! A few things I saw:
- button alignment:
Screen Shot 2021-02-12 at 9 30 32 AM
- Edit record header is meant to be on two lines. The alert and header section (.display-flex .flex-justify) could each use margin-bottom: 1.5rem so they don't run into the other sections. 
Screen Shot 2021-02-12 at 9 29 58 AM
Screen Shot 2021-02-12 at 9 27 48 AM

jlhogan:
@RyanHansz I made those changes; would you mind giving it one more look and then perhaps tossing an approval on here? :D

RyanHansz:
@jlhogan looks good to me!

.circleci/config.yml
```diff
@@ -57,6 +57,8 @@ jobs:
         aws_auth:
           aws_access_key_id: $AWS_ACCESS_KEY_ID
           aws_secret_access_key: $AWS_SECRET_ACCESS_KEY
+    environment:
+      NODE_OPTIONS: --max-old-space-size=1536
```
JordanGuinn, line 61: What's this for?

src/routes/fundingSpaces.ts
```diff
@@ -29,6 +29,7 @@ fundingSpacesRouter.get(
         children
       );
       res.send({ fundingSpacesMap });
+      return;
```
JordanGuinn, line 32: And this?

e2e-tests/utils/waitForHomepage.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  waitForHomepage: async function (browser, enforcePresubmit = false) {},
```
JordanGuinn, line 2: And this empty function?

JordanGuinn (commented):
Seems pretty cut and dry, since we're really just pointing to a different version of `component-library`.  Although I will add that, given the scope here, validation **really** should include a **functional** confirmation of all impacted components, in addition to a visual one.

JordanGuinn (approved):

.circleci/config.yml
jlhogan: tests were failing in circleCI, due to lack of memory in the node process which apparently happens when running it in the containers; will add the circleCI docs I used to resolve in a comment.

e2e-tests/utils/waitForHomepage.js
jlhogan: this should not be here !

src/routes/fundingSpaces.ts
jlhogan: without this return, were trying to send the response again below and getting`headers already sent` errors in the backend.

jlhogan: one word riders that fix other things are alloeed in my book hehehe

Issue #1050, opened 2021-01-14
User getting "booted" from system while entering child records


jlhogan:
From user: 
> I have started putting in my records and have been booted out twice after putting in a couple of records. This is going to take forever if this continues.

JordanGuinn:
Session timeout bumped to 60 mins - pausing on for now to see if that (hopefully) addresses this.

JordanGuinn:
Note to self: There have been a bunch of refresh token validation errors in WingedKeys logs over the past several days, that may correlate to the issue at play here.  Probably worth checking out our IdentityServer [refresh token configuration](https://docs.identityserver.io/en/3.1.0/topics/refresh_tokens.html).

JordanGuinn:
Additional note to self: This appears to happen because at times ECE Reporter is hitting the IdentityServer's `/connect/token` endpoint to refresh the user's access token twice **at the exact same time**.  The first request succeeds, getting the user a new access token _and_ refresh token, then the second request fails, because it uses the now-invalidated refresh token.  And of course that second failure results in us auto-logging the user out.

Now it's just a matter of figuring out what causes the `connect/token` endpoint to be hit twice at the same time.

PR #1148, opened 2021-02-03
Prevent intermittent and sudden session timeouts

JordanGuinn:
## Background
There were a number of different things at play that I attempted to address within this PR:

1. The refresh token buffer we currently use to determine when to request a new access token was set to 10 minutes, which means that a user could potentially sit inactive (read: doesn't leave the page) for **only 11 minutes** and get logged out.  I bumped this to 15 minutes for now, but we may consider revisiting that buffer again.
2. We weren't handling potential URL changes (and thereby potential access token requests) in **rapid succession** of one another, which turns out to be very possible if your browser's event handling is lagging or you're just really quick with the mouse.  That's no longer the case - if we already have an outstanding access token request in-flight while another URL change is made, we'll simply wait on that outstanding request instead of initiating another one.
3. Potentially the most frequent issue - refresh tokens were **simply lost** on page reload.  We're storing the access token in local storage (which is another conversation entirely), but not the refresh token, so when the page is reloaded the refresh token vanishes.  This means that if a user reloaded their page at any point, they would **definitively get logged out** if they remained on the page for the length of the access token expiration, **regardless of their activity**.  So instead we're now storing the refresh token in local storage and not the access token (as a new access token can be retrieved at any time).

## GitHub Issue
- #1050 

### Use Case 1
- [x] Log in to ECE Reporter
- [x] Sparingly use the application for the length of the access token expiration (default is 60 mins)
- [x] Confirm you're not logged out

### Use Case 2
- [x] Log in to ECE Reporter
- [x] Sparingly use the application for 5-10 minutes
- [x] **Refresh the page**
- [x] Sparingly use the application for the length of the access token expiration (default is 60 mins)
- [x] Confirm you're not logged out

## Automated Testing

## Additional Context
N/A


client/src/contexts/AuthenticationContext/AuthenticationContext.tsx
```diff
@@ -180,29 +217,45 @@ const AuthenticationProvider: React.FC<AuthenticationProviderPropsType> = (
...
 
-    // isValid includes a defaut 10 min expiration buffer.
-    if (tokenResponse.isValid()) return;
+    //  Hopefully this never happens - if there's no refresh token present, there's no way to
+    //  generate a new access token, so the user will inevitably get logged out, regardless of activity
+    if (!refreshToken) return;
+
+    const authExpirationBuffer: number = 15 * 60 * -1; // 15 minute expiration buffer
+
+    //  Don't generate new access token if current token doesn't expire within 15 mins
```
jlhogan, line 227: fwiw, `isValid()` has a default 10 min buffer i believe

jlhogan (commented):
validating now!

jlhogan (approved):
I have been unable to repro a boot; let's definitely deploy this to a live env and test again but from my pov: verified!

## Validation Plan
### Use Case 1
- [x] Log in to ECE Reporter
- [x] Sparingly use the application for the length of the access token expiration (default is 60 mins)
- [x] Confirm you're not logged out

### Use Case 2
- [x] Log in to ECE Reporter
- [ ] Sparingly use the application for 5-10 minutes
- [ ] **Refresh the page**
- [ ] Sparingly use the application for the length of the access token expiration (default is 60 mins)
- [ ] Confirm you're not logged out

client/src/contexts/AuthenticationContext/AuthenticationContext.tsx
JordanGuinn: Yep!  It just seemed not long enough

bamader (approved):
Code looks good to me--moving to validate as well!

kmagida110 (approved):
Code looks good to me, trying validation now

bamader:
As with Julia, couldn't repro a booting so something must be right here. This looks good to me to merge!

JordanGuinn:
WOOT - I'll deploy this evening when everyone's logged off.

Issue #1157, opened 2021-02-04
Empty birth state/town values interpreted as "Not Disclosed" in CSV uploads


JordanGuinn:
## Background
There is a difference between how empty values for `birth state` and `birth town` are handled in CSV file uploads vs Excel file uploads.  In Excel file uploads, they're properly translated to simply `undefined` values, whereas in CSV uploads they get parsed as `null`, which is then interpreted as `Not Disclosed` in the app.

## Repro Steps
- Upload a CSV with child rows containing empty `birth state` and `birth town` values
- Upon upload completion, navigate to those children within your roster

## Expected Behavior
- Birth state and birth town are simply blank values.

## Actual Behavior
- The `Not Disclosed` checkbox is selected for both birth state and birth town

## Additional Context
Given that this came up as part of the CSV upload but **not** the Excel upload, we should absolutely include the uploading of both file types across all upload file example templates within our e2e tests, as part of this work. For some extra context, there's a couple of notes that we learned and might be worth knowing:
- We believe this problem only occurs with the fields that are _conditionally required_ rather than just optionally required; optionally required fields have Undefineable Booleans to act as a sanity check that keeps them the same in both Excel and CSV uploads (if they're filled out, the value is the same, and if they're blank, then both undefined and null will get parsed as UndefinneableBoolean.NotCollected)
- The two properties in question (birthTown and birthState) don't even get created if a CSV is uploaded, so as part of the correction in any utils or functions that get modified (likely `generateCSV`), we need to make sure the properties exist and are explicitly set to undefined

PR #1163, opened 2021-02-05
Make csv uploading robust

bamader:
## Background
While auditing the optional and conditionally required data fields, Jordan and I discovered a bug in which blank cells in CSV files are simply ignored during parsing. This causes some properties in the Child entity schema to not exist on records, and these properties are mistakenly parsed as null when they should be undefined (this matters because of the Unknown/Not Collected state). This PR fixes this parsing oversight and diversifies our e2e tests to check uploads of both excel sheets and CSVs.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1157

## Validation Plan
I've tested these small changes with a variety of uploaded sheets, making sure everything we expect to be there is there. The big tests are the e2es, which have been upgraded to test uploading and parsing of both excel files and csv sheets.

## Automated Testing
- [x] If applicable, unit tests have been added to cover this changeset (not applicable).
- [x] If applicable, e2e tests have been added to cover this changeset (not applicable).

## Additional Context
I spent some time reading about the excel package we use, `xlsx`, as well as Excel's parsing strategies for sheets in general. Unfortunately, Excel does two things that make overriding the parsing behavior impossible: 1) it aggressively interprets/interpolates data it reads from _any_ sheet, leading to weird casts to store data in some proprietary representation, and 2) it stores sheets in a sparse array, so CSV blanks that get mistakenly ignored by the parser don't even show up in the row representation. This prevents us from doing anything at the parsing level to fix the problem, so we have to resort to manual corrections when we create child objects. I've audited all our optional and conditional fields to see where this might be a problem, and I believe the only two where it's an issue are `birthTown` and `birthState` (since null there clicks the Not Disclosed checkbox). All of this goes to say that manually correcting the values isn't the prettiest, but it _is_ the easiest. Because microsoft.


jlhogan (approved):

bamader:
Okay all changed tests are confirmed working in staging with stable passes. @JordanGuinn @jlhogan This should be ready for review now--I think it should solve at least some of your branch's e2e pain, Julia. Sorry for the delay in marking it ready!

src/controllers/enrollmentReports/map/mapUtils/child.ts
```diff
@@ -72,8 +72,18 @@ export const mapChild = (
...
     birthCertificateType: birthCertificateType,
-    birthTown: source.birthTown,
-    birthState: source.birthState,
+    // Need these corrections to defeat Excel: xlsx parses CSVs
+    // aggressively, storing things in a sparse representation.
+    // Blanks are interpreted as properties not existing, and
+    // there are NO parsing options to change this because the
+    // problem is with Excel itself. So we need to force conditionally
+    // expected properties to appear with the empty string.
+    birthTown: (source as Object).hasOwnProperty('birthTown')
```
JordanGuinn, line 81: HEH

JordanGuinn (approved):
Oh yeah!  I forgot about this CSV fun 😅

Issue #1154, opened 2021-02-04
Not all funding spaces are correctly marked in change request form 


jlhogan:
## Background
from user:
> I noticed that our funding for School Readiness was not checked on the Home Page regarding Sites/Funding.  Not sure if this has anything to do with the children not showing, but I did just make that change and checked the SR box. 

confirmed in prod, CDC spaces are checked, PSR and CSR are not
All funding spaces exist in DB

## Acceptance Criteria
- PSR AND CSR fundings are appropriately checked if organization has them
- `Request updated name` becomes `Request site name change` in the site change form (Users seem confused and a few put their own names in there)
- ChangeFundingSpaceRequest tracks funding_space id in addition to the funding space string (optional FK for existing funding spaces!)
- If possible, we only write rows that are diff from existing state (either when a funding space should be added or removed)

bamader:
Clarifying question: for the AC bullet `PSR AND CSR fundings are appropriately checked if organization has them` does this mean we want to split out PSR and CSR as separate sets of checkboxes? Or does it mean leave the one set of school readiness checkboxes, and if an organization has either PSR or CSR fundings, we check them?

jlhogan:
I think it makes most sense to have PSR and CSR as separate checkboxes -- we are retaining the specific naming everywhere else in the tool, and additionally have already run into one issue where a user had CSR and actually needed PSR.

bamader:
Awesome. That was my instinct, too.

JordanGuinn:
## Testing Strategy
- [ ] Up migration succeeds
- [ ] Down migration succeeds
- [ ] `funding_space_id` is present in the `change_funding_space_request ` table
- [ ] All funding source/funding time options are present
- [ ] `Request updated name` now reads as `Request site name changes` in the site change request form
- [ ] PSR fundings are checked if the organization has them
- [ ] CSR fundings are checked if the organization has them
- [ ] `funding_space_id` is captured for new site change requests
- [ ] Site change requests only capture the diff from current site state for the organization

JordanGuinn:
@bamader This looks good and ready to merge!

PR #1171, opened 2021-02-09
Reduce revision request confusion

bamader:
## Background
The revision request form was generating some confusion and issues regarding school readiness funding spaces. This PR corrects those and also makes storing requests in the DB better by only storing the diffs.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1154

## Validation Plan
Visually ensured the revision request page appears as it should (with new copy and new funding space checkboxes), and locally logged DB requests to make sure only the diff gets saved.

## Automated Testing


client/src/containers/Home/RevisionRequest.tsx
```diff
@@ -60,12 +60,15 @@ export const RevisionRequest: React.FC = () => {
...
-            const stringRep = ag + ' - ' + fst.displayName + ' - ' + time.value;
-            fsOptions.push({
-              fundingSpace: stringRep,
-              shouldHave: false,
-            } as ChangeFundingSpaceRequest);
+          fst.fundingSources.forEach((source) => {
+            const sourceName = source.split('-')[1].trim();
+            fst.fundingTimes.forEach((time) => {
+              const stringRep = ag + ' - ' + sourceName + ' - ' + time.value;
+              fsOptions.push({
```
JordanGuinn, line 67: FYI, type assertion like this (`... as ChangeFundingSpaceRequest`) can [end up being problematic](https://basarat.gitbook.io/typescript/type-system/type-assertion#assertion-considered-harmful) and should probably be avoided if possible.  Moving forward, if we're just looking to create a portion of a given interface, we should start using [partials](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype) to reflect that.

So in this case, fsOptions would be defined as

`const fsOptions: Array<Partial<ChangeFundingSpaceRequest>> = []`

and that would let us remove `as ChangeFundingSpaceRequest` completely.

JordanGuinn (commented):
Fix itself looks solid, just a few general comments and I think this'll be good to go 👍🏽

client/src/containers/Home/RevisionRequest.tsx
```diff
@@ -82,18 +85,16 @@ export const RevisionRequest: React.FC = () => {
...
-              fs.source.split('-')[1].trim() +
-              ' - ' +
-              fs.time;
+            const spaceName = fs.source.split('-')[1].trim();
+            const stringRep = fs.ageGroup + ' - ' + spaceName + ' - ' + fs.time;
             const match = userFundingSpaces.find(
               (fs) => fs.fundingSpace === stringRep
             );
             if (match) {
+              console.log(match);
```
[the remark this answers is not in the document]
bamader: lol yes my b 😂

src/routes/revisions.ts
```diff
@@ -46,15 +53,32 @@ revisionsRouter.post(
...
+              const spaceName = fs.source.split('-')[1].trim();
+              const stringRep =
+                fs.ageGroup + ' - ' + spaceName + ' - ' + fs.time;
+              return stringRep === r.fundingSpace;
             });
-            await getManager().save(dbReq);
+
+            // Create if it's either a new FS or the user is saying they
+            // shouldn't have this space
+            if (!match || (match && !r.shouldHave)) {
```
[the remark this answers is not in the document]
bamader: Turned out to not be too hard of an upgrade so why not 😎

client/src/containers/Home/RevisionRequest.tsx
```diff
@@ -82,18 +85,16 @@ export const RevisionRequest: React.FC = () => {
       apiGet('funding-spaces', accessToken)
         .then((res: FundingSpace[]) => {
           res.forEach((fs) => {
-            const stringRep =
-              fs.ageGroup +
-              ' - ' +
-              fs.source.split('-')[1].trim() +
-              ' - ' +
-              fs.time;
+            const spaceName = fs.source.split('-')[1].trim();
```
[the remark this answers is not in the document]
bamader: Yeah good call

client/src/containers/Home/RevisionRequest.tsx
bamader: Can do. I _think_ everything is working the same way after this change, but this is why I'm glad we've got an "other dev QA" step 😛 (I just say that because this particular change impacts several state variables and setState functions and react is finicky with those).

jlhogan:
Have you confirmed that ids appear in the DB as entered in the UI? 
I'm seeing registry/license ids that definitely don't look correct, and I think must be a type conversion thing?

bamader:
Hmm you're seeing weird IDs for the Add Site stuff? That's odd to me because they're just strings everywhere so I don't think there's any type conversion happening. Lemme check and see what's up.

JordanGuinn (approved):

jlhogan:
yeah, the add site stuff! could have been user error, but the 2 records both seemed wonky. 

thank you!!!

bamader:
Finished up a testing sesh and from what I can tell the DB is storing the exact same values that are being typed into the form. The UI matches the request object, which matches the DB created object, which matches what we save, which matches what's actually saved (I tried to be thorough 😝 ). I don't think there's any funky casting behavior going on that I can see. If you have the weird IDs, can I ask what they were? Wondering if it might be user error or whether the IDs are actually right (even if they are unnatural).

bamader:
@JordanGuinn Any extra thoughts on ways to validate this? Short of just futzing with it in QA?

jlhogan:
Haven't dug into but just repro-ed in staging:

Form:
Screenshot from 2021-02-11 09-12-23
Request:
Screenshot from 2021-02-11 09-13-18
DB:
```
Julia''s Newest Site: 64786435200000, -30294691200000, 4394839    
```

bamader:
Wtf is this nonsense...clearly needs a deeper dive so imma get on this. It looks like the same weird conversion as the zip code bug we dealt with a couple months ago (short digit strings got turned into negative longs). Don't know how that would happen but this is _weirdness_

bamader:
The DB thinks licenseId and registryId are _Moments_?!

jlhogan:
the json parser

bamader:
json parse is set to false rn in the apiPost call, though. Or are you saying we _should_ be using json parse?

src/index.ts
```diff
@@ -32,7 +32,11 @@ getConnectionOptions().then((connectionOptions) => {
       // Register pre-processing middlewares
       const dateReviver = (_: any, value: string) => {
```
jlhogan, line 33: :tada: last thing :D :D because we use a date reviver in client AND server for json parsing, could we define this in the `shared` part of client and use in  both places. That is here^, and [here](https://github.com/ctoec/data-collection/blob/base/client/src/utils/api.ts#L127) ?

jlhogan (approved):

src/index.ts
bamader: But of course!

PR #1177, opened 2021-02-16
Remove superfluous CircleCI workflow

JordanGuinn:
## Background
Removing this because CircleCI allows you to re-run workflows, both from a particular step as well as from the beginning.  Unless this was added for some other unknown reason, I don't see any need for this.

## GitHub Issue
N/A

## Associated PRs
N/A

## Validation Plan
No validation, just removing an unused CircleCI workflow.

## Automated Testing


jlhogan:
it was added for a particular reason, which is to be able to deploy to both staging (to run the e2e test) and then devsecure (so that environment stays up to date without us having to manually bump it along with prod). 

At the time, re-running the `deploy` step for a different env was not working in a single workflow wasn't working, although now I cannot find what the exact problem was.

But also I'm now looking at this and seeing that we're deploying a staging build to devsecure? so that's wrong anyway? So perhaps let's find a better way to automate deploys to devsecure if that's indeed a thing we want (which i think it is) and say goodbye to this ugliness.

jlhogan (approved):

bamader (approved):

vonbearshark (approved):
> done

> oh whoops

you love to see it

Issue #101, opened 2020-08-04
Spreadsheet upload: As a person updating data, I can upload a new spreadsheet with my latest enrollments.


RyanHansz:
## Background
As a user, I can upload a file containing new enrollment records, as well as changes to my existing roster, preview those changes, and opt to cancel the upload or save them to my roster.

### [Designs](https://share.goabstract.com/075833a6-f13d-4626-840e-38479241cadd)

**_NOTE:_** This should receive a test/design review prior to completion/merge/deploy.

## Acceptance Criteria
- I can upload a spreadsheet that will update existing records, add new records, and withdraw old records. Update types include:
  - aged up
  - changed site
  - changed funding
  - new
  - withdrawn
  - income determination
  - edited (i.e. added birth cert)
- ECE Reporter will detect if my upload contains data for existing records, identified as those with perfect child ID matches
  - If birth cert didn't exist before, but does in the new upload, match if DOB + name match, and update with birth cert data
- For now, only edits supported will be:
  - adding birth cert info if it doesnt exist (changing from "Unavailable" to some other birth cert type is supported)
  - adding/editing family address
  - any edits not captured in one of the above "update types" will not be handled yet, and will result in new entities being created (that is, changing site or entry/exit of existing enrollment results in new enrollment, changing funding first/last reporting period or funding space results in new funding, changing existing income determination information results in new income determination)
- API response needs to contain some data structure representing the uploaded data, augmented with the upload type of each record
- User is still able to upload a file with historical enrollments (**IS THIS TRUE?**) so the data structure may not map exactly to the incoming sheet (i.e. there may be multiple rows for a single child)
  - If historical records included in upload, they should end up in the response data structure as a single row with multiple update type tags (i.e. new and changed site)
- I no longer need to choose which site to replace data for, since my file upload will have that information already
  - as a result, deleting roster data is no longer part of file upload in backend

RyanHansz:
If we switch new components to Carbon, here's the progress indicator included in these designs:
https://www.carbondesignsystem.com/components/progress-indicator/usage/

bamader:
`API response needs to contain some data structure representing the uploaded data, augmented with the upload type of each record`
Is this data structure what's going to be used for the other ticket with the "upload wizard"? I don't quite understand why having a structure augmented with the type of action performed (new record, edited record, whatever) is necessary for the work in this ticket specifically. All the other bullets in the AC make perfect sense, but that one I'm not sure I get.

[...]

PR #1182, opened 2021-02-17
Batch update

bamader:
## Background
This PR provides the first pass of the Batch Update MVP. It allows users to upload spreadsheets containing either new or existing child records, and then creates or updates any relevant entities accordingly. For the list of supported updates and new entities, check out the ticket (right now, updates are limited in scope, but enrollment/funding updating is pretty sweet). Tests are still incoming on this, but wanted to get the mapping logic up for y'all to see.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/101

## Validation Plan
Validation right now has consisted of visual examination and console logging of a curated set of records. I've attached two spreadsheets here that I've used to test all changed features of the PR thus far. The spreadsheets include test cases for:
- 3 cases where child records _shouldn't_ match existing things, because one of the ID match fields is wrong
- 3 cases of updating birth certificate information (one for switching from undefined to defined, and one for providing updates to birthtown and birthstate)
- A case of creating a family address for a child whose family had no recorded address at all
- A case of updating a family address (including checking the homelessness field)
- A case of creating a new income determination (also showing this works if a family has a non-disclosed income)
- Creating a full new enrollment and funding (i.e. changing a child's enrollment)
- Adding a new funding to an existing enrollment.
I've also console logged the data structure the API sends back to the front end to make sure each record is associated with the correct change tag (which is a new enum denoting what kind of upload/update was performed on each record). Here are the files I used to test all the different cases of updating (if you wanna use these yourself, obviously `init` should be uploaded first, and then `changes`): 
[Batch_Update_Init.xlsx](https://github.com/ctoec/data-collection/files/6012823/Batch_Update_Init.xlsx)
[Batch_Update_Changes.xlsx](https://github.com/ctoec/data-collection/files/6012824/Batch_Update_Changes.xlsx)

## Automated Testing
- [ ] If applicable, unit tests have been added to cover this changeset (backend tests are soon to follow!).

## Additional Context
I think the comments should make everything pretty clear. I tried to follow the existing logic in `map` wherever possible, but there were still some design decisions that made this code pretty beefy so lemme know if anything is unclear.


JordanGuinn:
@jlhogan @bamader Not necessarily specific to this PR, but applicable to the update file MVP as a whole:

Since we don't want to ship this until the other two tickets are completed as well, I'm thinking we should probably make a `FEATURE-update-from-file` branch or something and point this PR at it.  Effectively, we probably want a designated space that _isn't_ `base` that we can use to hold/validate this work until it's fully ready for production.  That way it doesn't necessarily need to hold up any other hot ticket bugfixes we may want to release.

bamader:
@JordanGuinn I think that's an _excellent_ idea. This way, too, I can keep developing the other parts of the new upload wizard 🧙  while we review the logic here (gonna get crackin' on that today).

client/src/containers/Upload/Upload.tsx
```diff
@@ -124,33 +115,14 @@ const Upload: React.FC = () => {
     setErrorModalOpen(false);
   };
 
-  // Case where the user does have an existing roster; need to
```
jlhogan, line 127: byeeeeeeeeeeeeeeeeeeeeee :wave:

client/src/containers/Upload/UploadPreviewTableObject.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Basic type to hold the properties we want to display in
+ * the formatted table that shows a preview of a user's upload
+ */
+export type UploadPreviewTableObject = {
```
jlhogan, line 5: UploadPreviewTableRow?

src/controllers/enrollmentReports/formatUploadPreview.ts
```diff
@@ -0,0 +1,31 @@
...
+export const formatUploadPreview = (mapResult: MapResult) => {
+  const formattedPreview: Partial<UploadPreviewTableObject>[] = mapResult.children.map(
+    (c, idx) => {
+      const recentEnrollment = getLastEnrollment(c);
+      const recentFunding = getLastFunding(recentEnrollment);
+      const tableObj: Partial<UploadPreviewTableObject> = {
+        name: nameFormatter(c, { lastNameFirst: true, capitalize: true }),
+        tags: mapResult.changeTagsForChildren[idx],
+        birthDate: c.birthdate.format('MM/DD/YYYY'),
+        fundingSource: recentFunding
```
jlhogan, line 18: why the '-' for empties?

src/controllers/enrollmentReports/map/map.ts
```diff
@@ -81,10 +145,86 @@ export async function mapRows(
...
-    return children;
+    children.forEach((c) => {
+      mapResult.children.push(c);
+    });
+    return mapResult;
   }
 
-  // Otherwise, create entities in DB
+  // Batch update all the entities we modified fields for
+  const updatedFamilies = await doBatchedInsert<Family>(
```
jlhogan, line 155: should doBatchedInsert be responsible for adding `authorId` to entities to avoid duplicating that logic all over?

src/controllers/enrollmentReports/map/map.ts
```diff
@@ -81,10 +145,86 @@ export async function mapRows(
...
     });
-    return children;
+    children.forEach((c) => {
+      mapResult.children.push(c);
+    });
+    return mapResult;
   }
 
-  // Otherwise, create entities in DB
+  // Batch update all the entities we modified fields for
```
jlhogan, line 154: can we pull this (update existing entities) and the next chunk (create new entities) into functions to make this main function less long and more readable?

src/controllers/enrollmentReports/map/map.ts
```diff
@@ -166,9 +306,138 @@ export async function mapRows(
...
+    const orgNamesForError = `${orgNames
+      .slice(0, -1)
+      .join(', ')}, or ${orgNames.slice(-1)}`;
+    throw new BadRequestError(
+      `You entered invalid provider names\nCheck that your spreadsheet provider column only contains ${orgNamesForError} before uploading again.`
+    );
+  }
+  const site = lookUpSite(source, organization.id, userSites);
+
+  let matchingIdx = mapResult.children.findIndex((c) => c.id === child.id);
```
jlhogan, line 342: what's happening here?

src/controllers/enrollmentReports/map/map.ts
```diff
@@ -166,9 +306,138 @@ export async function mapRows(
...
   });
-  return createdChildren;
+  createdChildren.forEach((c) => {
+    mapResult.changeTagsForChildren.push([ChangeTag.NewRecord]);
+    mapResult.children.push(c);
+  });
+  return mapResult;
 }
 
+const updateRecord = async (
```
jlhogan, line 316: blahhhhhhhhhhh this is so many params, which feels like  a bad sign, but i'm not sure what is the useful improvement here. 
One thought is do we need to provide the *toUpdate arrays? Can we just build them in this func, and then return them to the calling context?

src/controllers/enrollmentReports/map/map.ts
```diff
@@ -166,9 +306,138 @@ export async function mapRows(
...
+    source,
+    familiesToUpdate
+  );
+  if (changedBirthCert || changedAddress) {
+    mapResult.changeTagsForChildren[matchingIdx].push(ChangeTag.Edited);
+  }
+
+  // Everything else involves making new entities
+  const currentDetermination = getLastIncomeDetermination(child.family);
+  const determination = mapIncomeDetermination(source, child.family);
```
jlhogan, line 366: can we pull these things out into sep funcs so we can read through this more easily to understand high level gist of whats up 
like 
`createDetermination(source, child, mapResult)`
or 
`updateOrCreateEnrollment(source,child, mapResult)`
 and extract the logic of these actions to some other file b/c this is loooooong :)

src/routes/enrollmentReports.ts
```diff
@@ -132,16 +119,24 @@ enrollmentReportsRouter.post(
...
 
-        let numActive = 0,
+        // Use tags to decide which kinds of uploads/updates were performed
+        // on each record
+        let numNew = 0,
+          numUpdated = 0,
           numWithdrawn = 0;
-        reportChildren.forEach((child) => {
+        mapResult.children.forEach((child, idx) => {
           if (child.enrollments?.every((e) => e.exit)) numWithdrawn += 1;
```
jlhogan, line 128: should we also use change tag to determine withdrawn, for consistency ?

jlhogan (commented):
overall looks good to me!!! a bit more function decomposition could make it easier to read and grok without heavy commenting 

holding off on approving because I want to do some manual validation later today, but wanted to share these initial thoughts

client/src/containers/Upload/UploadPreviewTableObject.ts
bamader: The probably overly wordy name for the data structure that holds all the info we'll use to format the table we show in the Preview tab of the upload wizard. I did a similar thing for a row in the error modal table (which is conveniently also being incorporated into the wizard).

src/controllers/enrollmentReports/formatUploadPreview.ts
bamader: Just so when we were displaying the table in the preview tab, we didn't have to check if properties existed if I made them optional or allowed undefined. We could just iterate through the properties and display whatever string is there.

client/src/containers/Upload/Upload.tsx
bamader: 😎 💥

src/controllers/enrollmentReports/map/map.ts
bamader: Oh definitely. That'll massively clean up each of the calls.

src/controllers/enrollmentReports/map/map.ts
bamader: _UGH_ I know this is the ugliest function signature. Passing them in is literally just there so we can consolidate all the entity changes of each kind into one place for BatchInserting. One thing I could do though would be to define them at the top of the file like `let *toUpdate: *[]` and then at the start of the `mapRows` function, initialize all the declared arrays to `[]`. That way the functions can reference them without having them explicitly passed into every call?

src/routes/enrollmentReports.ts
bamader: We can do that. That raises a question I have about what the "Withdrawn" tag should mean. Does withdrawn refer to a _child_ for which every enrollment has an exit date, or does it refer to a single _enrollment_ that has an exit date? Right now it seems to be the former, but I just realized I don't actually know if that's what it means to be a withdrawn record.

bamader:
Okay automated tests have been added and functions have been decomposed and separated to shorten files and increase readability. All new tests are passing, and the same validation suite I used before the decomposition works after, so I think this is ready for someone else to play around!

src/routes/enrollmentReports.ts
jlhogan: well given that we want these tags to tell the user what they did, we would only want the withdrawn tag to apply for students that were withdrawn and not enrolled in new space. Those records which were withdrawn but then had a new enrollment created would fall into aged up or moved within program buckets. 

so tldr; withdrawn means ALL enrollments are exited

client/src/containers/Upload/UploadPreviewTableObject.ts
jlhogan: oh yeah sorry i was just advocating for "row" over "TableObject" b/c that makes a lil more sense to my mind. not strongly held, just a preference

src/controllers/enrollmentReports/formatUploadPreview.ts
jlhogan: my vote would be for this to just have undefined for undefined data, and then let the display logic live in the UI; that way, if we want to display something other than `-` at some point we don't have to make a change back here

src/controllers/enrollmentReports/map/map.ts
jlhogan: eh; i dont know that globals feel better than somanyparameters. 
if we need to be updating in place the existing arrays, then this is fine

src/controllers/enrollmentReports/map/map.ts
```diff
@@ -53,17 +71,65 @@ export async function mapRows(
...
+        ],
+      });
+      const match = potentialMatches.find(
+        (c) =>
+          c.birthdate.format('MM/DD/YYYY') ===
+            row.birthdate.format('MM/DD/YYYY') &&
+          ((c.sasid && c.sasid === row.sasidUniqueId) ||
+            (c.uniqueId && c.uniqueId === row.sasidUniqueId) ||
+            (!c.sasid && !row.sasidUniqueId))
       );
```
jlhogan, line 111: can encapsulate all of this in a function called `doesChildMatchExist` or something which will return the match or undefined  and then no need for comments :)

src/controllers/enrollmentReports/map/map.ts
```diff
@@ -81,93 +147,108 @@ export async function mapRows(
         (det) => (det.family = undefined)
       );
     });
-    return children;
+    children.forEach((c) => {
```
jlhogan, line 150: why is this necessary? why can't the children array just be concatted to `mapResult.children` (if we expect it to already have values) or assigned to `mapResult.children` (if it won't already have values)

src/controllers/enrollmentReports/map/map.ts
```diff
@@ -81,93 +147,108 @@ export async function mapRows(
...
+      mapResult.children.push(c);
+    });
+    return mapResult;
   }
 
-  // Otherwise, create entities in DB
-  // Create families
-  const createdFamilies = await doBatchedInsert<Family>(
+  // Otherwise, update the entities that already exist, and create
```
jlhogan, line 156: :raised_hands: this reads so well!

src/controllers/enrollmentReports/map/map.ts
jlhogan: add a summary of this context in a comment ?

src/controllers/enrollmentReports/map/uploadTypes.ts
```diff
@@ -0,0 +1,13 @@
+import { Child } from '../../../entity';
+
+/**
+ * Type that holds the accumulation of all new and edited
+ * upload changes to child records from a spreadsheet. The
+ * index of child i in the `children` field is the same
+ * index where the tags for child i can be found in the
+ * `changeTags` field.
+ */
+export type MapResult = {
```
jlhogan, line 10: why not put the change tags on the children? was there a reason it was easier/desirable to have them in two separate arrays with matching indexing?

jlhogan (commented):
a few more syntaxy things
manually validating now; will then comment with my validation plan + approve !

jlhogan:
## Validation  plan
- [ ] can withdraw an active enrollment
- [ ] can change funding for an active enrollment
- [ ] can change enrollment by adding row for new enrollment
- [ ] can update family address
- [ ] can update child demographic info
- [ ] can update existing income determination 
- [ ] can create new income determination
- [ ] when changing existing record, only firstname, lastname, date of birth  are required (plus, obviously, whatever info the user is changing)

bamader:
Comments look good! I'll work through them here shortly (just gotta get a superset thing set up for Alex and Kyle). As you're testing, lemme know if there are other entities/things we want to support updates for. Right now it's just birth cert and family address, but we can easily broaden that.

src/controllers/enrollmentReports/map/map.ts
bamader: Ha. Haha. It totally can. TFW you're a thousand lines deep and you forget how lists work... 🤦

src/controllers/enrollmentReports/map/uploadTypes.ts
bamader: Honestly it was just the way my brain thought of doing this. To my perhaps convoluted mind it was just the most straightforward and intuitive way to keep track of change tags--have a list where there's one entry for each child. That way, no entities themselves get modified. Not at all saying this was the best to do it, just what I thought of first that seemed to do the job effectively!

bamader:
I literally just pushed a fix for a withdrawn enrollment thing I missed, so lemme try out these files now and see if it works

bamader:
Okay, so good news (and bad news but not really). The good news is that the code works fine, and the test cases for everyone in the files is successful.

Justus - gets withdrawn and has the enrollment and funding given an end date
Zita - already covered that her birth cert won't change cuz it's not supported
Shanel - current funding ends, another begins
Hermina- new det
Della - changed address

The bad news is that I drove myself slightly crazy trying to figure out what was wrong when in fact I forgot to check that everything that needed to be in the update sheet was actually there 😂 🙃 . We were having trouble because Model wasn't provided in the enrollments to change/withdraw and the matching function currently needs it, and because for a couple of the kiddos birthdate wasn't there so the child match function was failing (it needs first name, last name, and DOB). Here's an updated version of the "update" sheet for your testing purposes:
[update.xlsx](https://github.com/ctoec/data-collection/files/6013790/update.xlsx)

So just to document things:
- Child match needs first name, last name, and DOB at a minimum; if a child was originally defined with a SASID or a unique ID, it also needs that
- Enrollment and Fundings need all the same fields that were supplied at creation to find them when update matching. The big thing to note here is Care Model. I figured this was okay because users should still have it cuz that's how it was created, so they could just copy the four enrollment fields in a contiguous block.
- Update rows use any provided last dates if they're there, and if not, they assume that the end date for enrollments/fundings is one before the new one starts.

Happy testing when you get back to it!

src/controllers/enrollmentReports/map/uploadTypes.ts
jlhogan: interested in @JordanGuinn's thoughts for a "tie-breaker"
I'm not opposed to this solution, altho just sticking the tags on the children doesn't seem wrong to me

jlhogan:
@bamader  testing again, but just off the bat using your new upload sheet,I still get income det errors for all rows without income dets

bamader:
Wtf? I'm not seeing any errors locally. Do you have the latest commit pulled?

jlhogan:
HRMPH i did a command to pull your stuff, but had some changes on local i didnt realize it rejected it. ignore me! will test again soon! thanks!

bamader:
PHEW okay, was about to put my fist through my machine 😅

src/controllers/enrollmentReports/map/uploadTypes.ts
JordanGuinn: It'd make sense to me to either:

- Throw the change tags directly on the children
- Group the children by the change tag applicable to them

But either way, can we call this something other than `MapResult`?

src/controllers/enrollmentReports/map/mapUtils/funding.ts
```diff
@@ -84,3 +86,144 @@ export const mapFunding = (
...
 };
+
+/**
+ * Util function that determines whether the funding object mapped
+ * from a current EnrollmentReportRow contains new or different
+ * information from a child's current funding.
+ * @param fundingFromRow
+ * @param funding
+ */
+export const rowHasNewFunding = (fundingFromRow: Funding, funding: Funding) => {
```
JordanGuinn, line 97: Do we need both `rowHasNewFunding` and `rowEndsCurrentFunding` functions?  It looks like one is the direct inverse of the other - like one checks to see if two fundings are the same, and the other checks to see if they're different.

src/controllers/enrollmentReports/map/mapUtils/batchSave.ts
```diff
@@ -0,0 +1,223 @@
+import { UpdateMetaData } from '../../../../entity/embeddedColumns/UpdateMetaData';
```
JordanGuinn, line 1: Nit: None of these files really belong in this nested `mapUtils` directory within the enrollment reports controller

src/controllers/enrollmentReports/map/uploadTypes.ts
jlhogan: ^ agree :)

client/src/shared/payloads/BatchUpload.ts
```diff
@@ -1,5 +1,9 @@
+import { UploadPreviewTableObject } from '../../containers/Upload/UploadPreviewTableObject';
+
 export interface BatchUpload {
```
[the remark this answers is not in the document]
jlhogan: we should make this change for other poorly named payloads (lookin at me :eyes: )

jlhogan (approved):

src/controllers/enrollmentReports/map/mapUtils/batchSave.ts
jlhogan: @JordanGuinn can you explain this a bit more?

Issue #239, opened 2020-08-28
Batch Upload Preview: I can review all of my updates from an uploaded spreadsheet before adding them to my roster.


RyanHansz:
## Background
When the API is updated to process data updates from file, the user wants to see a preview of their updates before committing them to the system

## Acceptance Criteria
- Primary button in previous progress bar step ("review missing info") has text "next", in has missing info and no missing info state
- third step is added to file upload progress bar, "Preview updates"
- content is a roster-like view of added or updated records, separated by age group. Roster has additional "Update" column. Updates types will come from the backend in some data structure (like errors do)
- for now, do not include unchanged records in this view (i.e. ignore "show all unchanged records" in designs)
- Secondary button has text "cancel", primary button has text "Upload"
- upload action will commit all the new data to the user's roster, and navigate them there with a success banner summarizing the changes that were made

## Designs
[Design](https://share.goabstract.com/075833a6-f13d-4626-840e-38479241cadd?collectionLayerId=ff158304-b286-46b0-9dbb-2696444df5de&mode=design)

PR #1192, opened 2021-02-22
First pass at preview screen

bamader:
## Background
As part of the change from the Upload page to an Upload Wizard, we need a screen which shows users a preview of their upload, including the type of upload/update performed as well as whether any records are missing info. This PR creates the Preview screen, which includes the roster preview table, altered header copy, and cards for number of uploaded records of each type.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/239

## Validation Plan
Using the same spreadsheets that we use to test the batch update utility, we can visually test whether the records we expect to change (as well as the new records in our sheets) are correctly shown. I also made sure the number of uploaded records of each type (new, withdrawn, updated) match the correct number of each kind of record in the sheet itself.

## Automated Testing
- [ ] If applicable, unit tests have been added to cover this changeset (no new tests).
- [ ] If applicable, e2e tests have been added to cover this changeset (no new tests).

## Additional Context
This PR assumes content from my batch-update branch, which is why I've based it off of that. This is a prototype only, because Jordan is working on the Upload Wizard utility itself, so once he's done with that ticket and has it merged into the MVP feature branch, I can pull those changes in and incorporate the Preview screen into the wizard as opposed to keeping it on the Upload page.


src/controllers/enrollmentReports/formatUploadPreview.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export const formatUploadPreview = (mapResult: MapResult) => {
+  const formattedPreview: Partial<UploadPreviewTableObject>[] = mapResult.children.map(
+    (c, idx) => {
+      const recentEnrollment = getLastEnrollment(c);
+      const recentFunding = getLastFunding(recentEnrollment);
+      const tableObj: Partial<UploadPreviewTableObject> = {
+        name: nameFormatter(c, { lastNameFirst: true, capitalize: true }),
+        ageGroup: recentEnrollment ? recentEnrollment.ageGroup : undefined,
+        // Greater than 1 here because we didn't save nested enrollments and
```
jlhogan, line 23: what?

client/src/containers/Upload/Upload.tsx
```diff
@@ -223,28 +230,86 @@ const Upload: React.FC = () => {
...
+                  id="report"
+                  label="Choose a file"
+                  onChange={fileUpload}
+                />
+              </LoadingWrapper>
+            </form>
+          </div>
+        </ErrorBoundary>
+      ) : (
+        <>
```
jlhogan, line 261: instead of embedding this in Upload inline, let's define this content in it's own component so we can more easily plop it into wizard

src/controllers/enrollmentReports/formatUploadPreview.ts
bamader: Yeah this was mildly annoying. One of our validations (everyChildHasFundedEnrollment) makes a query to the DB to retrieve enrollments and fundings for a given childId, and it uses that result to determine if the child has a funded enrollment. But, for the preview view, we don't actually save any of the entities in the DB, we just make them as objects and have them point to each other cuz we just need it to display. So the validation thinks every child is wrong and every child's validationError's array winds up having an entry for it. This is to ignore that entry. I could also rewrite this as a filter for all validationErrors that aren't the hasFundedEnrollment, if that's clearer?

client/src/containers/Upload/Upload.tsx
bamader: Yeah can do! Once we actually start getting these things merged, I think this PR will be the last one to land, so I can make any necessary cleanups we want.

JordanGuinn (approved):

Issue #1147, opened 2021-02-03
Transient error with backend request logger


jlhogan:
```
exports.requestLogger = morgan_1.default(':remote-addr - :userId [:date[clf]] ":method :url HTTP/:http-version" :status :res[content-length]', { skip: (req) => !req.baseUrl.includes('api') });
^
TypeError: Cannot read property 'includes' of undefined
```

PR #1200, opened 2021-02-25
1147: Resolves transient error that kills node process

jlhogan:
## Background
causes type error which kills node proc. A problem in pre-prod, because this clears  + reseeds database, which should only happen on an intentional deploy or restart. 

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1147


JordanGuinn (approved):

bamader (approved):
lol

PR #1221, opened 2021-03-05
Updated docs for M1, added port mapping for db and winged-keys db

vonbearshark:


docker-compose.yaml
```diff
@@ -33,6 +33,7 @@ services:
       WINGED_KEYS_HOST: 'https://winged-keys:5050'
       # Change this to true if you want to create a test user with permissions for multiple organizations
       MULTI_ORG_USER: 'false'
+      PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: 'true'
```
vonbearshark, line 36: Not sure if this is an issue. If so, happy to just add it to the `README` instead, with the db `image` update

jlhogan: im fine with this :)

docker-compose.yaml
```diff
@@ -41,12 +42,16 @@ services:
     environment:
       SA_PASSWORD: TestPassword1
       ACCEPT_EULA: Y
+    ports:
```
jlhogan, line 45: why do you want the DB ports exposed? no issue, just curious. are u using a GUI for the DB stuff? which one?

jlhogan (approved):

docker-compose.yaml
vonbearshark: Yep, exactly. I use [TablePlus](https://tableplus.com/). It is fine, pretty simple/intuitive, which is my only standard

bamader (approved):

Issue #1194, opened 2021-02-23
Better funding reporting period mapping in Upload


jlhogan:
## Background
Currently, reporting periods in `mm/dd/yyyy` format are mapped to whatever is the corresponding `mm/yyyy` version. 

However, reporting periods have start/end dates that do not perfectly align with months and so this can result in a row having a first funding period that ends before the enrollment start date if for example the enrollment starts at the end of a month (eg. 03/31/2020), and the reporting period for the corresponding month (03/2020) ends before that date (03/27/2020). This is completely opaque to users, and hard to fix.

## Acceptance Criteria
- When parsing reporting periods, if they are in `mm/dd/yyyy` format, instead of just grabbing `mm/yyyy` from the date, look for the reporting period in the system that includes the specific date

bamader:
Gross thought here: because of serial number conversion magic, anyone that uploads a sheet using excel has all their dates automatically converted to numbers before it gets parsed by moment. We knew that. What I just realized during testing is that this means we don't ever actually get to see the string the user entered. So we don't know if their format was MM/YYYY or MM/DD/YYYY. For example, if we have the date value 03/01/2020 after moment parses the number, we don't know whether it came from the user input 03/01/2020 (in which case the 01 is meaningful and should stay) or from the user input 03/2020 (in which case the 01 was automatically added by moment because of the serialization). How should we handle this? It feels wrong to just assume that any funding period date that has an '01' in its day value just has a placeholder day. But also I don't think it's possible to recover the input string from the serialized number :/

bamader:
Actually wait I was way overthinking this. We can just look for the reporting period that includes the specific date no matter what the user's input string was. If  they enter MM/DD/YYYY, the more specific date will be preserved when parsed as a moment so there's no issue. If they enter MM/YYYY, the day wasn't important anyway so it's not harmful to use the '01' that  gets auto parsed by Moment and Excel. This way we only alter map stuff and not the parse layer. Sorry if that was obvious to everyone else, I'm done having a brain fart now!

jlhogan:
yup ^ you got it :)

PR #1232, opened 2021-03-09
Make RP months align with user entered dates

bamader:
## Background
Reporting periods suck because they don't align with normal month by month dates, and this makes a big headache for users. The mapping between reporting period dates and the dates they write in their sheets is completely opaque and does not need to cause them stress. This PR address a weakness in reporting period mapping by: 
1. not automatically mapping all parsed reporting period fields to have 'startOfMonth' as their day value
2. first checking if a user supplied date is "meaningful" and finding a reporting period whose start and end dates sandwich the user date (makes more sense in the code comments)
3. if a user didn't supply a day value, making sure that the found reporting period is the right one to cover their given reporting period date

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1194

## Validation Plan
Use the attached spreadsheet to:
- Confirm that the system can parse reporting periods entered in either MMM/YYYY or MM/DD/YYYY format
- Confirm that the reporting period chosen for dates entered as MM/DD/YYYY actually includes the date provided
- Confirm that the reporting period chosen for MMM/YYYY dates matches how the period should reasonably be parsed (see context below).
Spreadsheet: 
[reporting_period_test.xlsx](https://github.com/ctoec/data-collection/files/6111687/reporting_period_test.xlsx)

## Automated Testing
- [x] If applicable, unit tests have been added to cover this changeset (N/A).
- [x] If applicable, e2e tests have been added to cover this changeset (N/A).

## Additional Context
There ended up being a little bit of trickiness with knowing which RP to pull. On the one hand, we want to find RPs that actually contain the dates users give us, especially if they themselves entered the day value. But on the other hand, there are some reporting periods for which the first of a month falls in a _different_ reporting period, which generates an error that "funding can't occur before enrollment" through no fault of the user. The spreadsheet I used to test this changeset has a row with an example:

Say there's a child with an enrollment that began on March 4, 2020. Because the user was following our instructions, they entered the first funding period as 03/2020 or Mar-20. Moment will parse the period they gave as 03/01/2020, which is a date that's included in the _February_ reporting period by periodStart <= givenDate <= periodEnd. If we used this, the user would get an error and be told their funding was wrong. What the user actually intended to happen was to use the reporting period that covered dates in March. So this PR checks for whether the parsed reporting period date has the moment auto-parsed 01, and if so, it helps out the user by finding the reporting period whose _period_ (rather than start and end) matches the provided info. If that's over-commented int he code, lemme know and I'll cut it down, but I just thought this deserved some explanation (because although it overrides part of what the ticket said, I believe this is also the real intention of the ticket, to match our software's behavior with how users reasonably expect it to operate).

### DIFF
--- a/src/controllers/enrollmentReports/map/mapUtils/funding.ts
+++ b/src/controllers/enrollmentReports/map/mapUtils/funding.ts
@@ -14,6 +14,7 @@ import {
 import { EnrollmentReportRow } from '../../../../template';
 import { mapEnum, mapFundingTime } from '.';
 import { EnrollmentReportUpdate } from '../uploadTypes';
+import { Moment } from 'moment';
 
 /**
  * Create a Funding object from FlattenedEnrollment source,
@@ -55,19 +56,17 @@ export const mapFunding = (
   let firstReportingPeriod: ReportingPeriod,
     lastReportingPeriod: ReportingPeriod;
   if (source.firstReportingPeriod) {
-    firstReportingPeriod = reportingPeriods.find(
-      (rp) =>
-        rp.type === fundingSource &&
-        rp.period.format('MM-YYYY') ===
-          source.firstReportingPeriod.format('MM-YYYY')
+    firstReportingPeriod = getReportingPeriodFromRow(
+      source.firstReportingPeriod,
+      reportingPeriods,
+      fundingSource
     );
   }
   if (source.lastReportingPeriod) {
-    lastReportingPeriod = reportingPeriods.find(
-      (rp) =>
-        rp.type === fundingSource &&
-        rp.period.format('MM-YYYY') ===
-          source.lastReportingPeriod.format('MM-YYYY')
+    lastReportingPeriod = getReportingPeriodFromRow(
+      source.lastReportingPeriod,
+      reportingPeriods,
+      fundingSource
     );
   }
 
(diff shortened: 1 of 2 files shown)


src/controllers/enrollmentReports/map/mapUtils/funding.ts
```diff
@@ -214,21 +213,66 @@ export const handleFundingUpdate = (
...
+    (rp) =>
+      rp.type === fundingSource && reportingPeriodIncludesDate(rp, sourceRP)
+  );
+
+  // Case for correcting moment auto-parsing: if the user entered MMM/YYYY,
+  // moment auto-parsed it with '01' for the day--but, the reporting
+  // periods for some months don't include the first of the month,
+  // (i.e. "March 2020" goes from 3/2 to 3/29) so modify if needed
+  if (
+    reportingPeriod?.period.format('MM/YYYY') !== sourceRP.format('MM/YYYY') &&
```
JordanGuinn, line 267: If I'm grokking this correctly (which I may very well not be), then this in it's current state will potentially allow for a child with an enrollment that **does** in fact start on the 1st to be associated with a reporting period that starts some time later that month?

If we have to end up doing something like this to distinguish between dates with specific days and just months, we should probably try and use a specific timestamp as well (just to minimize our exposure here, to something smaller than `the first of the month`).

bamader: Yeah you're grokking correctly. The intention here was to correct for the fact that sometimes the first of the month fell in a reporting period named for another month (i.e. March 1 2020 falls in the February 2020 RP). Testing this out, there's actually no error that gets raised in the system because it's okay for a funding period to begin later than an enrollment start, it just can't occur before. The only place this causes an inaccuracy is if an enrollment and reporting period both explicitly begin on the first of the month. Here's an example:

Tyler Michaels has an enrollment that starts on 03/01/2020 with a first reporting period of 03/01/2020. The enrollment date gets parsed correctly as the date given. The reporting period initially is found to be Feb-20, because that's the reporting period that includes 03/01, but the "Feb." name of the period doesn't match the March part of the given date, so the function searches again to "month match." The reporting period it _actually_ picks is Mar-20, which is what I think the user _intends_, but which is technically inaccurate because the Mar-20 reporting period doesn't start until 3/2. I don't know if there's a right answer to how we handle this because I think it's just an edge case of using reporting periods that don't line up.

Re: time stamps: one thought I had was that when moment parses a user's given reporting period date, we could add an hour/minute/second value if the date came in with the format MM/DD/YYYY as a hacky way of saying "hey the user actually means this date, not this RP month." But this doesn't work because we can't reverse-serialize date inputs from excel numeric strings (in addition to being trashy af). 

TLDR: Reporting periods r icky and I don't know if we can (or potentially _should_) engineer some kind of cross-stitch support to correct their mismatches 😕

bamader: DUDE THEY'RE SO BAD 😭 😭. And yeah it's totally okay for it to begin later--in this "edge case," it's one day later internally, and visually no different to the user, so that's about as forgiving as I think we can get with reporting periods.

src/controllers/enrollmentReports/map/mapUtils/funding.ts
```diff
@@ -214,21 +213,66 @@ export const handleFundingUpdate = (
...
+ * the right reporting period.
+ */
+const getReportingPeriodFromRow = (
+  sourceRP: Moment,
+  reportingPeriods: ReportingPeriod[],
+  fundingSource: FundingSource
+) => {
+  // Base case: assume the day part of the date is valuable info
+  // and find a reporting period that includes the full date
+  let reportingPeriod = reportingPeriods.find(
```
vonbearshark, line 257: nit: I personally prefer immutability where possible. So `const` here and then `return`ing in the `if` later instead of reassigning

vonbearshark (approved):

src/controllers/enrollmentReports/map/mapUtils/funding.ts
```diff
@@ -214,21 +213,66 @@ export const handleFundingUpdate = (
...
     currentFunding.lastReportingPeriod = exitPeriod;
     fundingsToUpdate.push(currentFunding);
   }
 };
+
+/**
+ * Simple util to determine whether the periodStart and periodEnd properties
+ * of a given reporting period are date-inclusive of a given moment.
+ */
+const reportingPeriodIncludesDate = (rp: ReportingPeriod, date: Moment) => {
```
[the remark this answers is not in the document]
jlhogan: i dont have a strong preference, but i do think the whole `utils` thing is clunky and unweildy and keeping these things with the classes they pertain to could cut down on that

src/controllers/enrollmentReports/map/mapUtils/funding.ts
jlhogan: yes so now you all understand why I want to get rid of them :D

jlhogan (approved):

jlhogan:
Validation plan:
- [ ]  a row with enrollment that ends near the end of a month, and lists funding start date as the same date (which is after the end of that month's reporting period) and see the correct reporting period (that includes that date) is selected
- [ ] a row with enrollment that begins in a month that actually starts with a different reporting period, and lists the funding start as the enrollment start month (which, as the 01 day is actually a date  in the previous reporting period), and see that the correct reporting period (the one entered by the user, and the enrollment start month) is selected

Issue #1230, opened 2021-03-08
Changing funding source causes browser to become unresponsive


JordanGuinn:
## Background
For whatever reason, changing your funding source selection when updating child data is causing the browser to freeze.  Context from the user experiencing the issue:

```
I tried to add an enrollment for a withdrawn child.  I went to change enrollment, site, date, care model, and age group were all fine.   But when I got to funding source it wouldn't let me pick anything, it just kept saying page unresponsive.  I tried opening the site in different browsers but that didn't help either it still said page unresponsive at the same point (funding source) and wouldn't let me go any further.
```

## Repro Steps
1. Switch to site level perms
   - Delete row from `organization_permission`
   - Add row for your user to `site_permission` for whatever site
2. Create new enrollment from add new record button in roster
3. Go through all sections until enrollment/funding
4. Select infant toddler
5. Select a function option
6. Switch age group to preschool
7. Observe

## Expected Behavior
Funding source is updated without issue

## Actual Behavior
Browser becomes unresponsive.

PR #1233, opened 2021-03-10
Stop rebuilding state when single funding period display is rendered

JordanGuinn:
## Background
I'm still coming up to speed with the idea of immutable state in React, but this is my understanding of what's been happening:

It looks like whenever a funding source with a single option was selected, the corresponding component would directly update the form data and trigger a re-render of the parent component, thereby re-rendering the single option funding source component again and kicking off an endless loop.  Using immer's `produce` as we do with a good number of our other form values looks to have resolved it.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1230

## Validation Plan
- [ ] The following is confirmed when **adding a new child record**:
   - [ ] Ensure all age group/funding source combinations (specifically **single** funding source options) can be selected without killing the browser
   - [ ] Ensure funding space data is properly saved
- [ ] The following is confirmed when **changing a child's enrollment**:
   - [ ] Ensure all age group/funding source combinations (specifically **single** funding source options) can be selected without killing the browser
   - [ ] Ensure funding space data is properly saved
- [ ] The following is confirmed when **changing a child's funding**:
   - [ ] Ensure all age group/funding source combinations (specifically **single** funding source options) can be selected without killing the browser
   - [ ] Ensure funding space data is properly saved

## Automated Testing


bamader (approved):
I am like 10,000% sure this will solve it. Form errors of borked loading have been fixed by good ol' `updateData(produce(set()))` in so many ways. Will comment again after testing on my local.

client/src/components/Forms/Enrollment/Funding/Fields/ContractSpace/SingleContractSpace.tsx
```diff
@@ -31,14 +32,14 @@ export const SingleContractSpaceField = <
   fundingSpace,
   accessor,
 }: SingleContractSpaceProps<T>) => {
-  const { updateData, dataDriller } = useGenericContext<T>(FormContext);
+  const { data, updateData, dataDriller } = useGenericContext<T>(FormContext);
```
jlhogan, line 35: you can now pull `immutableUpdateData` from FormContext, which is wrapper around the produce/set pattern used below!

client/src/components/Forms/Enrollment/Funding/Fields/ContractSpace/SingleContractSpace.tsx
```diff
@@ -31,14 +32,14 @@ export const SingleContractSpaceField = <
...
   const currentFundingSpace = accessor(dataDriller);
 
   useEffect(() => {
     if (currentFundingSpace.at('id').value !== fundingSpace.id) {
-      updateData((_data) =>
-        set({ ..._data }, currentFundingSpace.path, fundingSpace)
```
jlhogan, line 40: If this fixes it then this fixes it and it definitely is better HOWEVER due to this object de+re-structuring, the original immutable `_data` object _shouldnt_ be being changed....

JordanGuinn: I _believe_ de-structuring perform a shallow copy, not a deep copy.  So if any of the parameters in `data` reference an object, then I _think_ that, even when de-structured, the reference is still kept intact 😬

client/src/components/Forms/Enrollment/Funding/Fields/ContractSpace/SingleContractSpace.tsx
JordanGuinn: Woot!  I'll update

bamader:
Jordan imma have to test again when you make whatever updates you're gonna do, cuz right now I have managed to explode the app 😂

bamader:
More specifically, I think Julia may be onto something with the de- and re-structuring thing; in create record, I made a child and got through to the enrollment/funding section, then started changing age groups and contract fields with reckless abandon to see if I could crash anything. All good so far. When I tried to save the record to the roster, the app errored out and told me it `couldn't read property 'uniqueIdType' of undefined`. So I think something in the data modification might have gone wonky. If it was something else, though, testing with your next fix should eliminate that as a concern source, so I'll investigate more then!

client/src/components/Forms/Enrollment/Funding/Fields/ContractSpace/SingleContractSpace.tsx
```diff
@@ -31,14 +32,14 @@ export const SingleContractSpaceField = <
   fundingSpace,
   accessor,
 }: SingleContractSpaceProps<T>) => {
-  const { updateData, dataDriller } = useGenericContext<T>(FormContext);
+  const { data, updateData, dataDriller } = useGenericContext<T>(FormContext);
   const currentFundingSpace = accessor(dataDriller);
 
   useEffect(() => {
```
vonbearshark, line 38: This is not helpful, but why do we need `useEffect` here at all? Still trying to trace the paths of the data, but is there a way to explicitly update the info elsewhere maybe?

JordanGuinn: I had that very same thought while I was doing this - there's assuredly a way to capture the selected funding space in the form outside of this otherwise read-only component, but I honestly just didn't want to invest the time to do that (would rather get this out and unblock folks)

vonbearshark: that makes sense to me!

client/src/components/Forms/Enrollment/Funding/Fields/ContractSpace.tsx
```diff
@@ -60,6 +60,7 @@ export const ContractSpaceField = <
     );
 
     if (fundingSpaceOptions.length === 1) {
+      console.log('UPDATING THE FORM');
```
vonbearshark, line 63: Seems like you're having a real one tonight. If you wanted to find some time I am free like all day tomorrow to pair on this together?

client/src/components/Forms/Enrollment/Funding/Fields/ContractSpace/SingleContractSpace.tsx
JordanGuinn: WELL, I _tried_ using `immutableUpdateData`, but for reasons that I'm still not clear on, the use of that function caused the majority of the form fields to get reset while using `updateData` does not 😐 I'm sure there's something I'm missing (given under the hood the behavior looks _identical_), but also ehhhh it works

client/src/components/Forms/Enrollment/Funding/Fields/ContractSpace.tsx
JordanGuinn: I think I finally got it!  Took a while before I remembered that setting state is **not** a synchronous operation, even though it reads like it is 😅

Would def appreciate another set of eyes on this again, though!  Long-term, this ticket in particular has me thinking that it might worth revisiting our form field updating at some point.  It could just be my lack of familiarity with certain React-y things, but this felt like it didn't need to be this complicated.

client/src/components/Forms/Enrollment/Funding/Fields/ContractSpace.tsx
```diff
@@ -58,11 +64,16 @@ export const ContractSpaceField = <
   }, [ageGroup, fundingSource, fundingSpaces]);
 
   if (fundingSpaceOptions.length === 1) {
+    updateData(
+      produce<T>(data, (draft) => set(draft, fundingAccessor(dataDriller).at('fundingSpace').at('id').path, fundingSpaceOptions[0].id))
+    );
```
bamader, line 69: Woof it sucks that we needed this specific of an accessor to get the right update pattern. I still can't even tell what was wrong with it before 😕

bamader (approved):

bamader:
From my POV this is validated and good to go, but I also didn't see the clearing stuff when using SmartState last time. If anyone else wants to validate, feel free, but otherwise I think Jordan's suffering on this can come to an end.

client/src/components/Forms/Enrollment/Funding/Fields/ContractSpace.tsx
```diff
@@ -48,6 +52,8 @@ export const ContractSpaceField = <
     FundingSpace[]
   >([]);
 
+  const { data, immutableUpdateData, dataDriller, updateData } = useGenericContext<T>(FormContext);
```
jlhogan, line 55: why not use the immutable update data?

jlhogan (approved):
this works; i still think use the immutable fun from th eform context over rewriting the produce/set here, but not more important than getting this fix out!

client/src/components/Forms/Enrollment/Funding/Fields/ContractSpace.tsx
JordanGuinn: Talked a bit about it [here](https://github.com/ctoec/data-collection/pull/1233#discussion_r592066798), but essentially I tried using that function and it kept resetting all the values of the change enrollment form.

jlhogan: that's bad and weird....

JordanGuinn: Yep!  I'm baffled by it, cuz under the hood `immutableUpdateData` is _identical_ to what I've written here 😬 It's for sure worth investigating, but probably not for _this_ ticket imo

vonbearshark:
merge/deploy totally makes sense. Following up, even if we push the actual refactor out we might want to sprinkle some `TODO`s in the meantime

Issue #1198, opened 2021-02-24
Full Time/Full Day and Part Time/Part Day funding spaces should be mapped equivalently


JordanGuinn:
## Background
The `Full time` and `Full day` funding spaces are pretty often used interchangeably within ECE Reporter, but aren't being treated as equivalent within file upload, naturally causing confusion for users.  The dropdown filter within the application already correctly provides them with the appropriate source time.

## Acceptance Criteria
- On file upload, both:
   - Time and Day are ignored in file uploads, and the `full` or `part` portion are used to determine the **correct** funding space

PR #1234, opened 2021-03-10
Make day and time equivalent in funding type mappings

bamader:
## Background
Our users interchangeably use 'day' and 'time' to describe funding types when speaking of full-day, part-time, etc. However, the app treats this as different things, and so flags pretty dumb things as errors when users upload their sheets. This PR forces the various permutations of full/part and day/time to map equivalently to the same root so they are safely processed no matter which form the user chooses.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1198

## Validation Plan
I've verified that the changes to how we're parsing these types (1) correctly interprets permutations for CDC fundings, (2) correctly interprets permutations for SR fundings, and (3) does _not_ break the other funding types, some of which may use full or party to describe years or school wraparounds and such. I used this spreadsheet to test all combinations of CDC and SR parsing and then verified that the system picked the correct type for each one, regardless of the sheet's actual text there.
[funding_mapping_test.xlsx](https://github.com/ctoec/data-collection/files/6119195/funding_mapping_test.xlsx)

## Automated Testing

## Additional Context
I noticed that the AC said things should be saved and displayed as [full/part] DAY for both CDC and SR fundings, but I don't actually remember that being the way our conversation about it went in estimation. I could 100% be wrong here, in which case there's some more work to do, but I remember Julia (correctly) calling out that when users were in the create flow or changing enrollment and funding, the system would already show the correct options for each funding type. I _thought_ we then came to the consensus that the only work that needed to be done, therefore, was in making the upload parsing more forgiving to the various permutations, rather than in enforcing that funding types always show up as full day or part day. Like I said, I could totally be wrong, so if that's not the case just lemme know!


src/controllers/enrollmentReports/map/mapUtils/fundingTime.ts
```diff
@@ -18,11 +23,26 @@ export const mapFundingTime = (
...
 };
+
+/**
+ * We only want to use the root 'full' and 'part' words in all of the
+ * full-day/full-time/part-day/part-time permutations, so this throws
+ * out the 'day'/'time' part that leads to unnecessary mixups.
+ */
+const trimToFullAndPartRoots = (value: string) => {
+  return value
+    .replace(FT_REGEX, 'full')
```
vonbearshark, line 44: Are these values literally just the `full`/`part` permutations? Like, can we just pull `full` or `part` from the string, or do we need to a `replace` with other things in the string?

vonbearshark: Also, does this handle `full-day` / `full day` or just `fullday`? (And does that matter?)

vonbearshark (approved):

src/controllers/enrollmentReports/map/mapUtils/fundingTime.ts
bamader: Yup this handles both of those cases! Because the string value that gets passed to this function has already gone through a `normalizeString` function which handles spaces and punctuation characters. 

I also thought the same thing you did at first, which was "hey let's just pick out only the full/part parts" because that's simplest. But, that actually doesn't work because there are some funding times for SHS and SS (which aren't covered in this ticket scope) that are called full-_year_ and part-_year_, as well as full/part-_schoolday_. Those funding times we actually _don't_ want to be interchangeably parsed with the day/time stuff because they're for entirely different funding sources and if we tried to assign full-time or full-day to a full-year that would be ineligible.

vonbearshark: 1. ug but 2. then yeah this makes sense, ha

JordanGuinn (approved):

JordanGuinn:
@bamader I think you're correct - there are no `*-day` times for CDC funding spaces, but rather `*-time`, while the inverse is true for PSR funding spaces.  I remember too that we just wanted to refer to the appropriate funding space and use whatever time was specified there, so I'll update the AC accordingly.

JordanGuinn:
Confirmed that, in file uploads:
- [ ] CSR funding spaces with `Part-time` space types are mapped to `Part-day`
- [ ] CSR funding spaces with `Part-day` space types are treated as `Part-day`
- [ ] CSR funding spaces with `Full-time` space types are mapped to `Full-day`
- [ ] CSR funding spaces with `Full-day` space types are treated as `Full-day`
- [ ] PSR funding spaces with `Part-time` space types are mapped to `Part-day`
- [ ] PSR funding spaces with `Part-day` space types are treated as `Part-day`
- [ ] PSR funding spaces with `Full-time` space types are mapped to `Full-day`
- [ ] PSR funding spaces with `Full-day` space types are treated as `Full-day`
- [ ] CDC funding spaces with `Part-day` space types are mapped to `Part-time`
- [ ] CDC funding spaces with `Part-time` space types are treated as `Part-time`
- [ ] PSR funding spaces with `Full-day` space types are mapped to `Full-time`
- [ ] PSR funding spaces with `Full-time` space types are treated as `Full-time`

Issue #1212, opened 2021-03-03
Improve upload error logs


jlhogan:
## Background
Currently, very little context about the cause of upload failures is available to dev/support. If we can improve ability to diagnose issues in uploads, we can reduce the lift required to help users resolve these issues. 

Logging out the children with validation errors that gets turned into the error response map would give devs/support insight into what errors are preventing user from uploading.

One thing to note -- currently, there are a LOT of logs because of all the query logs. Things get separated from the request log which has the identifying user id. at the very least, we should log user id again with this info. Will need to also think about how to format this data so that it's all visible (not just log out the raw POJO, b/c you'll get a lot of `[Object]` sort of things), and easily consumable (i.e. probably best if it's not broken up over multiple lines in CloudWatch, which can be very annoying to interact w/ programmatically). Maybe just one really long json string is best? And we can then grab and parse locally w a jq sort of thing as necessary.

## Acceptance Criteria
- A team member with access to CloudWatch logs can determine the cause of errors in an upload failure

PR #1248, opened 2021-03-18
1212: Log children with errors when upload has errors

jlhogan:
## Background
When a user attempts an upload and their data has errors, it is very hard to diagnose their issues. This adds logging (with redacted PII) to hopefully make it easier for us to debug + resolve data errors without requiring users to send us their data files.

## GitHub Issue
#1212 

## Associated PRs
N/A

## Validation Plan
Confirm logging in deployed envs when upload with errors is attempted

## Automated Testing

### DIFF
--- a/src/routes/enrollmentReports.ts
+++ b/src/routes/enrollmentReports.ts
@@ -72,7 +72,7 @@ enrollmentReportsRouter.post(
           childrenWithErrors
         );
 
-        // Log out the children with errors, without PII, to enable
+        // Log children with errors, without PII, to make troubleshooting easier
         if (enrollmentColumnErrors?.length) {
           logUploadErrors(req.user.id, childrenWithErrors);
         }
@@ -95,26 +95,27 @@ enrollmentReportsRouter.post(
   })
 );
 
+/**
+ * Helper function to log children with validation errors,
+ * without PII
+ * @param userId
+ * @param childrenWithErrors
+ */
 const logUploadErrors = (userId: number, childrenWithErrors: Child[]) => {
   console.log(
     `User ${userId} uploaded sheet with errors: ${JSON.stringify(
       childrenWithErrors
-        .filter(
-          (child) => !!child.validationErrors && child.validationErrors.length
-        )
-        .map((child) => {
-          const childToLog: object = child;
-          childToLog['lastName'] = child.lastName?.charAt(0);
-          childToLog['birthCertificateId'] = child.birthCertificateId?.replace(
-            /./g,
-            '#'
-          );
-          childToLog['birthdate'] = child.birthdate?.format('MM/YYYY');
-          childToLog['family'][
-            'streetAddress'
-          ] = child.family?.streetAddress?.replace(/./g, '#');
-          return childToLog;
-        })
+        .filter((child) => child.validationErrors?.length)
+        .map((child) => ({
+          ...child,
+          lastName: child.lastName?.charAt(0),
+          birthCertificateId: child.birthCertificateId?.replace(/./g, '#'),
+          birthdate: child.birthdate ? '##/##/####' : undefined,
+          family: {
+            ...child.family,
+            streetAddress: child.family?.streetAddress?.replace(/./g, '#'),
+          },
+        }))
     )}`
   );
 };
--- a/src/routes/enrollmentReports.ts
+++ b/src/routes/enrollmentReports.ts
@@ -109,7 +109,9 @@ const logUploadErrors = (userId: number, childrenWithErrors: Child[]) => {
         .map((child) => ({
           ...child,
           lastName: child.lastName?.charAt(0),
-          birthCertificateId: child.birthCertificateId?.replace(/./g, '#'),
+          birthCertificateId: child.birthCertificateId
+            ?.toString()
+            .replace(/./g, '#'),
           birthdate: child.birthdate ? '##/##/####' : undefined,
           family: {
             ...child.family,
--- a/src/routes/enrollmentReports.ts
+++ b/src/routes/enrollmentReports.ts
@@ -10,7 +10,7 @@ import {
   BatchUploadResponse,
   EnrollmentColumnError,
 } from '../../client/src/shared/payloads';
-import { ChangeTag } from '../../client/src/shared/models';
+import { ChangeTag, Child } from '../../client/src/shared/models';
 
 const CHANGE_TAGS_DENOTING_UPDATE = [
   ChangeTag.AgedUp,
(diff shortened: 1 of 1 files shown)


src/routes/enrollmentReports.ts
```diff
@@ -88,6 +95,30 @@ enrollmentReportsRouter.post(
   })
 );
 
+const logUploadErrors = (userId: number, childrenWithErrors: Child[]) => {
```
jlhogan, line 98: should this live not in this file?

src/routes/enrollmentReports.ts
```diff
@@ -62,18 +62,25 @@ enrollmentReportsRouter.post(
...
+            }),
           }))
         );
 
-        const errorDict: EnrollmentColumnError[] = await controller.checkErrorsInChildren(
+        const enrollmentColumnErrors: EnrollmentColumnError[] = await controller.checkErrorsInChildren(
           childrenWithErrors
         );
 
+        // Log out the children with errors, without PII, to enable
```
vonbearshark, line 75: "Log out", like log not like logout, right?

src/routes/enrollmentReports.ts
vonbearshark: Do we have a logging util already? Otherwise seems fine to me

src/routes/enrollmentReports.ts
```diff
@@ -88,6 +95,30 @@ enrollmentReportsRouter.post(
   })
 );
 
+const logUploadErrors = (userId: number, childrenWithErrors: Child[]) => {
+  console.log(
+    `User ${userId} uploaded sheet with errors: ${JSON.stringify(
+      childrenWithErrors
+        .filter(
+          (child) => !!child.validationErrors && child.validationErrors.length
```
vonbearshark, line 103: nit: `child.validationErrors?.length` ?

src/routes/enrollmentReports.ts
```diff
@@ -88,6 +95,30 @@ enrollmentReportsRouter.post(
...
+        )
+        .map((child) => {
```
vonbearshark, line 105: nit:

```
.map((child) => ({
  ...child,
  lastName: child.lastName?.charAt(0),
  birthCertificateId: child. birthCertificateId?.replace
  // etc
});
```

src/routes/enrollmentReports.ts
```diff
@@ -88,6 +95,30 @@ enrollmentReportsRouter.post(
...
+          const childToLog: object = child;
+          childToLog['lastName'] = child.lastName?.charAt(0);
```
vonbearshark, line 107: Do we / do we expect to filter this pii from logs elsewhere? Maybe worth it's own `removeChildPII` or something?

src/routes/enrollmentReports.ts
vonbearshark: If we do end up with a separate `rmChildPII` fn I am not sure this fn here is necessary versus just logging in the `if`, but 🤷🏻‍♀️

vonbearshark (approved):

src/routes/enrollmentReports.ts
bamader: Lol yes I thought the exact same thing when I saw the commit title 😂 . I was like when did we give children the ability to manage themselves in the system.

src/routes/enrollmentReports.ts
bamader: I think if we wanted to be really finicky with organization it shouldn't. In the past we've tried to limit the code that's actually in the `routes` file to just endpoints with minimal exposure to other logic. I personally think this is fine in the file, but from a nit perspective we maybe should move it.

src/routes/enrollmentReports.ts
bamader: This is the first I'm seeing of any function for removing PII, so my guess is no

src/routes/enrollmentReports.ts
```diff
@@ -88,6 +95,30 @@ enrollmentReportsRouter.post(
...
+          childToLog['birthCertificateId'] = child.birthCertificateId?.replace(
+            /./g,
+            '#'
+          );
+          childToLog['birthdate'] = child.birthdate?.format('MM/YYYY');
```
bamader, line 112: This might be overkill, but is keeping the month still leaving too much PII in the logged object?

src/routes/enrollmentReports.ts
```diff
@@ -88,6 +95,30 @@ enrollmentReportsRouter.post(
...
+          childToLog['family'][
+            'streetAddress'
+          ] = child.family?.streetAddress?.replace(/./g, '#');
```
bamader, line 115: My guess is no, but are income determinations considered protected information of the family?

bamader (approved):

src/routes/enrollmentReports.ts
jlhogan: will change the language :)

src/routes/enrollmentReports.ts
jlhogan: b.c i'm overwriting things in nested objects, would have to be:
```
{
  ...child,
  //etc,
  family: {
    ...family,
    // etc
  }
}
``` 
which to me is annoying. thoughts?

src/routes/enrollmentReports.ts
jlhogan: right now we don't log child objects anywhere else. all PII filtering happens at the DB query level.

src/routes/enrollmentReports.ts
jlhogan: yeah i'll just replace this with # as well

src/routes/enrollmentReports.ts
jlhogan: maybe but at this point how could u know what family it is

src/routes/enrollmentReports.ts
jlhogan: jk i wrote it out and it looks much better :)

vonbearshark (approved):

src/routes/enrollmentReports.ts
JordanGuinn: If we're literally only using this here then ehhhhh

JordanGuinn (approved):

JordanGuinn:
Ran the e2e test suite against this a couple times, and saw error crop up in Cloudwatch logs both times:

```
Unable to determine validation errors in spreadsheet: TypeError: _b.replace is not a function
at /var/app/current/dist/src/routes/enrollmentReports.js:114:234
at Array.map (<anonymous>)
at logUploadErrors (/var/app/current/dist/src/routes/enrollmentReports.js:112:10)
at /var/app/current/dist/src/routes/enrollmentReports.js:88:17
at Generator.next (<anonymous>)
at fulfilled (/var/app/current/dist/src/routes/enrollmentReports.js:24:58)
at runMicrotasks (<anonymous>)
at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

Also manually attempted to upload CSVs with fields missing, and that initial `api/enrollment-reports/check` call keeps returning 400s for what appears to be the same reason.

jlhogan:
yay for tests catching real issues!!!!!
weird that my tests with missing fields didn't trigger 
fixing now :)

vonbearshark (approved):

JordanGuinn:
Confirmed that uploading spreadsheets with missing columns now works, and the rows with validation errors are successfully written to Cloudwatch.  I'd say this is ready to go 🚀

jlhogan:
tests passed I just accidentally triggered _another_ rerun, and then cancelled it which is marking them as cancelled here. Passing run: https://app.circleci.com/pipelines/github/ctoec/data-collection/3137/workflows/0f3b22c9-42fc-475e-837c-42db104ebfbf

Issue #1190, opened 2021-02-21
Multi-org users generate lots of unnecessary API/RDS activity when retrieving report data


JordanGuinn:
## Background
Recently, the ECE Reporter ALB was unresponsive for a minute or two earlier this morning, and digging through Cloudwatch logs and metrics, it looks like it was the result of a sudden burst of API (and thereby RDS) activity.  The biggest culprit I see is the oec-report endpoint, and more specifically, how it’s called once per each organization that a user has permission to access.  This can’t scale effectively, as we already have a couple of users who have permission to access 18 organizations.  So when said users log into the application, they implicitly open **18 new, concurrent DB connections** and execute **18 SELECT queries** for _just_ oec report data for _just_ themselves.

The first and most immediate course of action is here to simply use a **single** API call to retrieve all OEC reports for a single user across all that user's organization.  But additionally, we should also consider beefing up our RDS hardware and seeing how much of an impact that has on situations like this.  Because while there's certainly an improvement to be had on the query-side here, at the end of the day this still wasn't _that_ much activity.

## Repro Steps
- Create a new user (or use an existing user)
- Configure the user to have 20+ organization permissions
- Log in to ECE reporter
- Navigate to the Home page
- Observe the number of `oec-report` API calls generated

## Acceptance Criteria
- OEC report data for a given user, across **all** organizations that user has access to, is provided in a **single** API call w/a **single** db query
- Enable [Performance Insights](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_PerfInsights.Overview.html) on our prod-like RDS instances (to get a better idea of any other queries that are particularly burdensome)
- Consider bumping our prod RDS instance from a **small** to a **medium or larger** ([it's the difference of like ~**$0.03/hour**](https://aws.amazon.com/rds/sqlserver/pricing/?pg=pr&loc=6))

JordanGuinn:
Did this get sprinted at some point?

jlhogan:
i rogue clicked, my bad, will return it to where it's supposed to be!!!

PR #1250, opened 2021-03-18
Make single endpoint instead of multiple calls

bamader:
## Background
This PR addresses the inefficiency of making an API call for each org a user is associated with, changing that instead to be a single API call that uses different SQL to fetch and check all accessible organizations at once.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1190

## Associated PRs
- https://github.com/skylight-hq/ctoec-devops/pull/206

## Validation Plan
Validation involved testing the home page as both a pre and a post-submit user. I ensured that the correct number of API calls were made in each case, and that the page was correctly shown to a user based on their submission status.

## Automated Testing
- [x] If applicable, unit tests have been added to cover this changeset (I did modify the `Home.test.tsx` file to wrap the page in an `AuthenticationContext.Provider`, since the test needs to also mock an `accessToken` now to trigger the call).

### DIFF
--- a/src/routes/oecReport.ts
+++ b/src/routes/oecReport.ts
@@ -17,7 +17,7 @@ oecReportRouter.get(
 oecReportRouter.get(
   `/:organizationId`,
   passAsyncError(async (req: Request, res: Response) => {
-    const organizationId = Number(req['organizationId']);
+    const organizationId: number = Number(req.params['organizationId']);
     const foundOrg = await getManager().findOne(OECReport, {
       where: { organizationId },
     });
--- a/src/routes/oecReport.ts
+++ b/src/routes/oecReport.ts
@@ -7,9 +7,17 @@ import { passAsyncError } from '../middleware/error/passAsyncError';
 export const oecReportRouter = express.Router();
 
 oecReportRouter.get(
-  '/:organizationId',
+  '/are-all-orgs-submitted',
   passAsyncError(async (req: Request, res: Response) => {
-    const organizationId: number = Number(req.params['organizationId']);
+    const allSubmitted = await controller.checkIfAllOrgsSubmitted(req.user);
+    res.send({ allSubmitted });
+  })
+);
+
+oecReportRouter.get(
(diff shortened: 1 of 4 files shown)


JordanGuinn (approved):

src/routes/oecReport.ts
```diff
@@ -7,9 +7,17 @@ import { passAsyncError } from '../middleware/error/passAsyncError';
 export const oecReportRouter = express.Router();
 
 oecReportRouter.get(
-  '/:organizationId',
+  '/are-all-orgs-submitted',
```
jlhogan, line 10: why not just have the default  `/` route ?

jlhogan (approved):

src/routes/oecReport.ts
vonbearshark: ++

vonbearshark (approved):

src/routes/oecReport.ts
JordanGuinn: This was my suggestion, namely because GET `api/oec-report` isn't indicative of what the API is returning (at least to me)

jlhogan: i have no strongly held opinion here so if you prefer then im all for it

JordanGuinn: If there are other naming conventions that make more sense to folks I'm all for them

JordanGuinn:
Confirming the following during testing:

- [ ] User with 20+ organizations is routed to the Pre-Submit home page when **none** have been submitted
- [ ] RDS CPU usage remains low in the aforementioned situation
- [ ] User with 20+ organizations is routed to the Pre-Submit home page when **some** have been submitted
- [ ] RDS CPU usage remains low in the aforementioned situation
- [ ] User with 20+ organizations is routed to the Pre-Submit home page when **all** have been submitted
- [ ] RDS CPU usage remains low in the aforementioned situation

vonbearshark (approved):
whoopsie

PR #1252, opened 2021-03-22
Disable SQL logging by default

JordanGuinn:
## Background
Our current CloudWatch logs are ~99% SQL queries at this point, making it very difficult to find any other more meaningful logs.  This PR disables SQL logs across the board so we can do so moving forward.

## GitHub Issue
N/A

## Associated PRs
N/A

## Validation Plan
- [x] Deploy to test environment
- [x] Smoke test and confirm everything's functional
- [x] Confirm SQL logs stop getting written

## Automated Testing

### DIFF
--- a/ormconfig.js
+++ b/ormconfig.js
@@ -12,7 +12,7 @@ module.exports = {
   database: process.env.DB_NAME || 'master',
   synchronize: false,
   migrationsRun: true,
-  logging: true,
+  logging: false,
   entities: ['dist/src/entity/**/*.js'],
   migrations: ['dist/src/migration/*.js'],
   subscribers: ['dist/src/subscriber/**/*.js'],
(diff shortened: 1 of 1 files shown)


ormconfig.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
   database: process.env.DB_NAME || 'master',
   synchronize: false,
   migrationsRun: true,
-  logging: true,
+  logging: false,
```
vonbearshark, line 15: shouldn't this be something like `process.env.ENABLE_SQL_LOGS || false`, in case we want to enable?

vonbearshark (approved):

ormconfig.js
JordanGuinn: Yeah, I was thinking that same thing.  We'll just have to parse it from a string (since these env vars can't be booleans), and then I guess be diligent about how we set it.

vonbearshark (approved):

bamader (approved):
So happy to see these disappear from Cloudwatch 😭

Issue #1187, opened 2021-02-19
Previous month roster view does not show all children w/active enrollments


jlhogan:
## Background
A user just reported that she was trying to use the `Roster for a past month` filter to confirm her enrollment data was correct.  However, a student withdrawn in **January 2021** (with enrollment dates in the tool confirmed) was not showing up in the **December 2020** filtered roster view.

## Acceptance Criteria
- ALL children with an active enrollment **within the filtered month** are visible in the filtered roster

## Additional Context
- **Active enrollment** = An enrollment without an explicit `exit` date

vonbearshark:
Investigated the code and it looks like on the backend we are correctly filtering the students according to the intention here. But then we are seemingly filtering _again_ on the frontend, separating out `active` and `withdrawn` children. This seems intentional. For what it's worth the condition is `child.enrollments?.every((enrollment) => !!enrollment.exit)`, which does match

> Active enrollment = An enrollment without an explicit exit date

technically. But not 

> active enrollment *within the filtered month*

Wouldn't a student withdrawn in January 2021 rightfully show up as "withdrawn" in the December 2020 filtered roster view? Is the issue that withdrawn students aren't listed in the roster view?

vonbearshark:
Or is the definition of `active` actually: either no `exit` or an exit in the future (past the filtered month)?

vonbearshark:
But I am also running into the potential bug: trying to filter to `Withdrawn records` shows no records...

JordanGuinn:
@vonbearshark We talked about this directly over Zoom, but putting here for tracking purposes.

Here's my understanding of what those 3 distinct views are supposed to reflect (@jlhogan plz correct me if this seems off):

- **Default Roster View:** Show all children that have **at least one** enrollment without an exit date
- **Withdrawn Records View:** Show all children that **don't have any** enrollments without an exit date
- **Filtered by Month View:** Show all children that either:
   - Have **at least one** enrollment without an exit date
   - Have **at least one** enrollment with an exit date _after_ the filtered month

vonbearshark:
Just sharing my thinking: that makes the `withdraw` and `month` query params mutually exclusive. It's a little hard for me to decipher, but I don't think that's the way the code is organized right now, regardless of this issue

jlhogan:
above sounds correct to me. Perhaps theres a way to simplify this a bit by fitting the default and withdrawn rosters into a time-based filtering definition:
[...]

PR #1253, opened 2021-03-23
Roster refactor

vonbearshark:
## Background

I link to relevant tickets, but the changes have broader implications, as I ended up touching most of the Roster code.

## GitHub Issue
#1265 
#1187 

## Associated PRs
N/A

## Validation Plan
- [x] Previous month active filter works, displays the correct children (#1187)
- [x] Withdrawn filter works, displays correct children
- [x] Withdrawn and month filters are mutually exclusive (clicking `withdrawn` from month resets state to `withdrawn`)
- [x] Site filtering still works as expected (including for multiorg users)
- [x] Multiorg filtering works as expected
- [x] For multiorg users:  First load of roster filters the initial `All Sites` to the appropriate org's data
- [x] Batch editing works?
- [x] Back buttons work in the Roster view and Child views (create, edit, etc.)

## Automated Testing
- [x] If applicable, unit tests have been ~added~ updated to cover this changeset.
- [x] If applicable, e2e tests have been ~added~ updated to cover this changeset.


vonbearshark:
We can _probably_ review merge these changes independent of the `TODO`s (without pagination, filtering, sql-izing), making this just an intermediary refactor step. Or by @jlhogan 's suggestion we could keep this as a feature branch and add those features in

vonbearshark:
@JordanGuinn I tried to cover all the changes from https://github.com/ctoec/data-collection/pull/1136/files, but lmk what I missed

client/src/containers/App/App.tsx
```diff
@@ -5,7 +5,6 @@ import { ErrorBoundary, Header } from '@ctoec/component-library';
 import routes from '../../routes';
 import MakeRouteWithSubRoutes from './MakeRouteWithSubroute';
 import UserContext from '../../contexts/UserContext/UserContext';
-import { mailToLinkProps } from '../../components/MailToLink';
```
vonbearshark, line 8: unrelated, unsued var

src/utils/processChild.ts
```diff
@@ -78,3 +65,97 @@ function distributeValidationErrorsToSubObjects<
...
+ * 	- remove deleted entities
+ * 	- sort entities by date
+ * 	- validate full object tree
+ * @param child
+ */
+export const postProcessChild = async (child: Child) => {
+  removeDeletedEntitiesFromChild(child);
+  sortEntities(child);
+  return validateObject(child);
+};
```
vonbearshark, line 161: Consolidated these helper fns here, otherwise untouched

client/src/containers/Home/PostSubmit.tsx
```diff
@@ -34,33 +34,26 @@ export const PostSubmitHome: React.FC = () => {
   ] = useState<NestedFundingSpaces>();
   const [siteCountDisplay, setSiteCountDisplay] = useState();
   useEffect(() => {
-    apiGet('children?count=true', accessToken)
-      .then((res) => setUserRosterCount(res.count))
-      .catch((err) => {
-        throw new Error(err);
-      });
+    apiGet(`children/metadata?showFundings=${showFundings}`, accessToken)
```
vonbearshark, line 37: was able to consolidate these calls into the `/metadata` endpoint, where showing funding is optional

client/src/containers/Home/PreSubmit.tsx
```diff
@@ -38,10 +38,8 @@ export const PreSubmitHome: React.FC = () => {
     // Determine the funding spaces map for the organization, if
     // the user has the permissions that enable this
     if (showFundingsAndSites) {
-      apiGet('funding-spaces?fundingMap=true', accessToken)
-        .then((res) => {
-          setFundingSpacesDisplay(res.fundingSpacesMap);
-        })
+      apiGet('funding-spaces/funding-map', accessToken)
```
vonbearshark, line 41: funding map also has it's own endpoint, which uses the same controller fn as the `metadata` endpoint

src/controllers/enrollmentReports/formatUploadPreview.ts
```diff
@@ -5,7 +5,6 @@ import {
   getLastEnrollment,
   getLastFunding,
 } from '../../../client/src/utils/models';
-import { validate } from 'class-validator';
```
vonbearshark, line 8: unrelated, unused var

client/src/contexts/RosterContext/RosterContext.tsx
```diff
@@ -1,39 +1,98 @@
-import React, { useState } from 'react';
+import React, { useContext, useEffect } from 'react';
```
vonbearshark, line 1: Refactored this to handle the roster data: the query and the child records. Now one course of truth for all child/sibling components

client/src/contexts/RosterContext/RosterContext.tsx
```diff
@@ -91,15 +150,37 @@ const RosterProvider: React.FC = ({ children }) => {
...
+        updateWithdrawnQuery: (withdrawn: boolean) => {
+          history.push({
+            search: stringify({
+              ...query,
+              month: defaultCtx.query.month,
+              withdrawn,
+            }),
+          });
+        },
       }}
```
vonbearshark, line 180: Updating the query really means updating the history, since the `useEffect` at the top of his ctx will respond to those changes, updating the query state

src/routes/children.ts
```diff
@@ -23,37 +23,90 @@ childrenRouter.get(
...
+  passAsyncError(async (req, res) => {
+    const organizationIds = parseQueryString(req, 'organizationId', {
+      forceArray: true,
+    }) as string[];
+    const validationErrors = await controller.getErrorCounts(
+      req.user,
+      organizationIds
+    );
+
+    res.send(validationErrors);
```
vonbearshark, line 109: separate routes rather than one `/` route with entangled param paths

client/src/containers/Roster/hooks/usePaginatedChildData.tsx
```diff
@@ -2,7 +2,7 @@ import { stringify } from 'query-string';
 import { useAuthenticatedSWRInfinite } from '../../../hooks/useAuthenticatedSWR';
```
vonbearshark, line 2: No real changes here, but would like a place to talk about pagination. I am wondering how we want to handle the separation by `site`, specifically. Right now the client is handling the `site` filter. That _might_ be okay. But it does get trickier with pagination, I think. Like, what happens when we are on a `site`'s tab and we `load more` by age? what do we fetch exactly? Seems like there is some complexity there and interested in thoughts (@jlhogan in particular)

client/src/containers/Roster/Roster.tsx
```diff
@@ -198,60 +107,33 @@ const Roster: React.FC = () => {
...
-        // Decide if site is empty here to keep site-switching
-        // controls available
-        siteIsEmpty ? (
-          <EmptyRosterCard boldText="This site doesn't have any records yet" />
-        ) : (
-          <RosterContent {...rosterContentProps} />
-        )
+  // Organization filtering happens on the server-side,
+  // but site filtering needs to happen in the client-side, if a
+  // site is requested
```
vonbearshark, line 112: Worth talking about this. It would be nice to at least group by site on the backend? Might run up against pagination changes, though--see the other comment

src/controllers/children/get.ts
```diff
@@ -134,109 +236,93 @@ export const getSiteCountMap = async (children: Child[]) => {
...
-      (determinationA, determinationB) =>
-        propertyDateSorter(
-          determinationA,
-          determinationB,
-          (d) => d.determinationDate
-        )
+  return fundingSpacesDisplay;
+};
+
+export const getErrorCounts = async (user: User, organizationIds: string[]) => {
```
vonbearshark, line 287: TODO: SQL-ize. But wondering if that's even possible given how we manage validation errors? Aren't they meta information? Are they ever in the db?

src/controllers/children/get.ts
```diff
@@ -19,79 +80,116 @@ import { getCurrentEnrollment } from '../../utils/getCurrentEnrollment';
...
-    return children.filter((c) => {
-      // filter out enrollments after the current month filter
-      c.enrollments = c.enrollments?.filter(
-        (e) => e.entry && e.entry.isSameOrBefore(activeMonth.endOf('month'))
+    extendQB: (qb) => {
+      // Filter children with enrollments in the active month
+      const start = activeMonth.startOf('month').format('YYYY-MM-DD');
+      const end = activeMonth.endOf('month').format('YYYY-MM-DD');
+
+      // TODO filter and only return matching enrollments
```
vonbearshark, line 123: In the previous code we are only returning enrollments that match--is that actually necessary, though, as long as the conditions are satisfied?

src/utils/processChild.ts
```diff
@@ -1,23 +1,10 @@
+import { Child, Enrollment } from '../entity';
```
vonbearshark, line 1: Why do we need to process the Child entities? Should the processing be happening as we go, rather than as we are fetching the data?

vonbearshark (commented):
`get.ts`, `children.ts`, `RosterContext.tsx`, and `Roster.tsx` have the big changes

src/utils/processChild.ts
vonbearshark: If it is necessary then `TODO` I am still handling it in the `get.ts` refactor

src/routes/children.ts
```diff
@@ -23,37 +23,90 @@ childrenRouter.get(
...
+      req.user,
+      organizationIds
+    );
+
+    res.send(missingInfoChildren);
+  })
+);
+
+childrenRouter.get(
+  '/validation-errors',
```
vonbearshark, line 99: I don't love this extra call for error data. When implementing the pagination it is possible there is an initial request that has the first paginated payload _plus_ the error data. On one hand, we probably do want to have that error count updating as errors are resolved. Otoh that might not be happening now as we expect, anyway...

vonbearshark: Following that thought: the existing functionality here is 1 broken due to edge cases in `active` v `withdrawn` 2 handled in the client. Handling in the client is cumbersome, but it does allow for "live" updating of the active errors. I am not sure how to best account for that with a backend-based system, since we need a client condition to trigger a recompute of the error state. We could do it 1 on initial pagination payload + 2 on submit attempt, but it is not ideal since the count won't update in the meantime

src/controllers/children/get.ts
```diff
@@ -19,79 +80,116 @@ import { getCurrentEnrollment } from '../../utils/getCurrentEnrollment';
...
+export const getActiveChildren = async (
   user: User,
-  filterOpts: {
-    organizationIds?: string[];
-    missingInfoOnly?: boolean;
+  organizationIds?: string[],
+  selectParms?: {
     activeMonth?: Moment;
     skip?: number;
     take?: number;
```
vonbearshark, line 107: Am I missing something? I thought we needed to `ORDER BY` to use these params/pagination?

client/src/containers/Roster/hooks/usePaginatedChildData.tsx
vonbearshark: capturing form irl: mvp could easily be to just throw out data from site filtering pagination calls without sophistication around reconciling previously fetched/cached state

src/utils/processChild.ts
JordanGuinn: What do you mean by `happening as we go` exactly?

I know that all children displayed in the roster need to be processed (validated) so that we can display that little `Missing Info` icon for those rows, but once we start applying pagination to the roster queries, that'll only be necessary for the children within the page being queried for.

src/controllers/children/get.ts
JordanGuinn: I think it'd be **really, really** annoying to generate equivalent SQL queries out of the processing behavior we have defined here 😕

JordanGuinn: Also, I can't help but hear `SQL-ize` as [Sequelize](https://sequelize.org/) and get very sad at all the trouble TypeORM has given us 😅😭

src/routes/children.ts
JordanGuinn: Ugh, that's tough.  

I honestly don't think it would be an issue to be overly chatty with this API, if it didn't require pulling _all the children in the roster_ off row 😬 We wouldn't even get the benefit of piggybacking off the child SELECT when retrieving paged roster data either, cuz it'd require an entirely separate, unpaged query anyway.

JordanGuinn: Certainly not applicable to this PR, but maybe it's time we start thinking about storing this ephemeral roster error metadata somewhere

JordanGuinn (approved):
Okay, first pass at this PR complete.  Got a handful of questions/comments, but overall this is looking _real real_ good 🙏🏽 Also, I do think it addresses everything I had in my previous PR!  I can incorporate that testing work into validation here just to be extra sure, though

client/src/containers/Roster/Roster.tsx
jlhogan: documenting what we discussed: site filtered vs unfilitered data does not need to be smartly re-used; can move this filter into the backend and re-request individual pages based on what roster the user is looking at. IMO only worth doing (moving site filter into db-level in server) once we've updated pagination to let the user do it.

client/src/containers/Roster/Roster.tsx
```diff
@@ -273,44 +155,55 @@ const Roster: React.FC = () => {
...
-          {!query.withdrawn && (
-            <RosterButtonsTable
-              filterByMonth={queryMonth}
-              setFilterByMonth={updateActiveMonth}
-              updateWithdrawnOnly={updateWithdrawnOnly}
-            />
-          )}
-          <LoadingWrapper text="Loading your roster..." loading={loading}>
-            {rosterContent}
+          {!query.withdrawn && <RosterButtonsTable />}
```
jlhogan, line 177: this is not a change that you're making, but why are we hiding the buttons table in withdrawn view?

client/src/containers/Roster/hooks/usePaginatedChildData.tsx
jlhogan: heh i put a comment capturing same idea above. yes! we will get there! once this, and then better pagination, lands.

src/routes/children.ts
```diff
@@ -23,37 +23,90 @@ childrenRouter.get(
...
       forceArray: true,
     }) as string[];
-    const count = parseQueryString(req, 'count');
-    const missingInfo = parseQueryString(req, 'missing-info') as string;
     const activeMonth = parseQueryString(req, 'month', {
       post: (monthStr) => moment.utc(monthStr, 'MMM-YYYY'),
     }) as Moment;
     const skip = parseQueryString(req, 'skip', { post: parseInt }) as number;
     const take = parseQueryString(req, 'take', { post: parseInt }) as number;
-    const siteMap = parseQueryString(req, 'siteMap');
```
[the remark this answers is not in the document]
jlhogan: yeah no it's bad. but at some point i was annoyed that no nice typescripty thing for this existed and wrote one. but overall i agree that it's weird and not actually very useful.

jlhogan (approved):
Overall this gets us into a much better state. In terms of validating, as I mentioned I think the best way I can understand to have HIGH CONFIDENCE that we didnt introduce regression would be to test with prod data against prod views.

Also happy to hammer at it in QA once we push it there! I have at least a bit more of an idea of what/how to manually validate so we don't feel like we're just flailing in the dark :)

client/src/containers/Roster/Roster.tsx
bamader: Only thing I can come up with is that we didn't want users adding children with withdrawn enrollments from the create flow, but that doesn't seem right or like something we'd do

client/src/containers/Roster/hooks/useIsSubmitted.tsx
```diff
@@ -0,0 +1,25 @@
...
+  } = useContext(RosterContext);
+
+  const [isSubmitted, setIsSubmitted] = useState(false);
+  useEffect(() => {
+    if (!organizationId) return;
+
+    apiGet(`oec-report/${organizationId}`, accessToken).then((res) => {
+      setIsSubmitted(!!res?.submitted);
+    });
+  }, [organizationId, accessToken]);
```
bamader, line 22: How many users are there with access to multiple orgs? Since we don't show the post-submit version of the home page until _all_ of a user's orgs have submitted their data (which is represented by the same `oec-report` check), is that something we need to consider here?

src/controllers/children/get.ts
```diff
@@ -1,16 +1,77 @@
...
+import { intersection } from 'lodash';
+
+/**
+ * Get base query for getting children for given user. Applies organizationId filter,
+ * all organizations the user has read-access for.
+ * If user has site-level access, also applies a site filter to only return children
+ * and enrollments for those sites
+ * Accepts QueryBuilder extension function to manipulate the SELECT query.
+ */
+const findChildrenOpts = async ({
```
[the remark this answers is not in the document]
bamader: Typescript will yell at us if we ever leave an `await` that doesn't do anything (it flashes a warning like 'this await has no effect') so I think we tend to correctly use it? But also, I'm just paranoid about race conditions and synchronous work so I will await errything if I need to.

src/controllers/children/get.ts
```diff
@@ -1,16 +1,77 @@
...
+  user,
```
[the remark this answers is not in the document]
bamader: I think the last sentence is probably the most relevant analysis. I'd hate to put in all kinds of effort parameterizing something to accept different search queries when to me at least it's not awful to just invoke whichever of `find` vs `findOne` we need. Code laziness ftw (at least until it impacts other stuff--then burn it all 🔥 )

src/controllers/children/get.ts
```diff
@@ -1,16 +1,77 @@
...
+  organizationIds,
+  extendQB,
+}: {
+  user: User;
+  organizationIds?: string[];
+  extendQB?: (qb: SelectQueryBuilder<Child>) => void;
+}) => {
+  const readOrgIds = await getReadAccessibleOrgIds(user);
+  const orgIds = organizationIds?.length
+    ? intersection(readOrgIds, organizationIds)
```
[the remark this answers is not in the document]
bamader: I think a lot of the org filtering scattered around here is more of a pseudo-security thing (access control is probs a better phrase but I'm feeling dangerous today). Like, we want users who are associated with particular organizations to be the only ones actually making changes for those organizations. If a user doesn't have that `orgId` credential, then typeorm query won't return anything from `find` and their operation will have no effect.

src/controllers/children/get.ts
bamader: I do my best to not think about TypeORM at all.




Except...
Sometimes, late at night...
That's when the demons come... 👻

bamader: But also yes I think it would be a really big frustration to try to turn our validation processing into SQL 😞

bamader (approved):
Wow that took some time and some staring to grok. I think I've wrapped my head around everything, though, and I _think_ everything looks good. Obviously, it'll be subject to copious amounts of validation, so once we as a team start getting on that train, I'll be happy to manually try to explode everything.

client/src/containers/Home/PostSubmit.tsx
```diff
@@ -34,33 +34,26 @@ export const PostSubmitHome: React.FC = () => {
...
-        .then((res) => {
-          setFundingSpacesDisplay(res.fundingSpacesMap);
-        })
-        .catch((err) => {
-          throw new Error(err);
-        });
-    }
   }, [accessToken]);
 
+  console.log('here');
```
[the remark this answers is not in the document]
vonbearshark: Weird we don't warn about this? Or do we?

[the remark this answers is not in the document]
JordanGuinn: We don't warn about this, unfortunately 😕 While we **do** have loggers set up for SQL queries and incoming HTTP requests, we **don't** currently have a dedicated logger for application usage, so there are a handful of places where we're using `console.log` meaningfully.

Maybe it's time we finally get an app logger set up, though 😁

client/src/containers/Home/PostSubmit.tsx
vonbearshark: I am not sure I see the difference. In the previous code we would make the two other calls and conditionally make the funding call. Here we are making one call (in place of the previous two), and passing this conditional param, which optionally adds the controller call on the backend.right?

client/src/containers/Upload/ReviewMissingInfo.tsx
```diff
@@ -11,7 +11,6 @@ import { BackButton } from '../../components/BackButton';
 import { EnrollmentColumnError } from '../../shared/payloads';
 import { getH1RefForTitle } from '../../utils/getH1RefForTitle';
 import { misingInfoTableColumns } from './missingInfoTableColumns';
-import { ReactComponent as Arrow } from '@ctoec/component-library/dist/assets/images/uswds/arrowRight.svg';
```
vonbearshark, line 14: changes in this file unrelated

client/src/containers/Home/PostSubmit.tsx
vonbearshark: (possible I might have edited this afterward, losing track at this point)

src/utils/processChild.ts
vonbearshark: Ah. That is kind of frustrating but understandable. Right now we are doing

```
export const postProcessChild = async (child: Child) => {
  removeDeletedEntitiesFromChild(child);
  sortEntities(child);
  return validateObject(child);
};
```

Do we need to remove deleted entries or sort entries? Why isn't the former done elsewhere (wherever entries are deleted)? And is there a specific reason we do the latter here? I assume it's for convenience for the ui, but it might make sense elsewhere post refactor

src/controllers/children/get.ts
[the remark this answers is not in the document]
vonbearshark: Hm. We should satisfy that kind of security ish requirement with the `findChildrenOpts` fn, which only allows you to select from orgs you have access to, and it also defaults to all of your accessible orgs.

What about from a product perspective. How do we handle multi-org users?

Right now there is a nontrivial amount of frustrating code (especially in the frontend) around providing a mandatory org ID for all Roster queries. I suspect we should be able to eliminate that, and worst-case we could just allow for a specific org ID override for multi org users selecting a particular org. But I don't understand the app well enough to say for sure. @JordanGuinn @jlhogan maybe

src/controllers/children/get.ts
vonbearshark: I guess I mean: how are we getting away without it now at all? Like, shouldn't typeorm be pissed at least?

JordanGuinn: Oh I get you - I honestly think it's just dumb luck.  I know that using `OFFSET` and `FETCH` without `ORDER BY` in SQL Server is generally a really bad idea (as there's no implicitly ordering that can be relied upon consistently), but I don't know it to be technically impossible.

Least it sure doesn't seem like it is here 😅

src/utils/processChild.ts
JordanGuinn: 1. Agreed, sort should def happen at the db layer, because of course it should
2. Manually removing rows with `deleteDate` values because apparently TypeORM doesn't too well support implicit filtering of rows with such values.  We have [an outstanding question, in ticket form,](https://app.zenhub.com/workspaces/ece-dev-board-5ff506028d35f30012e0e937/issues/ctoec/data-collection/1174) to figure this out

src/controllers/children/get.ts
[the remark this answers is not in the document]
JordanGuinn: I haven't been able to confirm the user experience for multi-org users yet, but if there isn't a way for those types of users to be able to filter their roster contents by organization, then agreed, org filtering ought be an implicit permissions thing handled on the back-end and not based in any way on parameters provided by the client.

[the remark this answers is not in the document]
vonbearshark: > if there isn't a way for those types of users to be able to filter their roster contents by organization

Wow yeah I have been operating under the assumption that this was necessary but it could be totally not? geez that would really simplify this code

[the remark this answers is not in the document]
vonbearshark: capturing convo with @jlhogan 

> for multi-org users, we give them one org roster at a time. they can select which org by an additional UI nav component which is another level of tab nav
> so for the multi-org user we will pass an org param, and all other users we wont and they'll implicitly get "all" of their data but it will just be for one org

So this doesn't save us complexity here/on the backend, but it _should_ save us some frontend complexity where we are working to inject/force an org ID where we don't have to be all the time...

jlhogan:
@vonbearshark is this awaiting another round of reviews or any other interaction?

vonbearshark:
still working off feedback from yesterday and whack a mol-ing tests, but later today, I think

src/controllers/children/get.ts
vonbearshark: Did an investigation with @bamader. It _seems_* like the only place we really use enrollments as returned from this fn is in the Roster view, in the `Enrollment Date`. However, this is actually kind of broken, as is. We use this `Enrollment Date` in past (`month`) filters. And that doesn't make sense, really. Children can have multiple enrollments in a past filter (especially once we expand it beyond just a month). Right now we just pick the "newest" one, but that is kind of arbitrary? My pitch would be _not_ to filter the enrollments for the past, and instead have this field always show the most recent enrollment date. We might want to update the language to  `Current Enrollment Date` or `Most Recent Enrollment Date` (to apply to the `withdrawn` filter as well, where "current" doesn't make sense).

*I still need to investigate. I am wondering if the other fields in the Roster view, like `Funding type`, `Space type`, and `Site` are derived from the most recent enrollment as well. I am thinking that they might, and that would kind of open a can of worms. Again, then they are _all_ broken/arbitrary for `month` and `withdrawn` filters right now, and my pitch would probably be to make it clear that _all_ of the fields are `Most Recent`

does that make sense? Thoughts? (cc @jlhogan maybe)

vonbearshark: I guess on second thinking picking the most "current" enrollment that also satisfies the filter (`withdrawn` or `month`), is not that arbitrary?

src/routes/children.ts
vonbearshark: Slight spoilers, but I sqlized the child data fetching (not the metadata calls). I realized while doing that they we can really _never_ sqlize the child validation stuff. Like, in multiple cases we need to grab all of the children and count al of their errors. That will end up being a bottleneck on some routes.

client/src/containers/Roster/hooks/useIsSubmitted.tsx
vonbearshark: ug. Is it? hm..

vonbearshark:
New commit just dropped!

Please re-review the changes (and thank you!)

Big diffs: 
- SQL-ized all child retrieval in `get.ts` 
    - including the processing (like sorting entries and removing soft deletes)
    - _not_ including validation error-based things or metadata calls
    - Needed to convert to querybuilder format. lmk what you think there
    - Not a typeorm fan!
- Updated the orgId and site query params' mgmt in the frontend
    - No longer force orgId injection throughout
    - mgmt now happens in roster ctx, again as source of truth (rather than individual components handing `history` directly

I will try and comment throughout the changes. I have all tests running locally, and I have been doing some misc manually QA. Next steps would be: finish code reviewing, then start on validation.

src/controllers/children/get.ts
```diff
@@ -112,35 +134,25 @@ export const getActiveChildren = async (
...
-  });
-  const children = await Promise.all(
-    preProcessedChildren.map(postProcessChild)
+  // Filter children with enrollments in the active month
+  const start = activeMonth.startOf('month').format('YYYY-MM-DD');
+  const end = activeMonth.endOf('month').format('YYYY-MM-DD');
+
+  const qb = await queryBuilderBuilder({ user, organizationIds });
+  qb.andWhere(
+    'enrollment.entry <= :end AND (enrollment.exit >= :start OR enrollment.exit IS NULL)',
```
vonbearshark, line 144: Please confirm the thinking here. active for time = overlapping with that range at all, so if an enrollment starts before `end` of range and either ends after `start` of range or doesn't end at all.

src/controllers/children/get.ts
```diff
@@ -152,35 +164,24 @@ export const getWithdrawnChildren = async (
...
-    ...opts,
-    skip,
-    take,
-  });
-  const children = await Promise.all(
-    preProcessedChildren.map(postProcessChild)
-  );
-  return children;
+  const qb = await queryBuilderBuilder({ user, organizationIds });
+  qb.andWhere('enrollment.exit IS NOT NULL');
```
vonbearshark, line 169: Confirm the thinking here: withdrawn = children where all enrollments have been terminated

(not _exactly_ sure why this `where` works with the query, but it seems to?)

src/controllers/children/get.ts
```diff
@@ -1,76 +1,100 @@
...
-  };
+  // Sort enrollments (and funding) and income determinations
+  qb.orderBy({
+    'family_incomeDetermination.determinationDate': 'DESC',
+    'enrollment.entry': 'DESC',
+    'enrollment_funding.firstReportingPeriod': 'DESC',
+  });
 
-  return opts;
+  return qb;
```
vonbearshark: Maybe I have been staring at it too long, but is this so bad? It is tedious, but at least it's clear what is happening?

src/utils/processChild.ts
vonbearshark: double check, but I think I was able to add this to the sql-izing commit. essentially I left join _and_ do this sanitization when bringing in the relations

client/src/containers/Roster/hooks/useIsSubmitted.tsx
vonbearshark: I am not sure I see how we were handling that case before

src/controllers/children/get.ts
vonbearshark: (addressed this with the new sql-izing commit)

src/controllers/children/get.ts
bamader: Would definitely appreciate other thoughts on these conditionals, cuz JJ and I came up with them during pairing, but these make sense to me. We ran through a couple of scenarios of when records should and shouldn't qualify for this filter and everything seemed to us to check out.

src/controllers/children/get.ts
bamader: Yeah this should be the right query, to me. `.exit` is the date a child left an enrollment, and it defaults to null for all enrollments that the child is presently in. So, it's only _not_ null if a child has left the enrollment, hence the withdrawn status.

bamader (approved):
This looks good to me. I like the SQL changes that were made, as well as some improvements to the query parsing and general simplification. Pending some thorough manual validations, this looks great!

client/src/contexts/RosterContext/RosterContext.tsx
```diff
@@ -43,122 +41,28 @@ const defaultCtx = {
...
   if (error) throw error;
 
-  /**
-   * useSWR provides a `mutate` callback to update cache values, but I cannot
-   * get it to work for the inifite flavor of the hook. This is a workaround
-   * that directly mutates the cache that underlies the SWR stuff without
-   * any of the other logic baked in to the mutate func (logic around if the
-   * changes should actually be applied, I guess? because they're not)
-   * @param updatedChild
-   */
```
vonbearshark, line 45: Was able to [fix](https://github.com/ctoec/data-collection/pull/1253/commits/8908d981b6abbe0b75115643d74c82b605a7355b#diff-ff0631363477e3e64ca2787bda67ba6b16fe865a67d3fbc52ccb90a033a7289dR84) this I think

client/src/hooks/useAuthenticatedSWR.ts
```diff
@@ -20,7 +20,7 @@ export function useAuthenticatedSWR<T>(
 ) {
   const { accessToken } = useContext(AuthenticationContext);
   return useSWR(
-    !path ? null : [path, accessToken],
+    !path || !accessToken ? null : [path, accessToken],
```
vonbearshark, line 23: correct me if I am wrong here, but this is what we want for "authenticated"  calls in particular?

client/src/hooks/useAuthenticatedSWR.ts
```diff
@@ -39,7 +39,7 @@ export function useAuthenticatedSWRInfinite<T>(
 
   const getKey = (index: number, prevData: T | null) => {
     const path = getPath(index, prevData);
-    const key = !path ? null : [path, accessToken];
+    const key = !path || !accessToken ? null : [path, accessToken];
```
vonbearshark, line 42: again correct me if I am off base

client/src/containers/Roster/hooks/usePaginatedChildData.tsx
```diff
@@ -31,15 +35,12 @@ export const usePaginatedChildData = ({
       return null;
     }
 
-    // Paginated api query (but only once we have organizationId param)
```
vonbearshark, line 30: Doesn't seem to be a reason for this

client/src/containers/Roster/hooks/useIsSubmitted.tsx
vonbearshark: @jlhogan do we have a switch for multiorg users to only select one roster/org at a time? Talking with @bamader it might be a vague design sitch

src/controllers/children/get.ts
jlhogan: no i think it's definitely the better way to do it -- initially was trying to stick with the niceness of the "find" operations, but they quickly became way less "nice". All the TypeORM docs indicate that this is the intended/expected usage approach for more complex querying. Overall very agree tedious but clear

src/controllers/children/get.ts
jlhogan: why do we still have a `getWithdrawnChildren` endpoint? Shouldn't this just be the time-filtered endpoint, but for start date = inifinity ago and enddate = yesterday ?

client/src/containers/Roster/Roster.tsx
```diff
@@ -132,8 +131,8 @@ const Roster: React.FC = () => {
       <div className="Roster grid-container">
         <BackButton
           location={
-            query.month || query.withdrawn
-              ? `/roster?organization=${query.organizationId}`
+            (query.month || query.withdrawn) && activeOrgId
```
jlhogan, line 134: i know this isn't actually a diff, but i dont understand this conditional -- go back to roster if we were looking at a filtered roster, otherwise go back to home page?

client/src/containers/Roster/hooks/usePaginatedChildData.tsx
jlhogan: `useSWR` doesn't make the network request if this is null. Previously if we didn't have org id, we didn't want to make the network request. Now, org-id isn't required to make the API request. 

However, for a multi-org user, will this api call be kicked off to grab data for all orgs without the org filter in some state prior to that "activeOrgId" being set? ((^^thing to validate))

client/src/contexts/RosterContext/RosterContext.tsx
```diff
@@ -43,122 +54,41 @@ const defaultCtx = {
...
 
 const RosterContext = React.createContext<RosterContextType>(defaultCtx);
 
 const RosterProvider: React.FC = ({ children }) => {
   const { user } = useContext(UserContext);
-
   const history = useHistory();
   const query: RosterQueryParams = parse(history.location.search);
 
-  // Parse and sanitize query params, update if missing (i.e. initial load) or invalid
```
jlhogan: :wave:

client/src/contexts/RosterContext/RosterContext.tsx
jlhogan: :joy:

client/src/hooks/useAuthenticatedSWR.ts
jlhogan: yessssssssssssssssss

jlhogan (approved):
code review approval -- manual validation happening now!

jlhogan:
manual validation:
- [ ] roster filtered for past month only includes records for children with active enrollment during that month
- [ ] roster filtered for past month displays current enrollment at time of filter

buggy tings:
- roster filter always says "Invalid date"
invalid_date
- upon opening filter modal, immediately presented with error state for empty default filter value: 
date_error
- not really within scope here, and maybe better addressed by the work @nickbristow is currently doing, but confusing that the example date for date picker is in "M D YYYY" format when the date input is only asking for "M YYYY"

client/src/containers/Roster/Roster.tsx
vonbearshark: I _think_ we do want the `Back` button the Roster page in a filtered view to go back to an unfiltered roster, which this effectively does. The `orgId` bit is a little wonky, because the logic as written only does that for a selected orgId, like in multiorg users. So that should probably change regardless

client/src/contexts/RosterContext/RosterContext.tsx
vonbearshark: well based on your investigation I am maybe not so sure...

src/controllers/children/get.ts
vonbearshark: Yeah that makes sense to me!

vonbearshark: Hm actually this might not align with my current thinking on the temporal filter. Right now it captures all enrollments that overlap with the range, as long as they start before the end of the range and either don't end or end after the start of the range. So it gets enrollments that started before and continued through, those that started in the middle and went on, and also those that both started and ended in the range. `-Infinity` to `yesterday` would capture children `withdrawn` today because they meet that last criteria, right? Since they have all their enrollments contained within a range.

jlhogan: ahhh yes, so i have the range wrong then. the full withdrawn roster is start: yesterday, end: yesterday). So that every enrollment going back in time meets first criteria (start before end of range) and all enrollments that are withdrawn as of yesterday meet second criteria (end after start of range). yeah ? time math is hard, but i know this is doable and we defintiely don't need a withdrawn end point.

vonbearshark: Ah yep, that tracks!

src/controllers/children/get.ts
```diff
@@ -1,16 +1,101 @@
...
+}) => {
+  const readOrgIds = await getReadAccessibleOrgIds(user);
+  const orgIds = organizationIds?.length
+    ? intersection(readOrgIds, organizationIds)
+    : readOrgIds;
+
+  const qb = await getManager()
+    .createQueryBuilder(Child, 'Child')
+    // Restrict via org access
+    .where('Child.organizationId IN (:...orgIds)', {
```
jlhogan, line 36: does there need to be a whereNotDeleted for the child entity?

vonbearshark: It seems like the [original](https://github.com/ctoec/data-collection/pull/1253/files#diff-8599112c8a94c9f008727779a95f5e7d9f89aa28f5241c0fcbef61489a6e75d8L21-L32) was only checking for the existence of the child itself, which should happen on the join anyway, right? But I could add it if it's that prop exists on the child potentially

jlhogan: looked into a bit more, createQueryBuilder automatically includes the deleted check for the top-level entity!

src/controllers/children/get.ts
vonbearshark: Hm. Thinking thought now. That yesterday,yesterday gets everything. And that does make sense, I think. As you said, all enrollments meet the first criteria. They also meet the second because they have an active enrollment, so that is still the first criteria: they started before yesterday and went on through...

vonbearshark: We have made the time filter super broad (any enrollment that overlaps with it at all) and the withdrawn filter super specific (at least one active enrollment), so I am actually less confident we can rm the withdrawn filter

vonbearshark:
re: bugs

> roster filter always says "Invalid date"
> upon opening filter modal, immediately presented with error state for empty default filter value:
I actually can't repro these...

Happy to keep at it but between these and the prompt format work I might say that if they are flaky bugs they can be fast-followed with either @nickbristow 's new date picker work, or the arbitrary time range ticket I was planning on taking right after (and should be short work). probably a lot of overlap in those follow tix..

src/controllers/children/get.ts
jlhogan: So, my understanding of the desire to move to this model of time-based filtering was in part explicitly to serve both of these views and remove the "withdrawn" filter. there is not a product need to give users the ability to select random time slices; providers will likely always just want a month roster. 

But now that i'm actually applying my brain to it, i see that you are right because we need to flip off that "exit is after start OR exit is null"

perhaps a thing to do here would be:
if start and end params provided:
- entry is before end AND (exit is after start OR exit is null)
if only end param provided (withdrawn roster!):
- exit is before end
if only start param provided (is this needed? current roster?)
- entry is after start

jlhogan: discussed IRL -- withdrawn =/= normal time filtered active roster!!!! keep separate route and investigate if withdrawn as a view/route/whatever is actually a need

client/src/containers/Roster/hooks/useIsSubmitted.tsx
vonbearshark: Ah so we are good here. This is only used in the `Roster` view, and it is on a per-org basis 👍🏻

client/src/containers/Roster/hooks/usePaginatedChildData.tsx
vonbearshark: Added the fix in for this. Essentially a way simplified version of the `useEffect` we had previously. We just inject the orgId for multiorg users on first load, but without the `useEffect`. Due to our refactor, that is fine

vonbearshark: When we end up at `/roster` from another page (via `history.push`, or updating the url in browser, refresh, etc.), it is injected on that first load. And now with the refactor that query is only updated from the fns exposed by the ctx. And those fns all preserve the pre-existing query, only updating a piece at a time. Finally, the fn that we have to update the query's orgId requires a new orgId.

So in short there is no way to "unset" the orgId once it is added, and it's added right away.

client/src/containers/Roster/Roster.tsx
vonbearshark: (updated this. think it was actually broken and now fixed)

client/src/contexts/RosterContext/RosterContext.tsx
vonbearshark: psych, yep we don't need this. we just need to update the `query` object on first load like normal, without `useEffect` (it is not really state, anyway)

vonbearshark:
Updating the latest commit (since the changes were relevant) and force pushing. Added context for the multiorg situation in comments.

Re: the weird auth bug: my [fix](https://github.com/ctoec/data-collection/pull/1253/files?file-filters%5B%5D=.tsx#diff-ae2208b31ad9bf2c689e61b320d244271531757d3d400a55ae4dcff806dced28R82) is a little unsatisfying. We were `throw`ing an error rather than just logging it. 

1. I think that is the preferred flow regardless? Correct me if I am wrong. There are other solutions now that I know what's going on here, but I generally don't think we should be `throw`ing like this in the client based on server responses?
2. The actual issue in this case is related to the `handleJWTError` flow ( in [`client/src/utils/handleJWTError`](https://github.com/ctoec/data-collection/blob/base/client/src/utils/handleJWTError.ts)), and it actually seems like a perfectly acceptable course of action that was unnecessarily blocking here with the `throw`. Logging it allows the flow to complete and the client to reauth without any noticeable issues at all. Again, I can see about updating this flow so we can keep `throw`ing, but idt that's what we want

jlhogan (approved):
Can confirm: 
- [x] cyclical logout loop is resolved
- [x] multi-org roster queries are filtered by org, and so only appropriate data is displayed in first roster all sites section

bamader:
General roster and flow functionality tests:
- [] Create record flow can add a completed child
- [] Create record saves progress after breaking at each step without crashing roster (new caching)
- [] Batch edit can make and save all edits for a record
- [] Batch edit can do some edits then refresh/return to roster; changes should be saved, nothing should crash
- [] Delete record gets rid of the record from roster (including after refresh) and updates the number at the top of the page
- [] Edit record: each of the five sections can make and save edits to children; these persist through saving via caching and selecting options does not cause non-responsive behavior from the app
- [] Withdraw record removes the record from the active page and updates numbers at the top of page (persists through refresh)
- [] Can time filter for a record immediately after withdrawing it
- [] Can change sites on roster page and see correct children (active + withdrawn); appropriate params are attached to / updated in query
- [] Can change orgs on roster page and see correct children (active + withdrawn)
- [] Submitted success alert persists on roster after submitting data (including after refresh)
- [] Number of errors claimed at the top of roster page matches real number of validation errors
- [] Can only submit data if an org-level user
- [] Roster returns to normal after clearing the time filter
- [] Preview of an upload is correct and proceeds to the roster with the advertised children

Test cases for specific retrieval / roster bugs we've had in the past:
- [] Adding income determinations doesn't cause render errors, and the new dets are retrieved successfully
- [] The correct income det UI displays for children with and without dets
- [] Not disclosed determinations display correctly
- [] The correct enrollment and funding UIs display for children with no enrollments, children with unfunded enrollments, and children with at least one funded enrollment
- [] Reporting periods are parsed forgivingly and without validation errors when uploading sheets
- [] Funding space times are still mapped equivalently and retrieved properly
- [] Pre and post submit pages show correct sections (sites, funding maps) for users at site/org perm levels

src/controllers/children/get.ts
vonbearshark: Ug so this is _not_ correct as is. This was my suspicion originally that I couldn't articulate: as is this filters down the enrollments that have exits, per child. However, we really want those children where _all_ the enrollments are not active.

You would think it would be straightforward to even just `COUNT` the enrollments here or something, but after an hour of wrestling with `typeorm` I am pretty burnt on trying. Likewise also tried to augment the active fn/query as @jlhogan suggested, but it ran into the same thing, essentially: I could create conditions that filtered to only withdrawn enrollments, but couldn't return only the children with _only_ withdrawn enrollments.

Open to thoughts here

vonbearshark:
New commit addressing the bugs @bamader  found. Retriggering tests and deploys and should be ready for revalidation!

Fixed:
- Batch edit banner issues (as well as unrelated bug found in batch edit process)
-  Can only submit data if there's no errors failure
- "Redetermine" bug
- Alert persists on the page after refresh and going through Edit Record work
- Preview -> Roster slow/wonkiness

Couldn't repro:
- Saving a child after inputting just section 1 in create record, then navigating back to roster, then to the create flow crashes the app
-  Number of errors claimed at the top of roster page matches real number of validation errors

Questions:
-  can navigate between pages without hanging or flashing UI from the app
    - This was unchecked (and its parent), is there an issue here you observed @bamader?

Notes:
- Some of the repro issues were tested after some fixes, so it is possible I just fixed them. Also possible I don't understand the context, though.
    - I fixed a pretty gnarly race condition that effected basically the whole Edit experience (and maybe more?) that could easily cloud my repro and affect related behavior
- Also realized the Error banner was wrong (in addition to not showing), so fixed that as well
- I fixed the Preview -> Roster loading lag with `revalidate`, which is [apparently](https://github.com/vercel/swr/discussions/812#discussioncomment-203624) being deprecated. You are supposed to be able to replicate the same functionality with `mutate`, but I haven't had any luck. Left a `TODO`.
- Unrelated, but after poking found that the [withdrawn filter](https://github.com/ctoec/data-collection/pull/1253#discussion_r610339924) is not working as intended 🙄 ...

bamader:
Ooooh lots of new things for me to test 👀 . Will start validating again in an hour or so after I clean up some other stuff. After I check that, I'll see if I can brainstorm any fixes for the Withdrawn filter.

My first thought is a weird hybrid combination of the current work and the old functionality of processing in post. Right now, the code finds all enrollments that have an exit, then gets the children associated with those enrollments. This should, by definition be a set that fully contains children for whom _all_ enrollments are inactive. So what we can do is use a modified version of the current query _without_ any of the joins on nested relations (no families, income dets, etc.), which will obtain a list of child IDs corresponding to children that have at least one inactive enrollment. From there, we make  second query that retrieves _all_ information about each of these children, including the nested relations (we can use the SQL join syntax for this). Then, we do a post-processing `childrenWithAtLeastOneExitedEnrollment.filter( (c) => c.enrollments.all( (e) => e.exit ) )`, which would leave us the list of children for whom all enrollments are inactive. While this may be somewhat inelegant because it uses two queries to accomplish what we _should_ be able to write as one query (curse you typeorm), I think there are two reasons this approach might be okay:

1. JJ has already spent an hour+ trying to get this to work and I'm sure his sanity is being eroded by typeorm idiosyncracies 😕 
2. The first thing we query is not only a containing set for children with all inactive enrollments, but is also still _much_ smaller than a list of all children overall, so maybe the drop in performance won't be a super big deal if we have to make a second query.

Just my two cents on getting this put out sooner so JJ can take some well-deserved time on non query things! I'll get working more on validating this soon.

vonbearshark:
Yeah I generally agree that this is not blocking because at any point I could post-process my way out of this. If anything more sophisticated comes to anyone, please share. In the meantime I will be trying to make it work the Right™️ way. But whenever validation is successful I know I have a non optimal fix that can be validated quickly in isolation.

Also all e2e tests are passing 🎉 , and the only integration test failure is due to `withdrawn` count being off.

src/controllers/children/get.ts
vonbearshark: Also capturing an irl potential to just remove the withdrawn filter from @jlhogan : that would be trickier than it seems. As is the active and withdrawn filters are compliments. Without the withdrawn filter we would have no way of accessing those children. And they currently contribute to counts throughout the app. So I do think it is at least most straightforward to keep the filter itself for now

JordanGuinn: Definitely don't have any good suggestions with query builder, but could we potentially use `NOT EXISTS` here?  Like only select children where there aren't enrollment rows without an exit date.  Forgive my pseudo-SQL, but something like:

```
SELECT *  
FROM Child as c  
WHERE NOT EXISTS  
(SELECT *   
    FROM Enrollment AS e
    WHERE e.childId = c.id 
    AND e.exit IS NULL) ;
```

client/src/containers/Home/PostSubmit.tsx
```diff
@@ -34,31 +34,22 @@ export const PostSubmitHome: React.FC = () => {
...
+        // user-accessible sites into individual cards
+        setSiteCountDisplay(siteCountMap);
 
-    // Get the site count data structure, too, so we can divide up
-    // user-accessible sites into individual cards
-    apiGet('children?siteMap=true', accessToken)
-      .then((res) => setSiteCountDisplay(res.siteCountMap))
+        // Also determine the funding spaces map for the organization, if
+        // the user has the permissions that enable this
+        if (fundingSpacesMap) {
```
JordanGuinn, line 46: nitty nit: should this conditional maybe be `(showFundings && fundingSpacesMap)`?  Just to confirm that we did in fact ask for it, in addition to it being provided.

JordanGuinn (approved):
Second review pass complete - still looks good to me 👍🏽👍🏽👍🏽

src/controllers/children/get.ts
vonbearshark: It does seem like any SQL-y solution would require a separate subqery on `Enrollment`s, which is not ideal (or theoretically necessary), and makes it a little tough to reuse with the current querybuilder builder. I was trying to avoid it, but I do think I could make it work like that, and I think it's preferable to post processing in memory.

JordanGuinn: Agreed on both.  If there _is_ a SQL solution here that doesn't require another subquery on `Enrollment`, I'm not seeing it.

client/src/containers/Home/PostSubmit.tsx
vonbearshark: yeah true

bamader:
Posting a fresh testing suite here so I can work off a clean slate:

General roster and flow functionality tests:
- [x] Create record flow can add a completed child
- [ ] Create record saves progress after breaking at each step without crashing roster (new caching)
  - Saving a child after inputting just section 1 in create record, then navigating back to roster, then to the create flow crashes the app
- [ ] Batch edit can make and save all edits for a record
  - unable to test because the missing info banner disappeared
- [ ] Batch edit can do some edits then refresh/return to roster; changes should be saved, nothing should crash
  - unable to test because the missing info banner disappeared
- [x] Delete record gets rid of the record from roster (including after refresh) and updates the number at the top of the page
- [ ] Edit record: each of the five sections can make and save edits to children; these persist through saving via caching and selecting options does not cause non-responsive behavior from the app
  - [x] edits correctly save
  - [ ] can navigate between pages without hanging or flashing UI from the app
- [x] Withdraw record removes the record from the active page and updates numbers at the top of page (persists through refresh)
- [x] Can time filter for a record immediately after withdrawing it
- [x] Can change sites on roster page and see correct children (active + withdrawn); appropriate params are attached to / updated in query
- [x] Can change orgs on roster page and see correct children (active + withdrawn)
- [x] Submitted success alert persists on roster after submitting data (including after refresh)
- [ ] Upload success alert shows on roster and disappears after other actions are performed
  - Alert persists on the page after refresh and going through Edit Record work
- [ ] Number of errors claimed at the top of roster page matches real number of validation errors
  - this isn't working for me; I can make the missing info banner alert at the top of the page disappear through a combination of uploading, refreshing, going to edit record and making some changes, saving, then refreshing again
- [x] Can only submit data if an org-level user
- [ ] Can only submit data if there's no errors
  - this flat out doesn't work for me; I submitted with tons of errors and it let it go through
- [x] Roster returns to normal after clearing the time filter
- [ ] Preview of an upload is correct and proceeds to the roster with the advertised children
  - Preview is correct; after proceeding to upload, changes are not reflected in roster until refresh is done; even changing pages and then going back to roster doesn't fix this)
  - Also, if you try to upload a second file after uploading a first without refreshing in-between, the roster hangs and gets stuck in the loading state (I saw sticks of 1, 3, and 5+ minutes)

Test cases for specific retrieval / roster bugs we've had in the past:
- [ ] Adding income determinations doesn't cause render errors, and the new dets are retrieved successfully
  - Pressing "Redetermine" to add a new det overwrites the current income det on file; the new information is saved, the old det disappears 
- [x] The correct income det UI displays for children with and without dets
- [x] Not disclosed determinations display correctly
- [x] The correct enrollment and funding UIs display for children with no enrollments, children with unfunded enrollments, and children with at least one funded enrollment
- [x] Reporting periods are parsed forgivingly and without validation errors when uploading sheets
- [x] Funding space times are still mapped equivalently and retrieved properly
- [x] Pre and post submit pages show correct sections (sites, funding maps) for users at site/org perm levels

src/controllers/children/get.ts
vonbearshark: Did something of this effect in last commit. Feel free to take a gander. Can confirm it seems to work

src/controllers/children/get.ts
```diff
@@ -19,79 +104,103 @@ import { getCurrentEnrollment } from '../../utils/getCurrentEnrollment';
...
+  // Subquery: find all active enrollments,
+  qb.andWhere((qb) => {
+    const subQuery = qb
+      .subQuery()
+      .select('e.childId')
+      .from(Enrollment, 'e')
+      .where('e.childId = Child.id AND e.exit IS NULL')
+      .getQuery();
+    return 'NOT Child.id IN ' + subQuery;
+  });
```
vonbearshark, line 178: Worth a 👀

bamader: Today I learned that subqueries are a thing 🤯 . This definitely looks right to me. Will incorporate into my validations.

src/controllers/children/get.ts
```diff
@@ -1,16 +1,101 @@
...
+
+const whereNotDeleted = (field: string) => `${field}.deletedDate IS NULL`;
+
+/**
+ * Get base query for getting children for given user. Applies organizationId filter to
+ * all organizations the user has read-access for.
+ * If user has site-level access, also applies a site filter to only return children
+ * and enrollments for those sites. It also sorts enrollment and financial information.
+ */
+const queryBuilderBuilder = async ({
```
JordanGuinn, line 21: +100000 for the accurately named `queryBuilderBuilder` 😎

src/controllers/children/get.ts
JordanGuinn: Confirmed - looks correct to me too

src/controllers/children/get.ts
vonbearshark: 👷👷

JordanGuinn:
## Final Test Pass

### Upload
- [ ] Confirm I can upload the complete dev file successfully
- [ ] Confirm I can save said file to my roster successfully
- [ ] Confirm I can upload the dev file with partially missing fields successfully
- [ ] Confirm I can save said file to my roster successfully

## Roster
- [ ] Confirm the following within the normal roster view
   - [ ] Children without any enrollments are included
   - [ ] Children with an enrollment without an exit date are included
- [ ] Confirm the following within the month filtered roster view
   - [ ] Children with enrollments that ended prior to/during the filtered month are included
   - [ ] **No** children with enrollments without an exit date are included
- [ ] Confirm the following within the Withdrawn roster view
   - [ ] Children without any enrollments with an exit date are included
   - [ ] No children with an enrollment without an exit date are included
   - [ ] The sites tabs properly filters children by site

Create/Update
- [ ] Confirm I can create a new child without issue
- [ ] Confirm I can save the newly created child to my roster
- [ ] Confirm I can see said child within my roster
- [ ] Confirm I can batch edit children
- [ ] Confirm I can save the batch edits without issue
- [ ] Confirm I can see the edits reflected in my roster

Issue #1237, opened 2021-03-11
Make birth certificate parsing more forgiving


jlhogan:
## Background
Right now, birth certificate fields (`id`, `state` & `town`) are **only** required if the `birthCertificateType` is `US`.  But this is not totally clear to users, who see `OPTIONAL` next to `Birth Certificate Type` in the upload template and then get confused when they get errors.

## Acceptance Criteria
- If **none** of the `id`, `town` or `state` fields are provided for a birth certificate of type `US`, force the birth certificate type to be `Unavailable`
- If `id` is provided, but not `town` or `state` for birth certificates of type `US`, keep the `id` value and use default `Unknown` for the `town` and `state` fields

PR #1254, opened 2021-03-24
1237 make birth certificate parsing more forgiving

nickbristow:
## Background
Updated birth certificate parsing to be more forgiving. 

If US type is selected, and no data (id, state, town) is provided, set birth certificate type to `Unavailable`.

If US type is selected, and ID is provided, and state or town is not provided, Set state and town to "null" aka "not provided"

If US type is selected and State or Town info is provided, but ID is not, throw validation error.

This is done to eliminate confusing errors for the end user. 

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1237

## Validation Plan
Manual test with file uploads with 
US, ID, No state, No County
US, No ID, No state, No County
US, No ID, State, County
Unavailable certificate type

## Automated Testing

## Additional Context


jlhogan:
@nickbristow is this ready for review? I see that it's an active PR, but doesn't have any reviewers. If not, not worries, just didn't want to leave it languishing :)

nickbristow:
> @nickbristow is this ready for review? I see that it's an active PR, but doesn't have any reviewers. If not, not worries, just didn't want to leave it languishing :)

Was having trouble fixing the TSC errors. I think we got it. Waiting for checks to run.

vonbearshark (approved):
lgtm but will defer to others on the conditions meeting the ac. Seems like there's some minor (non-blocking) style differences that it seems weird weren't caught by prettier/pre commit hooks, like spacing after `//` or between `) {`. but 🤷🏻‍♀️

src/entity/Child.ts
```diff
@@ -76,20 +77,50 @@ export class Child implements ChildInterface {
...
+      const stateEmpty = !this.birthState;
+      const idEmpty = !this.birthCertificateId;
+
+      if(this.birthCertificateType === BirthCertificateType.US){
+        //If birthCertificateType is set to US, but no info provided, change to Unavailable
+        if(townEmpty && stateEmpty && idEmpty){
+          this.birthCertificateType = BirthCertificateType.Unavailable;
+          this.birthTown = '';
+          this.birthState = '';
+        }else{
```
JordanGuinn, line 110: How _did_ you manage to commit these changes formatted like this?  Is our prettier commit hook not running for you?

client/src/shared/models/db/Child.ts
```diff
@@ -41,4 +41,5 @@ export interface Child extends ObjectWithValidationErrors {
   family?: Family;
   organization: Organization;
   enrollments?: Array<Enrollment>;
+  updateBirthCertificate?(): void;
```
JordanGuinn, line 44: nit: does this _need_ to exist as an attribute on the model?  Only asking cuz it's shared between client and server

src/entity/Child.ts
```diff
@@ -76,20 +77,50 @@ export class Child implements ChildInterface {
...
 
   @Column({ nullable: true })
-  @ValidateIf((o) => o.birthCertificateType === BirthCertificateType.US)
+  @ValidateIf((o) => {
+    //Only validate if Type is US and birthTown or birthState have data
+    //Will be set to null otherwise
+    if(o.birthCertificateType !== BirthCertificateType.US){
+      return false;
+    }
+    return o.birthTown || o.birthState
```
JordanGuinn, line 92: I don't think I quite understand the need for this line?

src/entity/Child.ts
```diff
@@ -76,20 +77,50 @@ export class Child implements ChildInterface {
...
+  })
   @IsNotEmpty()
   birthCertificateId?: string;
 
+  @BeforeInsert()
+  @BeforeUpdate()
```
JordanGuinn, line 98: Kinda really like these listener annotations, not gonna lie 😎

JordanGuinn (commented):
The meat of the functionality here looks good overall, just got a few minor questions

client/src/shared/models/db/Child.ts
bamader: For listeners, I don't actually think this does need to exist on the model. I think we have other entities that use a similar pattern but only define it on the entity.

src/entity/Child.ts
```diff
@@ -76,20 +77,50 @@ export class Child implements ChildInterface {
...
   @Column({ nullable: true })
-  @ValidateIf((o) => o.birthCertificateType === BirthCertificateType.US)
-  @ChildBirthCertificateSpecified()
   birthState?: string;
 
   @Column({ nullable: true })
-  @ValidateIf((o) => o.birthCertificateType === BirthCertificateType.US)
+  @ValidateIf((o) => {
+    //Only validate if Type is US and birthTown or birthState have data
+    //Will be set to null otherwise
```
bamader, line 88: How did you get this weird spacing to commit? 😂

src/entity/Child.ts
bamader: Yeah I think this line is actually against the AC. The second bullet of the AC is that if `id` is provided, then we keep it even if neither `town` nor `state` is provided. This seems to be validating `id` only if at least one of those fields is given.

src/entity/Child.ts
bamader: Yeah when they actually work they're really handy. I remember Melanie and I spending hours and hours trying to get them to work right for cascade deleting of linked entities. Might just be one of those situations where they work best if you have a narrow problem or specific scope on a single entity (such as this one).

bamader (approved):
LGTM. One question for @JordanGuinn : do we need to make a new migration since we're modifying this entity? My gut says no because we're not actually changing fields in the table, but just worth checking on.

client/src/shared/models/db/Child.ts
vonbearshark: Ah, yep. Whoops. I helped @nickbristow with type issues, and it looks like this change here is just not necessary

src/entity/Child.ts
nickbristow: @bamader that is what's happening.

From AC
```
If none of the id, town or state fields are provided for a birth certificate of type US, force the birth certificate type to be Unavailable
If id is provided, but not town or state for birth certificates of type US, keep the id value and use default Unknown for the town and state fields
```
So the only time we validate that ID is provided is when they give something for the state or town. 

That's how I interpreted the AC anyway.

src/entity/Child.ts
nickbristow: it...is not

src/entity/Child.ts
vonbearshark: Weird that the prettifying update _still_ didn't catch this. huh..

src/entity/Child.ts
nickbristow: Example:
ID: ''
State: ''
Town: ''
//Set to type Unavailable.

ID: '123'
State: ''
Town: ''
// Set state and town to null

ID: ''
State: 'va'
Town: ''
//Throw validation error, ID not provided

bamader: Mmmmm I see. Those examples make perfect sense to me, particularly that last one. I must've mistaken validating for keeping, but I think the behavior you've outlined here is entirely reasonable.

src/entity/Child.ts
bamader: Are you committing while the docker container is up and running? That's what sets off the hook.

nickbristow: @bamader i am.

bamader: ...I got nothin' 🤷‍♂️

JordanGuinn:
@bamader Database isn't being updated at all, so no migration needed.

src/entity/Child.ts
JordanGuinn: Ahhhh, okay that makes sense.  The alternative would be having **zero** situations where we validate `birthCertificateId`, in which case we'd just remove the `isNotEmpty` validator entirely and make this field fully optional, which I _don't believe_ we want to do at this point.

JordanGuinn (approved):

JordanGuinn:
Confirmed the following:

- [ ] Enrollment row with `US` birth cert type, a valid birth cert ID, and no state/country saves the first two fields as they are and saves town/country as `Unknown/Not collected`
- [ ] Enrollment row with `US` birth cert type but no id/town/country saves the birth cert type as `Unavailable` and displays nothing else
- [ ] Enrollment row with `US` birth cert type, a valid town, a valid country and no ID saves all specified fields appropriately and generates a validation error for ID
- [ ] Enrollment row with no birth cert type, but a valid id/town/country saves the birth cert type as `Unavailable` and displays nothing else (though all fields are still written to row)

I'd say this is ready to merge.

PR #1259, opened 2021-03-25
Fixed misc test types

vonbearshark:
## Background
Hopefully straightforward quixfix for tests' types.

I fixed an issue with my VSCode where I now check types project-wide on save. Errors related to these changes came up. At first I dismissed them figuring my project-wide config was broken. It kind of is in that it checks these types while our `tsc` doesn't. But the types _are_ wrong. So updating them here.

### DIFF
--- a/client/src/components/Forms/ChildInfo/ChildInfo.test.tsx
+++ b/client/src/components/Forms/ChildInfo/ChildInfo.test.tsx
@@ -4,13 +4,14 @@ import {
   accessibilityTestHelper,
 } from '../../../testHelpers';
 import { ChildInfoForm } from './Form';
-import { Family } from '../../../shared/models';
+import { Family, Organization } from '../../../shared/models';
 
 const child = {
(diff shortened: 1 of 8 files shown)


JordanGuinn (approved):
Wow, nice catch 👏🏽 So, then I guess `react-scripts test` just straight up doesn't care about invalid Typescript in `.tsx` files?  Like _how have these tests been passing_

bamader (approved):
😳 I know the tests _can_ fail (because I've done stuff that's made them fail before 😝 ), but I also can't explain how these things weren't important to the tests passing. Maybe they were somehow acquired like a virtual property? Maybe? Maybe I shouldn't read too much into it...

vonbearshark:
Yeah cards on the table I don't totally understand how it happened. Maybe these fixes aren't necessary given some kind of ts config in the pipeline or something? But they seemed reasonable enough anyway and it is a quality of life improvement for me, so 🤷🏻‍♀️

Issue #1195, opened 2021-02-23
Active enrollment count on post-submit home page does not match corresponding count in roster


jlhogan:
## Background
On the "post-submit" homepage, we include a count of "active enrollments" at the top of the page. This number does _NOT_ match the actual number of enrollments in the active roster in at least one instance (orgid = 118), where the reported active enrollments is 104, but the roster contains 69 students.

## Acceptance Criteria
- Active enrollment count on the post-submit home page matches the corresponding count in the roster

PR #1274, opened 2021-04-12
Fixed postsub count

vonbearshark:
Now on active children, previously on all children

## Background
Used the existing `getActiveChildren` fn and abstracted the query so we can now get the `count` and `children`, which should then be identical 

## GitHub Issue
#1195 

## Associated PRs
N / A

## Validation Plan
Seems like the issue is around withdrawn children, so:
- [x] Withdrawn a child from the roster and check the count is updated appropriately 

## Automated Testing

### DIFF
--- a/src/controllers/children/get.ts
+++ b/src/controllers/children/get.ts
@@ -10,6 +10,11 @@ import { NestedFundingSpaces } from '../../../client/src/shared/payloads/NestedF
 import { getFundingSpaces } from '../../controllers/fundingSpaces';
 import { intersection } from 'lodash';
 
+type skipTake = {
+  skip?: number;
+  take?: number;
+};
+
 const whereNotDeleted = (field: string) => `${field}.deletedDate IS NULL`;
 
 /**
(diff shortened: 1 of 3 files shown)


src/controllers/children/get.ts
```diff
@@ -152,17 +147,43 @@ export const getActiveChildren = async (
   if (skip) qb.skip(skip);
   if (take) qb.take(take);
 
+  return qb;
+};
+
+export const getActiveChildren = async (
+  user: User,
+  organizationIds?: string[],
+  selectParms?: {
```
JordanGuinn, line 156: nit: presumably this should say `selectParams`

src/controllers/children/get.ts
```diff
@@ -152,17 +147,43 @@ export const getActiveChildren = async (
...
+};
+
 export const getWithdrawnChildren = async (
   user: User,
   organizationIds?: string[],
-  selectParms?: {
-    skip?: number;
-    take?: number;
-  }
+  selectParms?: skipTake
```
JordanGuinn, line 186: same typo nit

JordanGuinn (approved):
LGTM 👍🏽👍🏽

src/controllers/children/get.ts
bamader: unless you have somehow found a way to add delicious parmesan to our codebase... 🧀

bamader (approved):
LGTM! I'll move to manual validation on this.

bamader:
Confirmed via validation: number updates perfectly after a withdraw

src/controllers/children/get.ts
vonbearshark: 🤌🤌

vonbearshark:
Moving to `Validation`, deploying to `qa`, and kicking off e2e tests

jlhogan (approved):

Issue #1264, opened 2021-03-31
Refactor and simplify mapRows logic


JordanGuinn:
## Background
The logic currently responsible for translating **enrollment report data** extracted from file uploads into **child data** written to row is massively complex and also largely prone to error.  We should take some time to simplify it and add any necessary automated testing to be confident in what gets built.  

Additionally, we should decouple the behavior of transforming enrollment report data into child data from the subsequent behavior to batch write that child data to row.  We often want to do one without the other, so those should just exist independently of each other.

## Acceptance Criteria
- The enrollment to child transformation and write logic is refactored to be simpler and more accessible to future engineers
   - Exists as the `mapRows` function in `src/controllers/enrollmentReports/map/map.ts`
- The enrollment to child data transformation logic is decoupled from the corresponding batch write logic
   - That is to say, they exist independently of each other and can be executed in sequence
- Unit/e2e tests are written capturing the largest functional pieces here
- All works as it did previously

PR #1285, opened 2021-04-19
1264 maprows refactor

jlhogan:
## Background
Refactors the mapping functionality to:
- consolidate matching logic to make conditional flows less complex
- more cleanly separate mapping from writing to DB
- remove complex guessing functionality: require users to explicitly end all fundings/enrollments that will be updated, following the instructions described here: https://help.ece-reporter.ctoec.org/update/
- when looking for DB match, add organizationId to where clause to ensure users don't update other folk's data if they have enrolled the same individual (this kind of matching can be layered in later to alert people to potential overutilization; deduping for analytics will happen in post-processing).

## GitHub Issue
#1264 

## Associated PRs
N/A

## Validation Plan
see https://github.com/ctoec/data-collection/pull/1182

## Automated Testing

## Additional Context
Need to finish up commenting and bringing back all the tests, however I have less capacity for ths work than I thought so maybe we should discuss me handing this work off to someone else to own getting over the finish line. Sorry!


src/controllers/enrollmentReports/map/mapRows/updateRecord/updateBirthCertifiate.ts
```diff
@@ -0,0 +1,31 @@
...
+import { MapThingHolder } from '../../setUpMapThingHolder';
+
+export const updateBirthCertificate = (
+  row: EnrollmentReportRow,
+  match: Child,
+  thingHolder: MapThingHolder
+) => {
+  if (
+    !row.birthCertificateType ||
+    row.birthCertificateType === match.birthCertificateType ||
```
bamader, line 17: Won't this not update `birthTown` and `birthState` for a record whose birthCertType is already US but has those marked as `Unknown/Not Disclosed`? We may to add an `&& row.birthCertType != BirthCertType.US` to this so that we can still pull new info.

src/controllers/enrollmentReports/map/mapRows/updateRecord/updateEnrollmentFunding.ts
```diff
@@ -0,0 +1,134 @@
...
+
+const getFundingMatch = (
+  fundingFromRow: Funding,
+  fundings: Funding[] | undefined
+) =>
+  fundings?.find(
+    (f) =>
+      // Funding spaces match
+      f.fundingSpace &&
+      fundingFromRow.fundingSpace?.id === f.fundingSpace.id &&
```
bamader, line 107: More of a check on my understanding, but the fundingFromRow can never have a funding space whose ID _doesn't_ match if it's supposed to be an existing funding, right? Because the mapFunding function assigns an ID based on the organization, and so for example any full-time CDC funding for Organization X will always have the same ID?

bamader: Phrased differently, checking the _id_ of the fundingSpace is better than checking the _text_ of the fundingSpace itself, correct?

src/controllers/enrollmentReports/map/mapRows/updateRecord/updateIncomeDetermination.ts
```diff
@@ -0,0 +1,51 @@
...
+import { mapIncomeDetermination } from '../../..';
+import { MapThingHolder } from '../../setUpMapThingHolder';
+import { getLastIncomeDetermination } from '../../../../../utils/getLastIncomeDetermination';
+
+export const updateIncomeDetermination = (
+  row: EnrollmentReportRow,
+  match: Child,
+  thingHolder: MapThingHolder
+) => {
+  const currentDetermination = getLastIncomeDetermination(match.family);
```
bamader, line 13: Potentially semantics, but I know for enrollments there's a difference between `getCurrentEnrollment` and `getLastEnrollment`. Is that a difference we need to be aware of / enforce here, as well?

src/controllers/enrollmentReports/map/mapRows/updateRecord/updateIncomeDetermination.ts
```diff
@@ -0,0 +1,51 @@
...
+  const newDetermination = mapIncomeDetermination(row, match.family);
+
+  if (rowHasNewDetermination(newDetermination, currentDetermination)) {
+    match.family.incomeDeterminations.push(newDetermination);
+    match.tags.push(ChangeTag.IncomeDet);
+  }
```
bamader, line 19: Since there's no `else`, this function doesn't let a user update information in an existing income determination (such as add missing fields or change from `not disclosed` to providing info). Is that desired behavior?

src/controllers/enrollmentReports/map/mapRows/updateRecord/updateRecord.ts
```diff
@@ -0,0 +1,22 @@
...
+) => {
+  // birth cert
+  updateBirthCertificate(row, match, thingHolder);
+  // family address
+  updateFamilyAddress(row, match, thingHolder);
+  // income determination
+  updateIncomeDetermination(row, match, thingHolder);
+  // enrollment
+  updateEnrollmentFunding(row, match, thingHolder);
+};
```
bamader, line 22: I'm on day 2 of this review so this could definitely just be me missing / forgetting, but where do updated entities get re-saved to the DB? These updateEntity functions only make changes to the object and then push a change tag onto the MapThingHolder, but I don't see where that added information is being persisted?

src/controllers/enrollmentReports/map/mapRows/utils/child.ts
```diff
@@ -101,49 +101,6 @@ export const mapChild = (
...
-    }
-    if (!child.birthState) {
-      child.birthState = source.birthState;
-      madeAChange = true;
-    }
-    if (!child.birthTown) {
-      child.birthTown = source.birthTown;
-      madeAChange = true;
-    }
-  }
```
bamader, line 141: Looks like we did used to allow updating missing or unknown fields for US birth certificates

src/controllers/enrollmentReports/map/mapRows/utils/funding.ts
```diff
@@ -0,0 +1,114 @@
...
+import { reportingPeriodIncludesDate } from '../../../../../utils/reportingPeriodIncludesDate';
+
+/**
+ * Create a Funding object from FlattenedEnrollment source,
+ * associated with a FundingSpace for given organization and ageGroup
+ * (already parsed from source, stored as enrollment.ageGroup)
+ * @param transaction
+ * @param source
+ * @param organization
+ * @param enrollment
```
bamader, line 25: Nit: doc params don't match the method signature anymore

src/controllers/enrollmentReports/map/setUpMapThingHolder.ts
```diff
@@ -0,0 +1,51 @@
...
+export type MapThingHolder = {
+  transaction: EntityManager;
+  user: User;
+  organizations: Organization[];
+  sites: Site[];
+  fundingSpaces: FundingSpace[];
+  reportingPeriods: ReportingPeriod[];
+
+  mappedChildren: Child[];
+};
```
bamader, line 21: Nit: can we change the name of this type to be more informative? Something like `mapSummary` or `mapResult`? 'Thing' makes it hard to understand what this type is actually holding in other parts of the code.

src/index.ts
```diff
@@ -1,4 +1,5 @@
 import express, { json } from 'express';
+
```
bamader, line 2: Ah the joys of forcing the code to re-compile and re-start itself 🙃

bamader (commented):
Looks solid overall. I think this is a pretty huge improvement in modularity, readability, and small abstracted code chunks. I particularly like the unification changes that were made to the front end and UI components. The only general comment I have that isn't really captured in specific code comments is that I find the new directory structure for map somewhat confusing. The names of containing folders (imo) aren't necessarily reflective of the kinds of code the folders hold, and there was definitely a period where I had to open VSCode and just stare at the file hierarchy mentally cataloguing the lay of the land. I would be in favor of moving some of the files around in a different way that to me is more consistent with what we do elsewhere in the code and a little more apt for their names. Something like:

```
/map
    createRecord.ts
    mapRows.ts
    batchSave.ts
    /entities
        [this is just a rename of the utils folder because that code seems more about entity management, and utils seems better used for other code]
    /updates
        [everything you've currently got in there]
    /utils
        getChildMatch.ts
        lookupOrganization.ts
        lookupSite.ts
        setupMapThingHolder.ts
```
This organization (or something like it) I think would be helpful when trying to find where a file or function that map uses lives. I.e. it wasn't obvious to me to look in `utils` for the `mapEnrollment` function. Not a big deal if others think the given organization is okay, just thought I'd take the opportunity to make the suggestion since it is a refactor. Otherwise, looks good!

client/src/containers/Upload/ReviewMissingInfo.tsx
```diff
@@ -1,109 +1,58 @@
...
           <div className="display-flex flex-row">
-            <InlineIcon className="missing-info-icon" icon="complete" />
+            <InlineIcon className="missing-info-icon" icon="incomplete" />
             <p className="margin-top-0 margin-left-1">
-              Your file has no missing information required for OEC reporting!
-              Go to the next step to review changes and upload to your roster.
+              The uploaded file had missing information OEC requires.
+              <br />
+              To make updates, go back to choose a resolved file or continue
+              with this upload and make edits in the roster.
```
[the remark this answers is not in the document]
jlhogan: not dumb, plain language matters, but i think this belongs as a convo somewhere else; let's bring this to jay/derek/emily to see where they'd want suggestions for revision of content to live

client/src/shared/models/ChangeTags.ts
```diff
@@ -6,4 +6,5 @@ export enum ChangeTag {
   WithdrawnRecord = 'Withdrawn Record',
   IncomeDet = 'New Income Determination',
   Edited = 'Edited Information',
+  Edit = 'Edit',
```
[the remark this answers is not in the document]
jlhogan: i don't remember making this change WEIRD! no no need.

src/controllers/enrollmentReports/map/mapRows/updateRecord/updateEnrollmentFunding.ts
jlhogan: if the funding space Ids are not the same, it's not the same funding OR our funding space mapping logic is wrong. In my opinion using identity checks when legit is more understandable, and it's less actual work so felt like the right move to me

src/controllers/enrollmentReports/map/mapRows/updateRecord/updateRecord.ts
jlhogan: b/c in javacsript objects are passed by reference, when we update this "match" we update the match that's in the `mappedChildren` cache. So when we then go on to save those entities, they have all the new bits n bobs

src/controllers/enrollmentReports/map/setUpMapThingHolder.ts
jlhogan: yeah this is not at all what i was planning to merge, i just didnt want to waste time thinking about a name. it's not a map result or a summary tho, it's an object that holds a bunch of reference info that is passed around to avoid the use of globals and avoid really long cluttered function signatures.

src/index.ts
jlhogan: oops

src/controllers/enrollmentReports/map/mapRows/updateRecord/updateRecord.ts
bamader: Ahhh I see where I goofed. I didn't notice that we were calling the `batchSave` on the mapped children _within_ `thingHolder`, I just thought we were batch saving on the new records and not the 'cache' so to speak. I blame morning brain 🧠

jlhogan: yup, anything in the cache in thing holder was in the spreadsheet, so we just upsert all of those entities

client/src/shared/models/db/Child.ts
```diff
@@ -41,4 +41,6 @@ export interface Child extends ObjectWithValidationErrors {
   family?: Family;
   organization: Organization;
   enrollments?: Array<Enrollment>;
+
+  tags?: ChangeTag[];
```
jlhogan, line 45: this was easier than a new composite type `Child & { tags?: ChangeTag[] }` due to orm expectations etc.

src/controllers/enrollmentReports/formatUploadPreview.ts
```diff
@@ -1,38 +1,35 @@
...
   getLastFunding,
 } from '../../../client/src/utils/models';
-import { validate } from 'class-validator';
 
 /**
  * Build the data structure that will map to the table we'll show
  * in the front end that represents a preview of a user's upload.
  * @param mapResult
  */
-export const formatUploadPreview = (mapResult: EnrollmentReportUpdate) => {
```
jlhogan, line 15: todo here: update controller to just send the children entities and the client side to render the preview as a table of children (modified roster, roster + column for change tag)

src/controllers/enrollmentReports/map/setUpMapThingHolder.ts
```diff
@@ -0,0 +1,64 @@
...
+ */
+export type MapThingHolder = {
+  transaction: EntityManager;
+  user: User;
+  organizations: Organization[];
+  sites: Site[];
+  fundingSpaces: FundingSpace[];
+  reportingPeriods: ReportingPeriod[];
+
+  mappedChildren: Child[];
```
jlhogan, line 27: todo: remove child cache from this transaction variables bag and have it initialized within `mapRows` func and then returned by that func (as it is now!)

src/routes/enrollmentReports.ts
```diff
@@ -49,38 +51,37 @@ enrollmentReportsRouter.post(
...
-            validationErrors: await validate(child, {
-              validationError: { target: false, value: false },
-            }),
-          }))
+        const reportChildren = await mapController.mapRows(
+          reportRows,
+          thingHolder
         );
 
         const enrollmentColumnErrors: EnrollmentColumnError[] = await controller.checkErrorsInChildren(
```
jlhogan, line 66: todo: this is gonna go away when we skip the "missing info" step and just use the children entities with validations applied (using the validateObejct function) as content for the preview roster and somehow display thsoe errors in that roster

src/controllers/enrollmentReports/map/mapRows/mapRows.ts
```diff
@@ -0,0 +1,42 @@
...
+  thingHolder: MapThingHolder
+) => {
+  for (const row of rows) {
+    await mapRow(row, thingHolder);
+  }
+
+  return Promise.all(
+    thingHolder.mappedChildren.map(async (child) => ({
+      ...child,
+      validationErrors: await validate(child, {
```
jlhogan, line 21: todo: use validateObject helper func here?

src/controllers/enrollmentReports/map/mapRows/mapRows.ts
```diff
@@ -0,0 +1,42 @@
...
+import { getChildMatch } from './getChildMatch';
+import { createRecord } from './createRecord';
+import { updateRecord } from './updateRecord/updateRecord';
+import { MapThingHolder } from '../setUpMapThingHolder';
+
+export const mapRows = async (
+  rows: EnrollmentReportRow[],
+  thingHolder: MapThingHolder
+) => {
+  for (const row of rows) {
```
jlhogan, line 14: const mappedChildCache = []
pass that into mapRow

src/controllers/enrollmentReports/map/batchUpsertMappingEntities.ts
```diff
@@ -0,0 +1,126 @@
...
+  // Upsert family entities
+  const upsertedFamilies = await doBatchSave(
+    user,
+    transaction,
+    children.map((c) => c.family)
+  );
+
+  // Update familyId references and
+  // unsert new income determination entities
+  // (Editing income determinations is not supported at present)
```
jlhogan, line 64: todo: we don't _need_ to do this filtering for only new income dets, could just upsert them all; nothing different would happen except would do more DB writes. in fact, def get rid of that filter!

src/controllers/enrollmentReports/map/batchUpsertMappingEntities.ts
```diff
@@ -0,0 +1,126 @@
...
+  // Update childId references and
+  // remove funding references and
+  // map fundings into enrollment-indexed array
+  // so they can be upserted with updated enrollment references and
+  // upsert enrollment entities with updated
+  const fundingsByEnrollment: Funding[][] = [];
+  const enrollmentsForUpsert: Enrollment[] = [];
+  children.forEach((child, idx) => {
+    child.enrollments?.map((e) => {
+      const enrollmentForUpsert = transaction.create(Enrollment, e);
```
jlhogan, line 98: this shouldn't be necessary, but was because at some point enrollments were not the right type anymore. maybe can just go.

src/controllers/enrollmentReports/map/batchUpsertMappingEntities.ts
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * SQL Server driver has an upper limit of 2100 parameters per query.
+ * Using object keys in the entity has proxy for number of parameters
+ * that'll end up in the query (couldn't find any better way to do this
+ * from typeORM), chunk the inserts into batches that will have <2000
+ * parameters per batch.
+ * @param transaction
+ * @param entities
+ */
```
vonbearshark, line 13: bummer but makes sense

src/controllers/enrollmentReports/map/batchUpsertMappingEntities.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  // Update familyId references and
+  // unsert new income determination entities
+  // (Editing income determinations is not supported at present)
+  const incomeDetsForUpsert = children.reduce((newIncomeDets, child, idx) => {
+    child.family.incomeDeterminations
+      ?.filter((det) => !det.id)
+      .forEach((det) => {
+        det.familyId = upsertedFamilies[idx].id;
+        det.family = undefined;
```
vonbearshark, line 70: gosh too bad we don't have some kind of library for _mapping_ these _relations_....

vonbearshark: I'll probably add a comment for the `undefined` assignments throughout for this reason, just personal todo

client/src/containers/Upload/ReviewMissingInfo.tsx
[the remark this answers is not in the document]
vonbearshark: I think Bloom folks were looking to take a crack at site copy as well, like Megan

client/src/containers/Upload/UploadWizard.tsx
```diff
@@ -0,0 +1,86 @@
...
+          </span>
+          <span className="usa-step-indicator__total-steps">
+            of {UPLOAD_WIZARD_STEPS.length}
+          </span>
+          {UPLOAD_WIZARD_STEPS[currentStepIndex].description}
+        </h2>
+
+        <StepContent
+          setCurrentStepIndex={setCurrentStepIndex}
+          setAlerts={setAlerts}
```
vonbearshark, line 75: we shouldn't need these props and we can just `useAlerts` in the children components, right?

vonbearshark (commented):
Altogether this looks like a huge improvement! Comments are mostly either my own todos or to open opportunities for discussion on some refactor ideas I am kicking around.

For all: I am relieving @jlhogan and taking it from here. We synced, and I _should_ have all the context I need. Reviews and continued discussion are still very appreciated! (That is what I would be looking for at this point in the code, anyway.) Next steps are incorporating this feedback along with my / @jlhogan 's plans for immediate refactor, fixing tests, then pushing back up for final review/validation (maybe later this week/early next).

src/controllers/enrollmentReports/map/mapRows/updateRecord/updateEnrollmentFunding.ts
bamader: That makes a lot of sense, particularly that first part: different IDs means the fundings _are_ actually different, or _we're_ doing something wrong (and both are informative!). Totally good with this, then.

src/controllers/enrollmentReports/map/map.test.ts.txt
```diff
@@ -0,0 +1,562 @@
+//import { EnrollmentReportRow } from '../../../template';
```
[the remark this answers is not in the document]
bamader: In my mind I'm imagining some tiny kid going up to shoot a basketball and a colossal 8 foot giant of a man just smashing away his hopes and dreams 😂 . That's how little power this has now.

src/routes/enrollmentReports.ts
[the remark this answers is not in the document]
bamader: Yeah there's some difference in what they do structurally. `validateObject` works on a single child and extends the record to include the `validationErrors` property (where errors get stored). `checkErrorsInChildren` is a function that operates on a whole array of children and accumulates the errors that they have across _all_ of them. It uses specific targeted attributes of validation errors (like `target` and `property` and whether there are nested errors) to figure out which entity actually caused the issue, so that we can build a data structure showing which spreadsheet columns in the user's upload actually contained the errors. It's a specific function for our "Preview" and "Missing Information" screens in the upload wizard.

src/routes/enrollmentReports.ts
```diff
@@ -200,3 +181,19 @@ enrollmentReportsRouter.post(
     });
   })
 );
+
+const getUploadCounts = (children: Child[]) => {
+  return children.reduce(
+    (response, child) => {
+      if (child.tags.includes(ChangeTag.NewRecord)) response.newCount += 1;
```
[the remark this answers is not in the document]
bamader: Dude when I was first reviewing this some of the bracketless functions set my head on fire 🤯

src/routes/enrollmentReports.ts
```diff
@@ -137,56 +138,36 @@ const logUploadErrors = (userId: number, childrenWithErrors: Child[]) => {
...
   passAsyncError(async (req, res) => {
-    return getManager().transaction(async (tManager) => {
+    return getManager().transaction(async (transaction) => {
       // Ingest upload by parsing, mapping, and saving uploaded data
       try {
-        const shouldSave = req.params['shouldSave'] === 'true';
         const reportRows = controller.parseUploadedTemplate(req.file);
-        const mapResult = await controller.mapRows(
-          tManager,
+        const thingHolder = await mapController.setUpThingHolder(
```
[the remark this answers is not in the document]
bamader: ➕  for `transactionData`! Although, if it just lived internally in `mapController` and we didn't export it for other update functions to use, wouldn't we not be able to track changes to existing children?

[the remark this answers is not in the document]
bamader: Oh wait duh that's not what you're saying. The `mapController` still calls those other functions. Ignore me.

src/utils/reportingPeriodIncludesDate.ts
```diff
@@ -0,0 +1,7 @@
+import { Moment } from 'moment';
```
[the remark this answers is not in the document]
bamader: As far as I know, that hasn't been deliberate, maybe just an incidental artifact. In this case, because the function is so specific to one kind of thing (reporting periods), I would be totally fine with just putting it directly in the util file that used it.

src/entity/Child.ts
```diff
@@ -213,4 +217,6 @@ export class Child implements ChildInterface {
   deletedDate: Date;
 
   validationErrors?: ValidationError[];
+
+  tags?: ChangeTag[];
```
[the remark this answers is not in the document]
bamader: I would argue no, for the same reason that we don't have a `ChildWithValidationErrors` type. Using the same 'root
 entity so to speak (in this case, a `child`) and then augmenting it with any additional properties that entity will need for processing saves duplicated code and probably some headaches relating to type (i.e. we don't have to specify for some functions that we're using a `ChildWithValidationErrors` or a `ChildWithTags`; we can just use `Child` to cast all functions.). Now, you could definitely counter and say that we probably _should_ be typing things that way so that the tight controls also make the code less brittle, but I think it might be a deeper rabbit hole than it's worth diving down.

src/entity/Child.ts
```diff
@@ -192,6 +193,9 @@ export class Child implements ChildInterface {
   @ManyToOne(() => Family)
   family?: Family;
 
+  @Column()
+  familyId: number;
```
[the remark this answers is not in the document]
bamader: I'm actually not sure why this is necessary as its own field? Because as long as a `family` is retrieved with a `child`, we should just be able to access `child.family?.id` anytime we need this?

src/controllers/enrollmentReports/map/setUpMapThingHolder.ts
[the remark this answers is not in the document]
bamader: Agreed!

src/controllers/enrollmentReports/map/mapRows/mapRows.ts
```diff
@@ -0,0 +1,42 @@
...
+  );
+};
+
+const mapRow = async (
+  row: EnrollmentReportRow,
+  thingHolder: MapThingHolder
+) => {
+  try {
+    const organization = lookUpOrganization(row, thingHolder);
+    const match = await getChildMatch(row, organization, thingHolder);
```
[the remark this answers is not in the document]
bamader: I think we left `getChildMatch` as a separate function to account for cases of there being one or more (historical) records for the same child in one sheet. We use (or at least, at one time used) the child match function to detect both whether there are DB matches for existing children, and whether the record currently being processed is the same as a record we've already seen in this sheet but haven't saved yet. So it might be worth keeping the `getChildMatch` separate, but I agree we can probably do in-mem cache creation and mapping at the same time.

client/src/routes.ts
```diff
@@ -113,17 +112,7 @@ export const routes: RouteConfig[] = [
...
-  },
-  {
-    path: '/missing-info',
-    component: ReviewMissingInfo,
-    unauthorized: false,
-  },
-  {
-    path: '/preview',
-    component: Preview,
+    component: UploadWizard,
```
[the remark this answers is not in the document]
bamader: Yuuuup I think those old routes are now whatevs 🗑️

src/routes/enrollmentReports.ts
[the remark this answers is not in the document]
vonbearshark: > It uses specific targeted attributes of validation errors (like target and property and whether there are nested errors) to figure out which entity actually caused the issue

Hm.

Would it be accurate to say that `checkErrorsInChildren` does a `validateObject` over all the children, then? So it's like the relationship between the two is that `validateObject` is a subcall in `checkErrorsInChildren`, which does more orchestration on top of that?

src/entity/Child.ts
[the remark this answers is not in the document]
vonbearshark: Agreed on both counts. I _would_ argue we should be typing things more deliberately in that way, but I would also argue that in an alternate universe where we can nitpick about these things

src/entity/Child.ts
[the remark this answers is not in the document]
vonbearshark: We have this pattern for other fields, like `organization` and `organizationId`, right? Just not sure why it's necessary, or in this case why it is necessary due to these changes specifically

src/controllers/enrollmentReports/map/batchSave.ts
```diff
@@ -0,0 +1,107 @@
...
+    transaction,
+    children.map((c, idx) => {
+      c.familyId = upsertedFamilies[idx].id;
+      c.family = undefined;
+      return transaction.create(Child, c);
+    })
+  );
+
+  // Upsert enrollment entities with updated child references and removed funding references
+  const fundingsByEnrollment: Funding[][] = [];
```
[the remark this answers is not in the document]
vonbearshark: Is this helpful enough? from earlier in the file https://github.com/ctoec/data-collection/pull/1285#discussion_r617041372. Could still be insufficient. It's a tricky and annoying concept

src/controllers/enrollmentReports/formatUploadPreview.ts
vonbearshark: Right now I have half of this: children entities are sent back and forth, but I just moved this formatting to the client, which seems like a good compromise for this pr, unless you think the rest would be simpler than I can see right now

src/controllers/enrollmentReports/map/mapRows/mapRows.ts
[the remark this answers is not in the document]
vonbearshark: Hm. With the new code we are `findOne`ing, so shouldn't be more than one possible result...

src/controllers/enrollmentReports/map/map.test.ts
```diff
@@ -301,215 +303,77 @@ describe('controllers', () => {
...
-        } as Child;
-        const other = {
-          birthdate: BIRTHDATE,
-          firstName: FIRSTNAME,
-          lastName: 'Other name',
-        } as EnrollmentReportRow;
-
-        expect(isIdentifierMatch(child, other)).not.toBeTruthy();
-      });
-    });
```
vonbearshark, line 436: Removed `isIdentifierMatch`

src/controllers/enrollmentReports/map/map.test.ts
```diff
@@ -523,83 +387,49 @@ describe('controllers', () => {
       const NEW_TOWN = 'new town';
       const NEW_STATE = 'new state';
       const NEW_ZIP = '98765';
-      const HOMELESSNESS = UndefinableBoolean.No;
-      const NEW_HOMELESSNESS = UndefinableBoolean.Yes;
-      let FAMILIES_TO_UPDATE: Family[];
-
-      it('change homelessness status if present', () => {
```
vonbearshark, line 530: Seems like we removed this. Intentional?

src/controllers/enrollmentReports/map/map.test.ts
```diff
@@ -681,27 +509,8 @@ describe('controllers', () => {
             ageGroup: NEW_AGE_GROUP,
             entry: NEW_ENTRY,
           } as Enrollment;
-          const row = { model: NEW_MODEL } as EnrollmentReportRow;
-          const isNewEnrollment = rowHasNewEnrollment(row, other, enrollment);
-          expect(isNewEnrollment).toBeTruthy();
-        });
-
-        it('does nothing if a row has no enrollment info at all', () => {
```
vonbearshark, line 689: This seemed to be legitimately broken with the changes. I don't _think_ this test makes sense anymore, but thoughts?

src/controllers/enrollmentReports/map/map.test.ts
```diff
@@ -716,61 +525,23 @@ describe('controllers', () => {
...
-        });
-        it('finds other exit reasons', () => {
-          const newEnrollment = {
-            site: SITE,
-            model: MODEL,
-            ageGroup: AGE_GROUP,
-            entry: NEW_ENTRY,
-          } as Enrollment;
-          const exitReason = getExitReason(prevEnrollment, newEnrollment);
-          expect(exitReason).toEqual(ExitReason.Unknown);
```
vonbearshark, line 760: Removed `getExitReason`

src/controllers/enrollmentReports/map/map.test.ts
```diff
@@ -787,21 +558,8 @@ describe('controllers', () => {
           fundingSpace: NEW_FUNDING_SPACE,
           firstReportingPeriod: NEW_FIRST_PERIOD,
         } as Funding;
-        const isNewFunding = rowHasNewFunding(other, funding);
-        expect(isNewFunding).toBeTruthy();
-      });
-      it('does nothing if row has no funding info', () => {
```
vonbearshark, line 793: Also, re: this seemed to be legitimately broken with the changes. I don't _think_ this test makes sense anymore, but thoughts?

client/src/shared/models/ChangeTags.ts
```diff
@@ -1,5 +1,4 @@
 export enum ChangeTag {
-  AgedUp = 'Aged Up',
```
vonbearshark, line 2: Unused so remove?

vonbearshark (commented):
This is good for another round of 👁️ .

All tests are passing locally except I haven't tried any e2e tests.

- Reorganized things
- Fixed tests
- Made some operations more functional
- Now pass child records client <-> server

src/controllers/enrollmentReports/map/batchUpsert.ts
```diff
@@ -0,0 +1,129 @@
...
+    children.map((c) => c.family)
+  );
+
+  // Update familyId references and
+  // unsert new income determination entities
+  // (Editing income determinations is not supported at present)
+  const incomeDetsForUpsert = children.reduce((newIncomeDets, child, idx) => {
+    child.family.incomeDeterminations.forEach((det) => {
+      det.familyId = upsertedFamilies[idx].id;
+      det.family = undefined; // For typeorm
```
bamader, line 69: +1 for these notes; I find it really helpful to denote something like this that doesn't make intuitive sense to me as existing because typeorm expects entities a certain way

src/controllers/enrollmentReports/map/createRecord.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Create a new record from an EnrollmentReportRow
+ * for which no matching child was found.
+ *
+ * Creates full object graph (family, income determinations, child, enrollment, funding)
+ * regardless of if the information provided is complete
+ * i.e. will create an empty funding if no funding information is provided,
+ * because every enrollment requires a funding.
+ * If this requirement changes, be sure to update the mapping utils.
+ *
```
bamader, line 23: ➕➕➕➕➕➕➕➕➕➕➕➕ _soooooooo_ many for this! Whether to create empty entities has been a source of back and forth confusion for so long, it's great that we're finally just explicitly documenting "this is the way we're doing things."

src/controllers/enrollmentReports/map/map.test.ts
bamader: Yeah that seems fair. There was a point in time where we were seeing weirdness from that function because the wrong types of objects were secretly being passed in, but stronger typing in the controller removes that concern.

src/controllers/enrollmentReports/map/map.test.ts
bamader: I don't know why this was removed, unless the update family function no longer allows a user to edit homelessness status? But that doesn't seem right because that's something we _should_ allow and is easy to do.

src/controllers/enrollmentReports/map/map.test.ts
bamader: It doesn't make sense anymore, no. Previously, behavior was that if a row contained no information at all, we wouldn't do the blank enrollment (even though at one time before that, we did make an empty entity--see what I mean about loving the standardized declaration?), so we expected a no-op. Now, we _do_ make something every time, so this test has no value. I agree with deletion.

src/controllers/enrollmentReports/map/map.test.ts
bamader: Yup! Since we got rid of the guessing logic this test doesn't matter anymore.

src/controllers/enrollmentReports/map/map.test.ts
bamader: Correct! It makes no sense for exactly the same reason as the `rowHasNewEnrollment` test I commented on above (re: blank entities). Delete this bad boi.

src/controllers/enrollmentReports/map/mapRows.ts
```diff
@@ -0,0 +1,120 @@
...
+ * - the cache of all mapped children
+ */
+export type TransactionMetadata = {
+  transaction: EntityManager;
+  user: User;
+  organizations: Organization[];
+  sites: Site[];
+  fundingSpaces: FundingSpace[];
+  reportingPeriods: ReportingPeriod[];
+};
```
bamader, line 35: ➕ for _great_ name and structure

src/controllers/enrollmentReports/map/mapRows.ts
```diff
@@ -0,0 +1,120 @@
...
+        row,
+        organization,
+        transactionMetadata
+      );
+      if (dbMatch) {
+        // DB match: update existing record, add to cache
+        updateRecord(row, dbMatch, transactionMetadata);
+        childRecordsCache.push(dbMatch);
+        continue;
+      }
```
bamader, line 107: This looks good to me. Seems like the right order of checking matches so that there's no way we could wind up adding a child to the cache twice (because for it to be a DB Match, it can't have been a cache match else the earlier `if` would've caught it; and, once the DB match triggers, it gets added to the cache, so that if the record pops up in the future, the version of it in the cache will get modified). Just writing that out to make sure that order and logic makes sense to everyone else (otherwise we'll get hella duplicate records in the wizard).

bamader (approved):
This is an _excellent_ refactor to me. The front end is clean and trimmed, and the back-end is split up in a way that makes perfect sense. The directory structure is solid, the functions are efficient and readable, and honestly probably most important to me, the comments and documentation that now come with each function are _explicit_ about every assumption that's made in their use. We now have perfectly defined use cases where there was a lot of confusion and back and forth before. This is a _major_ win and I am all about it. Awesome work, JJ and Julia both!

src/controllers/enrollmentReports/map/map.test.ts
vonbearshark: That is what I was thinking, but the logic is not present from what I can see. @jlhogan ?

client/src/containers/Upload/PreviewChanges.tsx
```diff
@@ -0,0 +1,168 @@
...
+    setLoading(true);
+
+    try {
+      const resp: EnrollmentReportUploadResponse = await apiPost(
+        `enrollment-reports`,
+        childRecords,
+        { accessToken }
+      );
+
+      await revalidate();
```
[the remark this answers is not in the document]
vonbearshark: This forces the `react-swr` cache to revalidate the `childRecords` post-`POST`ing. Otherwise `react-swr` would just spit back the stale cache of records it has. It's existing code from the roster refactor. I would add a comment, but we are going to move away from `react-swr`, so I probably won't.

client/src/containers/Upload/getPreviewTableAccordionItems.tsx
```diff
@@ -86,6 +89,33 @@ export function getPreviewTableAccordionItems(
     }));
 }
 
+/**
+ * Build the table preview of a user's upload.
+ * @param mapResult
+ */
+export const formatUploadPreview = (mappedChildren: Child[]) => {
```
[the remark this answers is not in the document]
vonbearshark: @jlhogan seemed to have a further thought where this formatting could just be passed into the rendering component, maybe?

It makes sense that it's here now because we are passing records server <-> client. So the client has the records anyway, so we might as well. 

Also arguably we don't have a consistent philosophy on this since, like, roster site filtering and stuff is happening on the client. And that's a much larger lift, ha.

src/controllers/enrollmentReports/map/batchUpsert.ts
vonbearshark: 💯 . Wish it weren't so

client/src/containers/Upload/getPreviewTableAccordionItems.tsx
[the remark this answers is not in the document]
jlhogan: IMO this isn't "computation" which indeed should be done on the server side. This is formatting, which should happen in the client. That way, we're more flexible with what we display; changing what we show in the FE will only require changes to the formatting logic, as opposed to changing what the routes actually do and return.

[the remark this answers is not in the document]
bamader: Gotcha, totally makes sense. Like I said, good with this move either way, but the explanation makes it doubly so!

vonbearshark:
Just amended my most recent commit. No big changes. Fixed button styling and `batchUpserts` now requires manually specifying the `Entity`, since we are passing object literals 🙄

vonbearshark:
Deploying this to QA and kicking off e2e tests

src/controllers/enrollmentReports/checkErrors.ts
```diff
@@ -68,11 +68,13 @@ export const checkErrorsInChildren = async (
   });
 
   await Promise.all(
-    children.map(async (child) => {
-      child.validationErrors.map((e) =>
-        processErrorsInFields(child, e, errorDict, errorOccursIn)
-      );
-    })
+    children
+      .filter((c) => c.validationErrors?.length)
```
vonbearshark, line 72: This change might actually be due to the error reporting flow not behaving as expected. Still confirming, but it seems like we might be missing the error reporting in the `Upload` flow

src/controllers/enrollmentReports/map/batchUpsert.ts
```diff
@@ -0,0 +1,146 @@
...
+ * from typeORM), chunk the inserts into batches that will have <2000
+ * parameters per batch.
+ * @param user
+ * @param transaction
+ * @param entities
+ */
+export async function doBatchSave<T>(
+  user: User,
+  transaction: EntityManager,
+  entity,
```
vonbearshark, line 25: 🙄  thanks typeorm

src/controllers/enrollmentReports/checkErrors.ts
jlhogan: what does "error reporting" mean in this context?

client/src/shared/models/ChangeTags.ts
jlhogan: yeah IMO if we want to get back that more granular explantion of changes (i.e aged up) then we need to add that as a separate step once all rows have been processed to look at the tags on the record, the enrollments for the record, and determine if that's what happened. However i dont think that extra complexity is worth the potential user benefit. let's wait and see if folks really want/need that before adding a lot of additional work here.

src/controllers/enrollmentReports/map/batchUpsert.ts
jlhogan: should this be type T?

src/controllers/enrollmentReports/map/mapRows.ts
jlhogan: confirming that i understand this to be the correct way to check the cache -- that is, check the in mem cache, if no match then check the I/O source, if no match create new record

vonbearshark: Yep, both the code and seemingly expected behavior follows: 
1. check mem cache
2. if no mem cache: check db
3. if no db match: create new

src/controllers/enrollmentReports/checkErrors.ts
vonbearshark: Yeah, my comment is all weird.

This code was breaking because some children did not have `validationErrors`. Based on the refactor that now uses `validateObject` that seems expected for children who genuinely do not have errors. However, it also appears that the file upload flow is not catching errors in the upload file right now; nothing shows up in the review of missing info, there is no warning in the preview. So it's possible this change is not necessary and we actually _do_ want all children to have `validationErrors`, even if empty. Or at least this change _is_ necessary _and_ we also need to fix reporting the validation error independently. Still confirming this, though

src/controllers/enrollmentReports/map/batchUpsert.ts
vonbearshark: Ah yep good catch. Turns out it's actually `EntityTarget<T>`. I'll update the name to be similar as well (`entityTarget`). And looks like I missed updating the comment, too

src/controllers/enrollmentReports/map/map.test.ts
jlhogan: why "_should_" we? I used this as the north star for changes we support: https://help.ece-reporter.ctoec.org/update/

client/src/shared/payloads/EnrollmentReportUploadResponse.ts
```diff
@@ -0,0 +1,6 @@
+export interface EnrollmentReportUploadResponse {
+  newCount: number;
+  updatedCount: number;
+  withdrawnCount: number;
+  unchangedCount: number;
```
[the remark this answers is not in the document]
jlhogan: if we don't change anything on the record, we still need to include it in the "total number of records" In this file

however, as discussed IRL with jj, this doesn't need to be tabulated in the backend now that we're handing back the full child records we can just use the length of that array to display the totals to the end user. so it is _not_ necessary, but accounting for unchanged records in the total count in preview _is_

src/controllers/enrollmentReports/map/mapRows/updateRecord/updateBirthCertifiate.ts
jlhogan: sure, i think that's an exceedingly rare edge case and i went for simplest most likely use cases to support here. If folks think there's a big need to support more ways to update birth cert, i have no objection, i just am not sure it's necessary. If i understood, the thought was that a person can _ADD_ bc ifno for records that don't have it, but we don't necessarily support editing existing BC info.

src/controllers/enrollmentReports/map/mapRows/updateRecord/updateIncomeDetermination.ts
jlhogan: we actually don't even outlinne that new income info can be added via an upload, so we need to reconcile that in the copy in help site (https://help.ece-reporter.ctoec.org//update/) and the copy in the upload page. 

but IMO, we do not need to support editing an existing det, we just need to support batch redetermining income

src/controllers/enrollmentReports/map/mapRows/utils/child.ts
jlhogan: what we were doing before =/= what the AC was or really needs to be. im in favor of writing clear code that robustly handles the most common and valuable edit/update actions, and simply explictly not supporting more edge case-y things that we can require people to do via single in-tool operations

src/controllers/enrollmentReports/map/mapRows/updateRecord/updateIncomeDetermination.ts
vonbearshark: Isn't the distinction that `last` is more correct than `current`? Like, is it: `current` can be undefined if they withdrew, but `last` can mean either `current` or `last` (in withdrawn cases)? If so, I think we want `last`, right?

src/controllers/enrollmentReports/map/mapRows/getChildMatch.ts
```diff
@@ -0,0 +1,72 @@
...
+  const dbMatchQuery = thingHolder.transaction
+    .createQueryBuilder(Child, 'child')
+    .where('organizationId = :organizationId', {
+      organizationId: organization.id,
+    })
+    .andWhere('firstname = :firstname', { firstname: row.firstName })
+    .andWhere('lastname = :lastname', { lastname: row.lastName })
+    .andWhere('birthdate = :birthdate', {
+      birthdate: row.birthdate?.format('YYYY-MM-DD'),
+    });
```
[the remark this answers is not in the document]
vonbearshark: Seems like we are good type-wise, otherwise `format` would fail on a plain string, no?

src/controllers/enrollmentReports/map/mapRows/updateRecord/updateIncomeDetermination.ts
bamader: Oh duh, come on brain 🧠 . What's funny is that I commented here as if that were a bad thing  but wrote in my validation plan to test that dets  _don't_ get edited 🙃

src/controllers/enrollmentReports/map/mapRows/updateRecord/updateBirthCertifiate.ts
bamader: I understood the desired functionality differently but don't feel strongly about it, so I'll update the validation plan to just check for the general case.

bamader:
Validation plan: this one's gonna be a doozy 😕. There are fundamentally three categories of tests I can see that need performed: (1) basic uploading and variables in a record (can each field in a sheet be parsed successfully, what happens if the field is missing, etc.); (2) historical enrollments within the same sheet; and (3) updating records that exist in the DB. Each of these will have a bunch of things dependent on fields and errors and such, so apologies in advance for what's about to be a super long test plan.

##Basic uploading / field checks##
- [ ] Cases for field: first name
  - [ ] Server can parse first name from uploaded sheet
  - [ ] Omitting first name is flagged with an error
  - [ ] Records without first names shown in the preview screen with appropriate name adjustment
- [ ] Cases for field: middle name
  - [ ] Server can parse middle name from uploaded sheet
  - [ ] Omitting middle name is not flagged with an error
- [ ] Cases for field: last name
  - [ ] Server can parse last name from uploaded sheet
  - [ ] Omitting last name is flagged with an error
  - [ ] Records without last names shown in the preview screen with appropriate name adjustment
- [ ] Cases for field: suffix
  - [ ] Server can parse suffix from uploaded sheet
  - [ ] Omitting suffix is not flagged with an error
- [ ] Cases for field: sasid/uniqueId
  - [ ] Server can parse sasid from uploaded sheet
  - [ ] Omitting sasid is not flagged with an error
  - [ ] sasid/uniqueId is mapped to the appropriate field depending on the child's organization
- [ ] Cases for field: date of birth
  - [ ] Server can parse DOB from uploaded sheet
  - [ ] Omitting DOB is flagged with an error
- [ ] Cases for field: birth certificate type
  - [ ] Server can parse type from uploaded sheet
  - [ ] Omitting type is not flagged with an error and defaults to unknown
- [ ] Cases for field: certificate ID
  - [ ] Server can parse ID from uploaded sheet
  - [ ] Omitting ID is not flagged with an error if type isn't US, but is flagged if type is US
- [ ] Cases for field: town of birth
  - [ ] Server can parse birth town from uploaded sheet
  - [ ] Omitting birth town is not flagged with an error if type isn't US, but is flagged if type is US
- [ ] Cases for field: state of birth
  - [ ] Server can parse state from uploaded sheet
  - [ ] Omitting state is not flagged with an error with type isn't US, but is flagged if type is US
- [ ] Cases for fields: race fields
  - [ ] Server can parse each race field (american indian, asian, black, native hawaiian, white, not disclosed) from uploaded sheet
  - [ ] Omitting all race fields is not flagged with an error and defaults to not disclosed
- [ ] Cases for field: hispanic
  - [ ] Server can parse ethnicity from uploaded sheet
  - [ ] Omitting ethnicity is not flagged with an error and defaults to not collected
- [ ] Cases for field: gender
  - [ ] Server can parse gender from uploaded sheet
  - [ ] Omitting gender is flagged with an error
- [ ] Cases for field: receives disability services
  - [ ] Server can parse disability services from uploaded sheet
  - [ ] Omitting disability services is not flagged with an error and is parsed as not collected
- [ ] Cases for field: street address
  - [ ] Server can parse address from uploaded sheet
  - [ ] Omitting address is flagged with an error
- [ ] Cases for field: town
  - [ ] Server can parse town from uploaded sheet
  - [ ] Omitting town is flagged with an error
- [ ] Cases for field: state
  - [ ] Server can parse state from uploaded sheet
  - [ ] Omitting state is flagged with an error
- [ ] Cases for field: zipcode
  - [ ] Server can parse zipcode from uploaded sheet
  - [ ] Omitting zipcode is flagged with an error
- [ ] Cases for field: foster
  - [ ] Server can parse foster from uploaded sheet
  - [ ] Omitting foster is not flagged with an error
- [ ] Cases for field: homelessness
  - [ ] Server can parse homelessness from uploaded sheet
  - [ ] Omitting homelessness is not flagged with an error
- [ ] Cases for field: household size
  - [ ] Server can parse size from uploaded sheet
  - [ ] Omitting size is flagged with an error unless income is not disclosed
- [ ] Cases for field: annual income
  - [ ] Server can parse income from uploaded sheet
  - [ ] Omitting income is flagged with an error unless income is not disclosed
- [ ] Cases for field: determination date
  - [ ] Server can parse date from uploaded sheet
  - [ ] Omitting date is flagged with an error unless income is not disclosed
- [ ] Cases for field: provider
  - [ ] Server can parse provider from uploaded sheet
  - [ ] Omitting provider is flagged with an error
  - [ ] Omitting provider does not block creation of an enrollment
- [ ] Cases for field: site
  - [ ] Server can parse site from uploaded sheet
  - [ ] Omitting site is flagged with an error
  - [ ] Omitting site does not block creation of an enrollment
- [ ] Cases for field: care model
  - [ ] Server can parse model from uploaded sheet
  - [ ] Omitting model is not flagged with an error
  - [ ] Omitting model does not block creation of an enrollment
- [ ] Cases for field: age group
  - [ ] Server can parse age group from uploaded sheet
  - [ ] Omitting age group is flagged with an error
  - [ ] Omitting age group does not block creation of an enrollment
- [ ] Cases for field: enrollment start date
  - [ ] Server can parse date from uploaded sheet
  - [ ] Omitting date is flagged with an error
  - [ ] Omitting date does not block creation of an enrollment
- [ ] Cases for field: enrollment end date
  - [ ] Server can parse end date from uploaded sheet
  - [ ] Omitting end date is not flagged with an error
  - [ ] Omitting end date does not block creation of an enrollment
- [ ] Cases for field: exit reason
  - [ ] Server can parse exit from uploaded sheet
  - [ ] Omitting exit reason is not flagged with an error
  - [ ] Omitting exit reason does not block creation of an enrollment
- [ ] Cases for field: funding type
  - [ ] Server can parse funding type from uploaded sheet
  - [ ] Omitting funding type is flagged with an error
  - [ ] Omitting funding type does not block creation of a funding
- [ ] Cases for field: space type
  - [ ] Server can parse space type from uploaded sheet
  - [ ] Space types are mapped to equivalent day/time pairings (e.g. Full-Time <==> Full-Day)
  - [ ] Omitting space type is flagged with an error
  - [ ] Omitting space type does not block creation of a funding
- [ ] Cases for field: first funding period
  - [ ] Server can parse period from uploaded sheet
  - [ ] Reporting periods are parsed in the corrective "user-friendly" way
  - [ ] Omitting period is flagged with an error
  - [ ] Omitting period does not block creation of an enrollment
- [ ] Cases for field: last funding period
  - [ ] Server can parse period from uploaded sheet
  - [ ] Reporting periods are parsed in the corrective "user-friendly" way
  - [ ] Omitting period is flagged with an error
  - [ ] Omitting period does not block creation of an enrollment
- [ ] "New" records appropriately tagged as such
- [ ] Correct counts of new children shown in preview
- [ ] New records uploaded to roster matches new records shown in preview
- [ ] Roster upload can handle "large" sheets (approx. 1000 rows)

##Historical enrollment tests##
- [ ] Cache match function correct matches
  - [ ] Records with no provided sasid or unique ID
  - [ ] Records with a provided sasid
  - [ ] Records with a provided unique ID
  - [ ] Does not match records with any of first/last name, birthdate, or sasid/uniqueID different
- [ ] Can successfully add historical income determinations (with correct sorting)
- [ ] Can successfully add fundings to an existing enrollment
  - [ ] Fundings are sorted correctly
  - [ ] Information that ends previous funding is saved
  - [ ] Errors in ending previous funding do not block adding new funding, but are flagged
- [ ] Can successfully add enrollments to a record
  - [ ] Enrollments are sorted correctly
  - [ ] Information on previous enrollment exit is saved
  - [ ] Errors in ended enrollment do not block new enrollment, but are flagged
- [ ] Historical records still tagged as "new" in the preview screen (because historical non-DB match records are in the sheet but not already in the system)

##Updating records tests##
- [ ] DB query correctly matches children
  - [ ] Records with no provided sasid or unique ID
  - [ ] Records with a provided sasid
  - [ ] Records with a provided unique ID
  - [ ] Does not match records with any of first/last name, birthdate, or sasid/uniqueID different
- [ ] Update birth certificate
  - [ ] Do nothing if sheet row has unknown type
  - [ ] Do nothing if both types match
  - [ ] Update info for all fields if sheet row type is not unknown and differs from existing type
  - [ ] Updated records are given the Updated tag
- [ ] Update family address
  - [ ] Do not update homeless status if it's different
  - [ ] Allow changing of address fields
  - [ ] Updated records given the Updated tag
- [ ] Update enrollment
  - [ ] Match function for existing enrollment is correct
  - [ ] Save information to an existing enrollment (exit date / reason)
  - [ ] Can create new enrollments on existing record (with any supplied fields)
  - [ ] Absence of exit for old enrollment does not block creating new enrollment
  - [ ] Errors in enrollments (overlapping dates, no exit, no exit reason, missing fields in new enrollment) are flagged in preview where appropriate
  - [ ] Records that get a new enrollment are given the Updated tag
  - [ ] Records that receive only exit information are given the Withdrawn tag
- [ ] Update funding
  - [ ] Match function for existing funding is correct
  - [ ] Save information to an existing funding (last reporting period)
  - [ ] Can create new fundings on existing enrollment (with any supplied fields)
  - [ ] Absence of last period for old funding does not block creating new funding
  - [ ] Errors in fundings (overlapping reporting periods, funding starting before enrollment, no last reporting period, missing fields in new funding) are flagged in preview where appropriate
  - [ ] Records that get a new funding or receive new info for an old funding are given the Updated tag (even if the only update is a funding being given a last reporting period)
- [ ] Cannot update an existing income det (even if dates are identical); new det should be created if attempted
- [ ] Number of updated and withdrawn tags shown in preview is correct
- [ ] Records that move to the withdrawn part of the roster page are those correctly identified by tags and previews
- [ ] Back end can handle large sheet of updates (approx. 1000 rows)

bamader:
Ugh can't test this locally (after a 30 min. startup from winked keys 🙄 ) because we have that unresolved `cannot property 'setAtribute' of null` error that pops up after choosing a file, and now when you refresh you're taken back to step 1 and asked to choose a file again.

vonbearshark:
It's really frustrating, I know. Every piece takes a long time.

In that particular case you should be able to `X` it out in the top right corner

vonbearshark:
Just pushed a new commit with a fix for the validation errors not showing the Review or Preview pages. Hopefully this is good to go for validation now!

src/controllers/enrollmentReports/map/updates/updateEnrollmentFunding.ts
```diff
@@ -0,0 +1,184 @@
...
+  );
+  // Look up an existing funding that matches row funding
+  // (may not exist)
+  const matchingFunding = getFundingMatch(
+    newFunding,
+    matchingEnrollment?.fundings
+  );
+
+  // If row exits a current enrollment, end enrollment and funding
+  if (rowHasExitForCurrentEnrollment(newEnrollment, matchingEnrollment)) {
```
JordanGuinn, line 56: Confirming all of these scenarios are mutually exclusive, yeah?  (Just cuz of the `else`s)

src/controllers/enrollmentReports/map/updateRecord.ts
```diff
@@ -0,0 +1,35 @@
...
+ * - updating enrollment/funding information:
+ * 		- ending enrollment
+ * 		- ending funding
+ *    - adding new enrollment
+ *    - adding new funding
+ * @param row
+ * @param match
+ * @param transactionMetadata
+ */
+export const updateRecord = (
```
JordanGuinn, line 26: **This** 👏🏽👏🏽👏🏽 About as clear and readable as it gets - work inside each of the referenced functions is still necessarily complex, but said work is broken up thematically and then facilitated nicely from here.  If I wore a hat I'd take it off 🍻🍻

src/controllers/enrollmentReports/map/mapRows.ts
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * A param bag for the mapping process containing:
+ * - DB data for mapping, so it can be re-used in each mapping with only one initial lookup
+ * - the transaction EntityManager
+ * - the request user (needed for utils that get DB entities)
+ * - the cache of all mapped children
+ */
+export type TransactionMetadata = {
+  transaction: EntityManager;
```
JordanGuinn, line 28: Understand that this is largely necessary here in a number of places, but just a general thought that we should try and limit the passing around of DB transactions as much as possible, so as to avoid operations against those transactions with unintended consequences.

src/controllers/enrollmentReports/map/entities/family.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Create Family object (with IncomeDetermination) from FlattenedEnrollment
+ * source.
+ * TODO: Lookup existing families before creating new one
+ * @param source
+ */
+export const mapFamily = (
+  source: EnrollmentReportRow,
+  organization: Organization
+) => {
+  const homelessness: UndefinableBoolean = mapEnum(
```
JordanGuinn, line 18: I don't have any immediate solutions or anything for this, but just wanna call out that this feels needlessly complicated

src/controllers/enrollmentReports/map/createRecord.ts
```diff
@@ -0,0 +1,55 @@
...
+ * @param row
+ * @param organization
+ * @param transactionMetadata
+ */
+export const createRecord = async (
+  row: EnrollmentReportRow,
+  organization: Organization,
+  transactionMetadata: TransactionMetadata
+) => {
+  const family = mapFamily(row, organization);
```
JordanGuinn, line 34: nit: can we rename all these `mapX` functions to `createX` or `addX` instead?  At first glance here, it wasn't clear that anything was actually being written to row

src/utils/validateChild.ts
```diff
@@ -64,17 +62,13 @@ function distributeValidationErrorsToSubObjects<
   return parentObject;
 }
 
-export async function validateObject<T extends ObjectWithValidationErrors>(
-  object: T | undefined
-) {
-  if (!object) return;
-
-  let validationErrors = await validate(object, {
+export async function validateChild(child: Child) {
```
JordanGuinn, line 65: +1000 for not making a function generic when it currently has a single use case

JordanGuinn (approved):
OKAY, finally got through it all.  Overall this looks _great_, and a significant readability/comprehensibility improvement over what was there previously 🔥🔥 Got a handful of small nits and questions, but all-in-all 👌🏽👌🏽

src/controllers/enrollmentReports/map/utils/getChildMatch.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Looks in the mappedChildren cache for a record that matches the row on:
+ * - firstname
+ * - lastname
+ * - birthdates
+ * - sasid or unique id if it is present in the row
+ * @param row
+ * @param organization
+ * @param transactionMetadata
+ */
+export const getChildMatchFromCache = (
```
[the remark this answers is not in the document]
bamader: I believe we had unit tests for this matching function that were removed in this PR. Maybe we just put them back in?

src/controllers/enrollmentReports/map/updates/updateIncomeDetermination.ts
```diff
@@ -0,0 +1,57 @@
...
+    match.family.incomeDeterminations.push(newDetermination);
+    match.tags.push(ChangeTag.IncomeDet);
+  }
+};
+
+/**
+ * Determine whether an income determination mapped
+ * from an EnrollmentReportRow is a "new" determination.
+ *
+ * Considered new if:
```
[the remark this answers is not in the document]
bamader: Personally I really like them because it makes things explicit _without_ having to even scan the code. I hadn't thought of the maintenance burden with them, which is a good point, but I do know that one area where we've gone back and forth on in the past (sometimes frustratingly so because a ticket would say one thing and one of us would say something else entirely), is what our exact assumptions should be around some of the updates. (updating birth certificate info is a good example; the first batch update ticket initially said write code that would update under conditions X, Y, and Z; PR comments later said only update on X and Z; then a new ticket was written to add Y back in; and now this PR is taking Y and Z both out). And so I feel like using a comment to document those assumptions and design decisions is as much about being able to understand the code as it is to serve as a record of what the intended function is. So to me, it might actually be a strength to have double maintaining, because if we update the comment it means we're making the deliberate design choice to change how the code should work. Just my two cents!

client/src/components/FixedBottomBar/_index.scss
```diff
@@ -1,6 +1,7 @@
 .fixed-bottom-bar {
   position: fixed;
   bottom: 0;
+  left: 0;
```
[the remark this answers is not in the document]
vonbearshark: Yeah so that was the weird button UI you and I saw. We need this so that this fixed bottom bar always stays in the bottom left, rather than being pulled to the right if it's inside a container

client/src/containers/Upload/PreviewChanges.tsx
```diff
@@ -0,0 +1,168 @@
+import {
```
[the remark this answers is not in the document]
vonbearshark: _Mostly_, yes. We had to refactor `UploadWizard`, `ReviewMissingInfo` and this together similarly. And we now use the `childRecords` directly on the client. Otherwise unchanged

src/controllers/enrollmentReports/map/updates/updateEnrollmentFunding.ts
```diff
@@ -0,0 +1,184 @@
...
+ * - entry date is the same AND
+ * - if exit date exists, exit date is the same
+ *
+ * Enrollments with missing site age group and entry will not
+ * be considered matching, so enrollments with out site/agegroup/entry
+ * cannot be updated by file upload.
+ */
+export const getEnrollmentMatch = (
+  enrollmentFromRow: Enrollment,
+  enrollments: Enrollment[] | undefined
```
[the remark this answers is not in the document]
vonbearshark: Hm yeah good q. Maybe @jlhogan knows..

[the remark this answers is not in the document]
vonbearshark: And in this case, wouldn't it be the opposite? That the `row` from the CSV might not have the `enrollments`, but the `match` always should, right?

src/controllers/enrollmentReports/map/updates/updateEnrollmentFunding.ts
```diff
@@ -0,0 +1,184 @@
...
+  newEnrollment: Enrollment,
+  matchingEnrollment: Enrollment | undefined
+) => !!matchingEnrollment && !matchingEnrollment.exit && !!newEnrollment.exit;
+
+/**
+ * Determines if newEnrollment from row is a new enrollment.
+ * Enrollment is new if:
+ * - no existing matching enrollment was found
+ */
+export const rowHasNewEnrollment = (
```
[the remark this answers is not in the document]
vonbearshark: I went back and forth on this, since it is descriptive what is happening this way. But I think you're right to remove it, probably

src/controllers/enrollmentReports/map/updates/updateEnrollmentFunding.ts
vonbearshark: @jlhogan should confirm, but I think so?

src/controllers/enrollmentReports/map/mapRows.ts
vonbearshark: Oh, interesting. How do you mean?

src/controllers/enrollmentReports/map/createRecord.ts
vonbearshark: Yeah, I just noticed that convention last night, and agreed. Confusingly in some of these functions it's not..

src/controllers/enrollmentReports/map/updates/updateEnrollmentFunding.ts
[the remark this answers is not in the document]
bamader: That definitely sounds right

src/controllers/enrollmentReports/map/utils/getChildMatch.ts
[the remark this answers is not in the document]
vonbearshark: Hm, I don't see where they could have gone? I wouldn't be opposed to putting things back, but I am hesitant to create more tests for this PR right now

src/controllers/enrollmentReports/map/updates/updateIncomeDetermination.ts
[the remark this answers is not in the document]
vonbearshark: I am generally against them for the reasons @JordanGuinn mentioned. But in this case they are more trouble to remove imo

src/controllers/enrollmentReports/map/updates/updateEnrollmentFunding.ts
```diff
@@ -0,0 +1,184 @@
...
+  !matchingFunding.lastReportingPeriod &&
+  !!newFunding.lastReportingPeriod;
+
+/**
+ * Determines if a new funding from row is a new funding.
+ * Funding is new if:
+ * - existing matching funding was not found
+ * @param matchingFunding
+ */
+export const rowHasNewFunding = (matchingFunding: Funding | undefined) =>
```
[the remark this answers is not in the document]
vonbearshark: I hear you on these. I have gone back and forth. I have decided to keep them for now because I find them to be clarifying: why exactly this field means that the row has new funding, for example. And we do use them for useful tests

src/controllers/enrollmentReports/map/mapRows.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * This should not be necessary.
+   * We need `validateChild` to add validation errors to the records in memory.
+   * It uses `class-validator`, which requires that objects be class-instantiated.
+   * `typeorm` only allows us to do that through `create`.
+   * Doing so destroys the `tags` built on the object.
+   * So:
+   * 1. class instantiate `create`, 2. preserve `tags`, 3. then `validateChild`.
+   */
```
vonbearshark, line 132: 👀 ?

src/controllers/enrollmentReports/map/updates/updateEnrollmentFunding.ts
```diff
@@ -0,0 +1,179 @@
...
+    newFunding,
+    matchingEnrollment?.fundings
+  );
+
+  // If row exits a current enrollment, end enrollment and funding
+  if (rowHasExitForCurrentEnrollment(newEnrollment, matchingEnrollment)) {
+    matchingEnrollment.exit = newEnrollment.exit;
+    matchingEnrollment.exitReason = newEnrollment.exitReason;
+    if (matchingFunding)
+      matchingFunding.lastReportingPeriod = newFunding.lastReportingPeriod;
```
vonbearshark, line 60: 👀 ?

(this was throwing errors)

vonbearshark (approved):

vonbearshark:
Pushed up a fix for the breaking errors @bamader caught. And another commit addressing some of the lowering hanging nits. I agree with the feedback, but to limit the time and complexity of this change I prefer these were potential followups (or, just, not), but open to pushback. I highlighted a couple areas worth re-reviewing for these changes. Nothing major

Deployed to QA so it's good for more validation Monday.

src/controllers/enrollmentReports/map/mapRows.ts
JordanGuinn: The more functions have access to that transaction, the more opportunities there are for committing/rolling back said transaction, potentially without taking into account another consumer of it.

That being said, I'm now realizing this isn't really a transaction _at all_, so maybe it ought just be renamed?

src/controllers/enrollmentReports/map/updates/updateIncomeDetermination.ts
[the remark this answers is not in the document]
JordanGuinn: image

src/controllers/enrollmentReports/map/mapRows.ts
jlhogan: it is a transaction tho BUT ALSO we don't need to pass around the transaction anymore!!! we used to create each child as it was parsed, now we do that in batches at the end. we can just use normal typeorm `getManager` to grab the db data ahead of parsing/mapping, and then before we start on the batch creating we can start a transaction

JordanGuinn: `transaction-scoped entity manager` THAT'S the phrasing I was looking for!

Like it's definitively not a transaction, but rather an EntityManager that currently _happens_ to be spawned while opening a transaction in the singular instance where it's used.  But, as you just mentioned, this very same `EntityManager` can _also_ be generated simply by calling `getManager` independent of _any_ opened transaction.  In which case I'm even _less_ worried about it being passed around relatively haphazardly and just vote for a rename

vonbearshark: Haven't pushed yet but now just use `getManager` all over. We were doing that almost everywhere anyway

vonbearshark (approved):
Catching up:

I just pushed changes that should fix outstanding issues found in validation. Specifically:
- Removed passing `transaction` around, based on feedback
- SasId/uniqueId comparison match in mem cache*
- Update funding
- Large datasets

Noting these other issues, but marking as `wontfix`. Happy to discuss further, but tl;dr is that @jlhogan made compelling pitches IRL for the remote nature of these edge cases versus the code and documentation complexity it would take to address them. At any rate, they should not be blocking on these PR. Arguably we should forget about them until a case emerges, but we can also create tickets for followup if there's a strong counterargument.
- Omitting provider has no effect on record
- *while I do think I addressed the SasId matching issue, generally it's undefined behavior if a child record matches on everything but Sasid
- Adding historical income determinations
- Adding historical fundings
- Withdrawn funding tags: some undefined UX here. Tag counting is now mutually exclusive, which I do think is preferred, and `withdrawn` is the lowest weight, which I also think is probably preferred, so the only time we can trigger a withdrawn count/tag is when that is the only operation in the upload

I marked the changes in the code with my comments. We should not need to completely revalidate, but it would be nice to confirm the fixes and give it a once-over @bamader

bamader (approved):
New code looks good--will comment back here shortly this evening after re-running a smattering of validations (just to verify core functionality and the new fixes).

bamader:
Okay validations are good! I put the newly updated code through a little bit more of a ringer than I intended and it held up well. All base functionality remains, no regressions in tests, and successful fixes for everything JJ reported. Not bothering to mark off the individual tests in the validation plan cuz I feel like the other comments address those well individually. Also, after seeing the existing state of the code and thinking of how often the remaining (uncovered) cases are actually likely to come up, I concur with JJ and Julia that we're okay deferring those other problems to other tickets (or ignoring them entirely).

The only one I'd push back on is the withdraw tag and count. I actually believe this is something we should very much address in a fast follows ticket. If I'm a provider uploading a batch update and I'm withdrawing a bunch of records (even something like 5 or 10% of my total sheet volume; really anything more than 1 or 2 records), if that count shows up as 0 I'm going to be concerned that something is broken or I'm doing something wrong. We clearly tag the withdrawn records, so we've already got the base functionality; I would strongly advocate for putting withdrawn over edited in the count hierarchy because I think that's more intuitive. The withdraw operation feels more important than the edited operation, since it's a transition between states for a  child (active to non-active) rather than just modifying a record that's remaining in its state. Also, edited is likely to happen way more often than withdrawn in an upload, so a "wrong" or off count there is less likely to be noticed and a cause for alarm than a wrong withdrawn count, which will be highly visible if it's always at 0.

TLDR: This PR has the official validation green seal of approval 🐳 (there's no seal emoji in markdown 😒), 99% of the validation and comment discussions seem super reasonable, and I think we do need to fix the withdraw count issue in a fast follows.

PR #1286, opened 2021-04-20
Bump z-index of all modals

JordanGuinn:
## Background
Modals sometimes don't sit highest on the browser window - this PR should fix that.

## GitHub Issue
https://github.com/ctoec/component-library/issues/133

## Associated PRs
https://github.com/ctoec/component-library/pull/148

## Validation Plan
- [x] Confirm that the **withdraw child** modal sits in front of all other content on the page
- [x] Confirm that the **delete child** modal sits in front of all other content on the page
- [x] Confirm that the **roster submit success** modal sits in front of all other content on the page
- [x] Confirm that the **previous month filter** modal sits in front of all other content on the page
- [x] Confirm that the **change enrollment** modal sits in front of all other content on the page

## Automated Testing

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@ctoec/component-library": "^1.1.14",
+    "@ctoec/component-library": "https://github.com/ctoec/component-library.git#modal-fix",
     "@openid/appauth": "^1.2.7",
     "@testing-library/dom": "^7.24.1",
     "@testing-library/jest-dom": "^4.2.4",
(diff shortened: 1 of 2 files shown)


bamader (approved):
From quick pass validation, it looks like all modals in the app (of which there are 5--withdraw record, delete record, change enrollment, data submit success, and filter roster for previous month) now behave the same way, with no background text breaking through the overlay.

vonbearshark:
qq, if we bump the component-lib version, we'll get the new date picker changes, right? probably not what we want right away, fyi

JordanGuinn:
That's correct - my plan was gonna be to cut a `1.1.14.1` version and do some rebasin', which isn't _great_ but will get us what we need

jlhogan (approved):

Issue #1227, opened 2021-03-08
ECE Reporter unit tests are leaking memory


JordanGuinn:
## Background
Every now and again, during a unit test execution of the `data-collection` repo in CircleCI, the following error will be thrown and kill the entire workflow:

`Failure message: ENOMEM: not enough memory`

This is almost assuredly indicative of us leaking memory in some capacity in our unit test suite, most likely during either the setup or teardown portion.  We should instrument memory consumption of our unit tests to determine exactly what's going on here, and fix whatever's causing this leak so that our unit test suite will run more consistently.

## Acceptance Criteria
- Underlying cause of unit test memory leak is discovered and fixed
- Unit test suite can be run multiple times consecutively without any notable change in memory consumption

PR #1290, opened 2021-04-22
Reduce unit test memory usage

JordanGuinn:
## Background
The Docker executors our build/deploy/tests workflows run in on CircleCI have a pretty limited memory ceiling (given we're using the free version of CircleCI).  With that in mind, this PR adds the `--runInBand` parameter to our Jest unit test commands, which, [per Jest documentation](https://jestjs.io/docs/cli#--runinband), runs `"all tests serially in the current process, rather than creating a worker pool of child processes that run tests"`.  This also appears to have the added bonus of somehow [speeding up our total unit test execution time slightly](https://jestjs.io/docs/troubleshooting#tests-are-extremely-slow-on-docker-andor-continuous-integration-ci-server)!

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1227

## Validation Plan
- [x] Unit tests CircleCI workflow was run several times in succession without failure

## Automated Testing

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -48,7 +48,7 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "TZ=America/New_York CI=true react-scripts test --env=jest-environment-jsdom-sixteen --silent --testPathIgnorePatterns ./*/integrationTests --logHeapUsage",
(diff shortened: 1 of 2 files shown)


vonbearshark (approved):
🙌🏻 . I was confused at first bc I thought we already have it, and we do for the _integration_ tests

bamader (approved):

PR #1291, opened 2021-04-22
Slightly reduce excessive number of roster calls made

JordanGuinn:
## Background
While we have tickets created long-term to both [replace SWR with a library that supports a pull-through cache mechanism](https://app.zenhub.com/workspaces/ece-dev-board-5ff506028d35f30012e0e937/issues/ctoec/data-collection/1009) as well as [refactor our most expensive DB validators](https://app.zenhub.com/workspaces/ece-dev-board-5ff506028d35f30012e0e937/issues/ctoec/data-collection/1289), this PR provides some immediate relief to our DB in terms of number of GET children API requests made.  More specifically, this PR should prevent subsequent GET children API requests under the following circumstances:
- If an OEC admin user has logged in (as they don't have a roster)
- If a user clicks to another application on their computer, then clicks back
- If a user loses their internet connection briefly, and then regains it (they probably ought just refresh anyway)

## GitHub Issue
N/A

## Validation Plan
- [ ] Confirm that no GET children requests are made when logged in as an OEC admin
- [ ] Confirm that general updates made to children in-app update the cache as they did previously

## Automated Testing

### DIFF
--- a/client/src/contexts/RosterContext/RosterContext.tsx
+++ b/client/src/contexts/RosterContext/RosterContext.tsx
@@ -71,6 +71,11 @@ const RosterContext = React.createContext<RosterContextType>(defaultCtx);
 
 const RosterProvider: React.FC = ({ children }) => {
   const { user } = useContext(UserContext);
+
+  if (user?.isAdmin) {
+    return <></>;
+  }
+
   const history = useHistory();
   const query: RosterQueryParams = parse(history.location.search);
 
--- a/client/src/hooks/useAuthenticatedSWR.ts
+++ b/client/src/hooks/useAuthenticatedSWR.ts
@@ -42,5 +42,8 @@ export function useAuthenticatedSWRInfinite<T>(
     const key = !path || !accessToken ? null : [path, accessToken];
     return key;
   };
-  return useSWRInfinite(getKey) as SWRInfiniteResponseInterface<T, string>;
+  return useSWRInfinite(getKey, {
+    revalidateOnFocus: false,
+    revalidateOnReconnect: false,
+  }) as SWRInfiniteResponseInterface<T, string>;
 }


client/src/contexts/RosterContext/RosterContext.tsx
```diff
@@ -71,6 +71,11 @@ const RosterContext = React.createContext<RosterContextType>(defaultCtx);
 
 const RosterProvider: React.FC = ({ children }) => {
   const { user } = useContext(UserContext);
+
+  if (user?.isAdmin) {
+    return <></>;
+  }
+
```
vonbearshark, line 78: hm yeah this might be trickier than expected. This won't work, right? At least it needs `<>{children}</>`, right? And even then we have to provide `defaultCtx`. That _could_ be an issue for children that subscribe to that context...

JordanGuinn: Are there _any_ components that OEC admins would have access to that would need to subscribe to these roster changes?  I could very well be oversimplifying this, but it looks like all the components that need to be subscribed to roster changes made here are only visible by non-admin users

vonbearshark: Nothing comes to mind immediately. @jlhogan might know better. It could be simple! We probably do need to pass `children` and `defaultCtx` either way, but maybe that's all we need...

bamader: Nothing comes to mind for me either. What's the advantage of passing back `defaultCtx` though?

bamader (approved):

client/src/contexts/RosterContext/RosterContext.tsx
vonbearshark: If you truly have no subscribers then I _guess_ it's fine. But theoretically, even just based on the typing alone, we _should_ always return the context's shape. And these defaults _should_ also let unknown/unintended subscribers handle the empty data usecase gracefully. Like, rather than throwing up that `childrenRecords` doesn't exist it should just do nothing when it's empty

jlhogan (approved):

JordanGuinn:
Closing, cuz this is super old and will inevitably be addressed in the roster redesign work 😭

Issue #1271, opened 2021-04-09
Redetermining income overwrites the current income determination


JordanGuinn:
## Background
Found during validation of https://github.com/ctoec/data-collection/issues/1265.  Whenever a user redetermines the income for a given enrollment, the income determination that was there previously is overwritten, as opposed to a new one being created.

## Repro Steps
- Navigate to your roster.
- Go the edit view of a given child that already has an income determination associated with their enrollment.
- Go to the `Family Income` tab
- Click `Redetermine`
- Update the data in the various income fields
- Save and observe

## Expected Behavior
- New income determination is added to the enrollment.

## Actual Behavior
- Current income determination is overwritten with new data.

PR #1297, opened 2021-04-27
Redetermine income creates a new determination instead of updating the last one

nickbristow:
## Background
When clicking redetermine income, the app updates the existing determination instead of the expected behaviour of creating a new entry

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1271

## Validation Plan
Manual validation:
- Create new user with determination
- use redetermine link
- use edit link
- test no income option

## Automated Testing

## Additional Context


client/src/components/Forms/FamilyIncome/Form.tsx
```diff
@@ -88,8 +90,10 @@ export const FamilyIncomeForm: React.FC<FamilyIncomeFormProps> = ({
 
   let determination: IncomeDetermination;
   const dets = child?.family?.incomeDeterminations || [];
-  if (inCreateFlow) {
-    determination = dets[0] || ({} as IncomeDetermination);
+
+  if (inCreateFlow || redetermine) {
+    const { id, ...newDet } = dets[0] ?? {};
+    determination = newDet as IncomeDetermination;
```
vonbearshark, line 96: Is `as IncomeDetermination` still necessary with `let determination: IncomeDetermination;`

vonbearshark (commented):
I see this is `Draft`, but code lgtm

client/src/components/Forms/FamilyIncome/Form.tsx
nickbristow: typescript complains without it since the ID is removed.

vonbearshark: And we _don't_ just want to add the `id`?

nickbristow: yeah, because it would grab the current id then it would update that entry instead of creating a new one

jlhogan: why does the new det need to be object spread from the existing det ? im confused as to why the new det doesn't get added to the determination array, versus being like reversed engineered from changes applied to the existing det but then not saved in the backend and rather pulled out and plopped into a new det

JordanGuinn (approved):

vonbearshark (approved):

vonbearshark:
I was able to confirm on QA with the manual steps listed 👍🏻 . Going to merge this and call it validated, but folks are free to double-check my work

Issue #1298, opened 2021-04-28
Users can create organizations w/new names


JordanGuinn:
## Background
Admin OEC users will want the ability to create new organizations.

## Acceptance Criteria
- `Create New Organization` button is added to admin org list view
- A new organization can be created with `Name` **only**, for the time being
   - `Sites`, `Funding Spaces` and `Users` sections will all come in later tickets
- Error is displayed to the user if the name they've provided is already taken

## Designs
[Full designs can be found in Figma file](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=97%3A728).

![ECE Reporter – Create org.png](https://images.zenhubusercontent.com/6050f59bb12f6b5f4da9aeb1/6abd3f73-7d67-4c9f-8100-c8c94595daa5)

PR #1322, opened 2021-05-10
Add create org capability to admin users

bamader:
## Background
This changeset gives admin users the ability to add organizations with _name only_ from the organization view of the admin page. If an entered name is unique (i.e. not being used by another org), an org is created with that name. If the name is taken, the user is alerted with an error banner. Since Nick is currently finalizing that work, this PR will get updated once his changes land in the epic (because I'll move the create organization button from the overview page to the organization page).

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1298

## Validation Plan
Created an admin level user, logged into the app, and attempted to create unique (which were successful) and non-unique (which were not) organizations using the name field.

## Automated Testing


vonbearshark (commented):
added some thoughts. happy to take another look when it's ready but generally lgtm so far!

client/src/routes.ts
```diff
@@ -71,6 +72,12 @@ export const routes: RouteConfig[] = [
     component: OrganizationOverview,
     unauthorized: false,
   },
+  {
+    path: '/createOrg',
```
jlhogan, line 76: kebab case for consistency please

src/routes/organizations.ts
```diff
@@ -0,0 +1,32 @@
...
+export const organizationsRouter = express.Router();
+
+organizationsRouter.post(
+  '/',
+  passAsyncError(async (req: Request, res: Response) => {
+    try {
+      const { name } = req.body;
+      const nameExists = await doesOrgNameExist(name);
+      if (nameExists) {
+        res.status(200).send({ id: undefined });
```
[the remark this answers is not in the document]
jlhogan: agreed that it doesn't make sense to fail on this post route and return a 200. I think this should be 4XX

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const createNewOrg = async () => {
+    const { id } = await apiPost(
+      `organizations/`,
+      { name: newOrgName},
+      { accessToken, jsonParse: true }
+    ).catch((err) => {
+      throw new Error(err);
+    });
+    const newAlerts: AlertProps[] = [];
```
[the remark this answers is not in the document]
bamader: Yep, mutually exclusive. Was trying to be economical and just have one setAlerts call that worked based on an if/else push, but it's just as good to set alerts directly rather than instantiate a new list first.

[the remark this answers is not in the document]
jlhogan: i would go for clear code over avoiding this repetition here

src/routes/organizations.ts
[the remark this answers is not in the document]
bamader: Figured out why it wasn't working before with just sending a 4XX error. The client `api` util expects an error object with a `message` and `data` body so it can return a rejected promise; I wasn't providing those the first time and was just sending a 400, so it was triggering the catch rather than any kind of logic. This time, sending back a 400 error with meaningful `message` and `data` fields did the trick.

bamader:
Ready for a re-review

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -0,0 +1,74 @@
...
+      
+      // Special 4XX error if an org with given name already existed
+      if (err.message.includes('exists')) {
+        setAlerts([{
+          type: 'error',
+          heading: 'Organization already exists',
+          text: `An organization with name "${newOrgName}" already exists.`
+        }]);
+      }
+      else throw new Error(err);
```
vonbearshark, line 31: do we typically `throw` or `log` errors? (again, I dunno, jw)

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const createNewOrg = async () => {
+    const res = await apiPost(
+      `organizations/`,
+      { name: newOrgName},
+      { accessToken, jsonParse: true }
+    ).catch((err) => {
+      
+      // Special 4XX error if an org with given name already existed
+      if (err.message.includes('exists')) {
```
vonbearshark, line 24: re: comment later on: would be nice to condition this on the status code rather than this, but fine either way

src/routes/organizations.ts
```diff
@@ -1,10 +1,36 @@
...
 export const organizationsRouter = express.Router();
 
+organizationsRouter.post(
+  '/',
+  passAsyncError(async (req: Request, res: Response) => {
+    try {
+      const { name } = req.body;
+      const nameExists = await controller.doesOrgNameExist(name);
+      if (nameExists) {
+        throw new BadRequestError('Name already exists.', { id: undefined });
```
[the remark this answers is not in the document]
vonbearshark: especially since we condition on `id?` anyway

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { FixedBottomBar } from '../../components/FixedBottomBar/FixedBottomBar';
+import AuthenticationContext from '../../contexts/AuthenticationContext/AuthenticationContext';
+import { useAlerts } from '../../hooks/useAlerts';
+import { apiPost } from '../../utils/api';
+import { getH1RefForTitle } from '../../utils/getH1RefForTitle';
+
+const CreateOrg: React.FC = () => {
+  const h1Ref = getH1RefForTitle();
+  const { accessToken } = useContext(AuthenticationContext);
+  const [newOrgName, setNewOrgName] = useState<string>("");
```
vonbearshark, line 13: Do we need the `<string>` with the initial value

vonbearshark (approved):
some qs but seems fine

src/routes/organizations.ts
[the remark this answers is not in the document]
bamader: Oh I like the idea of having a new error class. They're not complex to make and even as I wrote this I had the feeling that `BadRequest` wasn't fully the right thing to capture. I'll make that swap.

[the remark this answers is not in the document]
bamader: Also yeah we don't need to send back data if the error is already specific enough

src/routes/organizations.ts
```diff
@@ -1,10 +1,36 @@
...
+      }
+      else {
+        const newOrg = await getManager().save(
+          getManager().create(Organization, { providerName: name })
+        );
+        res.status(201).send({ id: newOrg.id });
+      }
+
+    } catch (err) {
+      if (err instanceof ApiError) throw err;
```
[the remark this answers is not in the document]
bamader: I prefer the `else` here because I think it helps a bit with flow and clarity, and you're right that multiple errors can't actually be thrown here

src/routes/organizations.ts
```diff
@@ -1,10 +1,36 @@
...
+organizationsRouter.post(
+  '/',
+  passAsyncError(async (req: Request, res: Response) => {
+    try {
+      const { name } = req.body;
+      const nameExists = await controller.doesOrgNameExist(name);
+      if (nameExists) {
+        throw new BadRequestError('Name already exists.', { id: undefined });
+      }
+      else {
```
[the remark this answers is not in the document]
bamader: It's definitely _my_ style. All code I write, the `else` goes on a newline. I don't think it's ever been a problem so I've been 🤷‍♂️  about it (cuz if nobody else cares, why not do what I like lol).

client/src/containers/CreateOrg/CreateOrg.tsx
bamader: For the API backend, we consistently `throw` errors. Check out any catch block or `try` part where access isn't there and it's always throw. On the client side, I also believe we tend to throw, but the water is murkier and I don't think any firm decision was ever made to that effect.

client/src/containers/CreateOrg/CreateOrg.tsx
bamader: So initially I definitely tried to condition on status code, but the problem is that the `api` file in the utils directory of the client has specific logic for what to do with any error that comes back with a `4XX` status. Check out line 101: you'll see that the error itself isn't actually returned, it's just a rejected promise with the message giving the reason for rejection (since we don't return an `{ id: undefined }` data anymore). So conditioning on the message seemed much easier to me and looks like the way this api file was set up anyway.

client/src/containers/CreateOrg/CreateOrg.tsx
bamader: Nope!

src/routes/organizations.ts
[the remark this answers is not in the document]
vonbearshark: 👍🏻  ignore me then!

client/src/containers/CreateOrg/CreateOrg.tsx
vonbearshark: that makes sense

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const createNewOrg = async () => {
+    const res = await apiPost(
+      `organizations/`,
+      { name: newOrgName},
+      { accessToken, jsonParse: true }
+    ).catch((err) => {
+      
+      // Special 4XX error if an org with given name already existed
+      if (err.includes('exists')) {
```
jlhogan, line 24: does the api middleware parse away the status code? that feels like a cleaner check to me but it might not exist at this point (which is stupid + my fault if true)

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -0,0 +1,74 @@
...
+          type: 'error',
+          heading: 'Organization already exists',
+          text: `An organization with name "${newOrgName}" already exists.`
+        }]);
+      }
+      else throw new Error(err);
+    });
+    
+    // Got a DB ID from the newly created org
+    if (res?.id) {
```
jlhogan, line 35: again, if we're here and there's no error then res.id exists so i dont think this check is necessary and if anything it's more confusing thn not to my brain

client/src/containers/CreateOrg/CreateOrg.tsx
jlhogan: depends on the error. if its something that isn't fully going to block/break what the user is doing then can log it. Throwing the error should trigger the display of the error boundary which will alert the user to contact support or refresh the page. there's no one decision made b/c different errors have diff effects and thus are handled differently

src/routes/organizations.ts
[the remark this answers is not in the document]
jlhogan: can we just get rid of this else? throwing the error will break out of the func

src/routes/organizations.ts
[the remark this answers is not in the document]
jlhogan: i dont prefer unnecessary elses -- leads to extra indented code which ends up harder to read IMO.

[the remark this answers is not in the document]
vonbearshark: that's fair enough. We have two situations here with if -> throw, then an else. I would most prefer consistency and no `else` seems fine

client/src/containers/CreateOrg/CreateOrg.tsx
bamader: The `api.ts` util has an `if` that checks if the status returned is a 4XX error. All it does after that is return a rejected promise containing an error message (in all cases), and the error data if the error that the server threw had a `data` property set. So I think the answer to your question is yes, the middleware parses away the status code, and the only thing that's returned to the client is the promise rejection message.

bamader: Also just realized as I'm working on the create sites portion of this page that we're going to need to check the message anyway. Since we're doing an all or nothing approach to creation, if any of the site names are taken and we fail creating the org, that will still return a 409 (since it's the same conflict as with an org name being taken). That means we'll have to check the message to know which failure reason to show the user.

client/src/containers/CreateOrg/CreateOrg.tsx
bamader: I got rid of the check and just had the code execute, and it was falling through after the catch so that the success alert was always being shown regardless of whether the call actually succeeded. I kept the check gone but had to switch to `.then()` to prevent fallthrough. Functionality is unchanged though.

vonbearshark: New code looks good to me, but 1. seems like there's an unrelated comment still and 2. we shouldn't have to send the `id` at all now, right?

bamader: Ah yeah the `got a DB ID` comment. The only reason I sent back ID is because in some other places our post requests send back data, but I just found some posts that also send back nothing but a status (i.e. I thought it was our convention but isn't). So will remove the comment and the ID sending!

src/routes/organizations.ts
```diff
@@ -1,19 +1,37 @@
...
 export const organizationsRouter = express.Router();
 
+organizationsRouter.post(
+  '/',
+  passAsyncError(async (req: Request, res: Response) => {
+    try {
+      const { name } = req.body;
+      const nameExists = await controller.doesOrgNameExist(name);
+      if (nameExists) throw new ResourceConflictError('Name already exists.');
+      const newOrg = await getManager().save(
```
vonbearshark, line 17: nit nit: no need for variable

vonbearshark (approved):

JordanGuinn (approved):

JordanGuinn:
Confirmed the following during validation of this work:

- [ ] Organizations are created successfully when name is available
   - [ ] Success messaging indicates as much
- [ ] Organizations are NOT created successfully when name is unavailable
   - [ ] Error messaging indicates as much

I'd say this is ready for merge, once e2e are green

bamader:
Aaaaaaaand we're greeeeeeeeen

Issue #1318, opened 2021-05-07
As an OEC Admin User, I can view a table of all users


JordanGuinn:
## Background
As an OEC Admin User, I can view a table of all users, along with the organizations they're associated with.

## Acceptance Criteria
- `Users` tab is added to the navbar at the top of the page
- On navigation to the `Users` tab, admins are presented with a **non-paginated list view** of all users, displaying
   - `Last Name, First Name`
   - Email Address
   - Comma separated list of organizations
- Default sort is alphabetically, by **last name**
- `Create new user` button is ignored for the time being, to be addressed in another ticket

## Designs
[Full Figma designs can be found here.](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=2%3A2)

#### Users List View
![Screen Shot 2021-05-07 at 9.13.56 AM.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/56a965cd-0c4a-435e-a6e3-a9fa4a1b1cbf)

PR #1323, opened 2021-05-11
1318 admin user tab

nickbristow:
## Background
As an OEC Admin User, I can view a table of all users, along with the organizations they're associated with.
## GitHub Issue
#1318

## Validation Plan
Verify as admin
verify regular user cannot access

## Automated Testing

## Additional Context


vonbearshark (approved):
lgtm

client/src/containers/Users/UsersTable.tsx
```diff
@@ -0,0 +1,43 @@
...
+  },
+  {
+    name: 'Email address',
+    className: 'text-no-wrap',
+    sort: (row) => row.email ?? '',
+    cell: ({ row }) => <td>{row.email ?? ''}</td>,
+  },
+  {
+    name: 'Organization(s)',
+    cell: ({ row }) => <td className="maxw-card-lg">{row.organizations}</td>,
```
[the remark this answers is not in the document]
nickbristow: done in the sql query

nickbristow:
> I meant `2 for Unused variable, import, function or class`, in the LGTM bot, not the `local` tests

ah...yes. I glossed over it because i assumed.

client/src/containers/Users/Users.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { UsersTable } from './UsersTable';
+import { apiGet } from '../../utils/api';
+
+const Users: React.FC = () => {
+  const { user } = useContext(UserContext);
+  const h1Ref = getH1RefForTitle();
+  const { accessToken } = useContext(AuthenticationContext);
+
+  const [data, setData] = useState<UserSummary[]>([]);
+  apiGet(`users/`, accessToken).then((res: UserSummary[]) => setData(res));
```
bamader, line 16: Nit: I think this should have an `await` in front of it or be inside a `useEffect` that returns if the access token or response isn't present yet; that's typically how we prevent weird flashing behavior as screens load and change.

bamader (approved):

client/src/containers/Users/Users.tsx
vonbearshark: Does the new code prevent the flashing? Not clear to me how..?

JordanGuinn (approved):

client/src/containers/Users/Users.tsx
nickbristow: @vonbearshark i think it would really just prevent it from calling the api before its ready.

client/src/containers/Users/Users.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const { user } = useContext(UserContext);
+  const h1Ref = getH1RefForTitle();
+  const { accessToken } = useContext(AuthenticationContext);
+
+  const [data, setData] = useState<UserSummary[]>([]);
+
+  useEffect(() => {
+    (async function loadUsers() {
+      if (user && accessToken) {
+        await apiGet(`users/`, accessToken).then((res: UserSummary[]) =>
```
jlhogan, line 20: shouldn't need the trailing slash, right?

client/src/containers/Users/UsersTable.tsx
```diff
@@ -0,0 +1,41 @@
+import { Column, Table } from '@ctoec/component-library';
+import React from 'react';
+import { UserSummary } from '../../shared/payloads/UsersResponse';
+
+type UserTableProps = {
+  users?: Array<UserSummary>;
+};
+
+export const UsersTable: React.FC<UserTableProps> = ({ users = [] }) => (
+  <Table<UserSummary>
```
jlhogan, line 10: so this is something I just though of but might it be worth it to try pulling in the carbon data table component now instead of continuing to build more tables with our very complex and very bespoke table component?

client/src/containers/Users/UsersTable.tsx
```diff
@@ -0,0 +1,41 @@
...
+  />
+);
+
+const TableColumns: () => Column<UserSummary>[] = () => [
+  {
+    name: 'Name',
+    sort: (row) => row.name,
+    cell: ({ row }) => (
+      <th scope="row">
+        {/* Deactivate links until org detail pages are built */}
```
jlhogan, line 25: copy pasta comment

src/controllers/users.ts
```diff
@@ -28,3 +28,12 @@ export const addDataToUser = async (user: User) => {
   user.sites = sites;
   user.organizations = organizations;
 };
+
+export const getUsers = async (): Promise<User[]> =>
+  getManager().query(
```
jlhogan, line 33: can we try to keep these handwritten queries styled a bit better for readability? my pref:
```
SELECT
  u.id,
  CONCAT(u.lastName, ', ', u.firstName) as name,
  u.email,
  STRING_AGG(o.providerName,' ', ') as organizations
FROM [user] u 
JOIN organization_permission op on u.id = op.userId
JOIN organization o on op.organizationid = o.id
GROUP BY u.id, u.lastName, u.firstName, u.email
```

jlhogan (approved):

client/src/containers/Users/UsersTable.tsx
JordanGuinn: **Fully** agreed, assuming this doesn't hugely bump the amount of effort of this ticket (which I wouldn't think would be the case, this is as a simple of a table as can be)

client/src/containers/Users/UsersTable.tsx
```diff
@@ -0,0 +1,42 @@
+// import { Column, Table } from '@ctoec/component-library';
```
JordanGuinn, line 1: nit: this can be deleted now

client/src/containers/Organizations/Organizations.tsx
```diff
@@ -14,9 +14,17 @@ const Organizations: React.FC = () => {
   const { accessToken } = useContext(AuthenticationContext);
 
   const [data, setData] = useState<OrganizationSummary[]>([]);
-  apiGet(`organizations/`, accessToken).then((res: OrganizationSummary[]) =>
-    setData(res)
-  );
+
+  useEffect(() => {
+    (async function loadUsers() {
```
JordanGuinn, line 19: 1.  Why this change here?
2.  Presumably this should be named `loadOrganizations`?

client/src/containers/Users/UsersTable.tsx
```diff
@@ -0,0 +1,42 @@
...
+  TableBody,
+  TableCell,
+} from 'carbon-components-react';
+
+type UserTableProps = {
+  users?: Array<UserSummary>;
+};
+
+export const UsersTable: React.FC<UserTableProps> = ({ users = [] }) => (
+  <Table className="oec-table oec-table--full-width">
```
JordanGuinn, line 18: Oh HELLO there, Carbon

JordanGuinn (approved):

client/src/containers/Organizations/Organizations.tsx
nickbristow: 1. @bamader pointed this out for the users page which is essentially the same as organizations. This prevents unnecessary reloading and only loads the data once its ready. 
2. Yes, updated name

bamader (approved):

bamader:
Validated that this works. Regular users cannot access the admin/user page, and on the page, all users are displayed with their associated organizations. Since we don't yet have the functionality to create a user, that's basically all we have to check for this.

PR #1324, opened 2021-05-12
Updated snapshots to fix local on base

vonbearshark:
## Background
I don't know how this happened, but pretty sure this is all that is needed to fix `local` tests on `base`. Seems like it might be related to the date picker changes? Maybe @nickbristow can confirm these snapshot diffs, but seem fine to me. Also updated the audit failures. Generally seems reasonable although the `.lock` did bump some deps not explicitly touched in the `resolutions` changes, but seems fine.

## Automated Testing
- [x] All unit tests are passing. (I mean, I would hope)


vonbearshark:
hm.. CI build is failing. local seems fine?

bamader (approved):
Yeah this is some weirdness. Good eye to find which ones were different, and I'm almost positive you're right that it had to do with the date picker. In any event, looks like this should get sorted out pretty soon.

vonbearshark:
Local is also broken, I see now. Seems like it has to do with the `moderate` `postcss` bump, 

`postcss-custom-media: postcss.plugin was deprecated. Migration guide: https://evilmartians.com/chronicles/postcss-8-plugin-migration`

JordanGuinn:
Confirmed - looks like CRA (and by proxy `react-scripts`) [doesn't yet support PostCSS 8](https://github.com/facebook/create-react-app/issues/9664).  As a final `why not` measure, I tried pinning resolutions of postcss in `component-library` to `^8.2.10` as well, but to no avail (still encountered the same error).

I think the only _real_ option in front of us at THIS point is to just ignore these failures, until `react-scripts` adds support for PostCSS 8 😕

bamader:
@JordanGuinn Do you think that's likely any time soon? I guess if you've already tried pinning versions and resolutions I won't waste my time doing the exact same thing. But is there anything else we can do on this?

JordanGuinn:
Who knows, honestly.  The only alternative available to us beyond that is contributing to the CRA repository, which... no.

vonbearshark:
Can't approve since it's technically mine, but lgtm with the latest

JordanGuinn:
Once the e2e go green, I'll merge to `base` and the epic admin view branch, then it'll be ready to merge across all feature branches and we can go on with business as usual

vonbearshark:
💼

Issue #1299, opened 2021-04-28
Admin users can include sites during org creation


JordanGuinn:
## Background
During organization creation, OEC admin users will need the ability to add sites to be included within said organization.

## Acceptance Criteria
- `Sites` section is added to the OEC admin `Create Organization` view
- Users can add multiple sites to a single organization
- Fields included are outlined in the designs below
- Sites are saved to the organization once the entire form is submitted
   - Any errors preventing sites from being saved are bubbled up to the user

## Designs
[Full designs can be found in Figma](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=97%3A728).
![ECE Reporter – Create org.png](https://images.zenhubusercontent.com/6050f59bb12f6b5f4da9aeb1/d19622b1-5885-4d96-9b34-f3a167407442)

bamader:
When creating sites, which fields do we need to enforce uniqueness on in the DB? (I.e. for organizations, we block the request to create an org if the name is taken; so for sites, which of their fields should block creation if it would duplicate something in the DB already) I'm guessing name only (because all of facility code, NAEYC ID, registry ID, and license number are optional) but thought this was worth checking.

Also, couple of questions about desired behavior in the app (do objects get created, what alerts if any should we show the user, etc.) for some different circumstances:
- What happens if one of the sites an admin enters can't be created, but other sites can (and the org name is available)?
- What happens if none of the sites an admin enters can be created, but the org name is available?

JordanGuinn:
I think just `siteName` right now is fine regarding `Site` table column uniqueness (this should already be enforced at the database level anyway).

As far as situations where site(s) can't be created for whatever reason, I think our API layer should operate as our UX does and follow an `all-or-nothing` approach (_especially_ for **creating** new orgs), meaning if we can't create anything, we create nothing.  I don't have any specific ideas around what kind of error messaging/UX we'd want in these scenarios, though.

@Derekskylight @ehuang525 @jlhogan Any thoughts around UX error handling here?

bamader:
That sounds good. And honestly, I kind of prefer the all-or-nothing approach, _especially_ for this ticket. Because we're working on base functionality now and I'd love to start parallelizing other work, I think doing an all or nothing approach would help get this ticket along. We can always fast follow with more complex behavior or different types of handling in a later ticket.

[...]

PR #1326, opened 2021-05-13
Add sites cards to create org page

bamader:
## Background
Expands the create org page of the Admin view to allow a user to create new sites and associate them with the organization.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1299

## Validation Plan
Tested that new sites are successfully created in the DB for valid fields on the create org page. Tested that failure of org name or one or more site names blocks creation, and that the correct error banners are shown to the user.

## Automated Testing


client/src/shared/models/NewSite.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+ export type NewSite = {
+  name: string;
+  titleI: boolean;
+  region: Region;
+  facilityCode?: string;
+  licenseNumber?: string;
+  registryId?: string;
+  naeycId?: string;
+}
```
vonbearshark, line 15: Do we use this pattern elsewhere? Any chance we could instead use `Site` or `Partial`/`Pick` of `Site`?

src/controllers/organizations.ts
```diff
@@ -10,12 +12,38 @@ async function doesOrgNameExist(name: string) {
...
+  // org name is taken or *any* site name is taken, fail the whole thing
+  const orgNameExists = await doesOrgNameExist(name);
+  if (orgNameExists) throw new ResourceConflictError('Name already exists.');
+  const siteNamesAllUnique = await siteController.areSiteNamesUnique(
+    sites.map((newSite) => newSite.name)
+  );
+  if (!siteNamesAllUnique) throw new ResourceConflictError('A site name is already taken.');
+
+  // Create org first so we can give its ID to created sites
+  const newOrg = await getManager().save(
```
vonbearshark, line 27: Do we need to `save`? Will `create` give us the `id` we need?

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -27,22 +48,51 @@ const CreateOrg: React.FC = () => {
...
+      licenseNumber: '',
+      registryId: '',
+      naeycId: '',
+    }]);
+  }
+  useEffect(() => {
+    if(newSites.length === 0) {
+      addNewSite();
+    }
+  }, [newSites]);
```
vonbearshark, line 74: what's happening here?

client/src/containers/CreateOrg/getCreateSiteCard.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { Region } from '../../shared/models';
+import { NewSite } from '../../shared/models';
+
+/**
+ * Function that maps the state information in a new site object
+ * the card-contained JSX field style that will show up on the page.
+ * Each card maintains its own state data and we collect it at
+ * the end.
+ */
+export const getNewSiteCard = (card: NewSite, numberOnPage: number) => {
```
vonbearshark, line 12: isn't this just a component? why have the `get...` convention versus `CreateSiteCard`?

client/src/containers/CreateOrg/getErrorMessage.tsx
```diff
@@ -0,0 +1,23 @@
+import { AlertProps } from "@ctoec/component-library";
+
+export const getErrorMessage = (err: any, newOrgName: string) => {
+  
+  // 409 Resource conflict error if an org with given name already existed
+  if (err.includes('exists')) {
```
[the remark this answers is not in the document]
bamader: Sure! Now that we have multiple 409 types that makes sense. I'll change it to `organization`  in the includes.

client/src/containers/CreateOrg/getErrorMessage.tsx
```diff
@@ -0,0 +1,23 @@
...
+    return [{
+      type: 'error',
+      heading: 'Organization already exists',
+      text: `An organization with name "${newOrgName}" already exists.`
+    }] as AlertProps[];
+  }
+
+  // Different 409 error if at least one site name is taken
+  if (err.includes('site name')) {
```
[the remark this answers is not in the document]
bamader: If one of the `if`s triggers, the clause ends with a return and the other `if` gets short-circuited, so yeah I set them up to be mutually exclusive.

client/src/containers/CreateOrg/getErrorMessage.tsx
```diff
@@ -0,0 +1,23 @@
+import { AlertProps } from "@ctoec/component-library";
+
+export const getErrorMessage = (err: any, newOrgName: string) => {
+  
+  // 409 Resource conflict error if an org with given name already existed
+  if (err.includes('exists')) {
+    return [{
+      type: 'error',
+      heading: 'Organization already exists',
+      text: `An organization with name "${newOrgName}" already exists.`
```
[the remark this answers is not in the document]
bamader: Honestly? No, it's not really needed. It's not adding any value to the error message because there's only one org being created at a time. Not like we need to remind them of what's literally already typed in the input field lol.

client/src/shared/models/NewSite.ts
bamader: Oooh using a Partial of Site is a good thought. There's only one minor annoyance, and it's that the last four variables (`facilityCode` through `naeycId`) are actually numbers in the entity schema for site. But really that just means we have to use a `parseInt` in the `onChange` functions for each of the fields in the `NewSiteCard`s. I'll give that a try.

src/controllers/organizations.ts
bamader: We do unfortunately need the `save` call. Create does not give the `id` we need (that was the first thing I tried) so we're forced into the mildly roundabout way of saving the orgs then re-saving them after we attach sites to them.

client/src/containers/CreateOrg/CreateOrg.tsx
bamader: This is just to start the `newSites` list off with one object when the page is loaded, so that there's always a site card open for the admin when they first come onto the page. I thought it was ugly to instantiate the state call with a partial site and then duplicate all those variables again in the `addNewSite` call. Not a fan? Can def switch this out.

client/src/containers/CreateOrg/getCreateSiteCard.tsx
bamader: When you say it's a component, do you mean you'd prefer to see it defined like `export const CreateSiteCard: React.FC  = () =>  {}`? That would take some parameter typing for a `Partial<Site>` and then would get instantiated with inline props, which felt uglier to me than just a function. Plus, this is the convention (the `get` that takes an object and maps it to a visual element) we used for the funding source cards on the pre-submit home screen. Open to pushback here, but if it's not a dealbreaker I'd personally stick with the function style here just cuz I like it.

Unless of course I'm completely misunderstanding what you meant, which is very possible (post-dinner me gets food induced brain fog 🧠 ). Any clarifications?

client/src/shared/models/NewSite.ts
bamader: `Partial<Site>` works smashingly so imma call this one resolved 👍

nickbristow (commented):
I'm not able to create an org when i run this locally. I'm enter random letters here so shouldn't be a conflict. 

Create_Organization

client/src/containers/CreateOrg/CreateOrg.tsx
vonbearshark: How about we create a `const emptySite`, then init `useState([emptySite])` and here `setNewSites([...newSites, emptySite]);` ?

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -27,22 +48,51 @@ const CreateOrg: React.FC = () => {
...
+      naeycId: undefined,
+    }]);
+  }
+  useEffect(() => {
+    if(newSites.length === 0) {
+      addNewSite();
+    }
+  }, [newSites]);
+
+  const siteSection = (
```
vonbearshark, line 76: any reason we separate these here?

client/src/containers/CreateOrg/getCreateSiteCard.tsx
vonbearshark: That is what I meant, and I am fine as is 👍🏻

client/src/containers/CreateOrg/CreateOrg.tsx
bamader: Not quite sure I follow what you mean here. You're saying instantiate an `empty site: Partial<Site> = { ...props go here... }`, then use that to init `useState()`? How would that address having two separate places of defining all those props? (Fairly confident I'm misunderstanding intent here though 😂 )

vonbearshark: could be _I_ misunderstand. Edited the code to maybe make it a little more clear? The "props" just refer to the empty state object? I mean that you could define a variable for that and use it for both `useState` and `addSite`, which feels clean?

bamader: OH I think I see now. Your clarification of defining one variable to use in both places helped a lot (I think we're in agreement that we don't want to write out that emptySite object with all its fields twice). If we do that, won't the code just re-use the same object each time the `addNewSite` function is called? As in, it won't _not_ be adding new objects but instead just having multiple fields that edit the same object?

vonbearshark: ug yeah I forgot our form stuff is not immutable.. Updated code to use spread, should fix it

bamader: Yeah don't you just love all our systems little...quirks 🤔 😖 ? In any case yeah what's there now makes perfect sense to me so I'll stick this in.

client/src/containers/CreateOrg/CreateOrg.tsx
bamader: Lol nope. This was an artifact from early work, where I thought the site section would be longer and more "intensive" than it ended up being. Can now just directly integrate the section into the `return( )` part without a variable.

bamader:
Okay, some things fixed in the last couple PRs (based on feedback and some pairing on Friday):
- defined bottom margin between the add sites button and the fixed bottom bar (it's a bit hacky because we have to override the USWDS styles, but it works well)
- replaced janky button with an anchor that uses button styling (🙃 )
- added a check to stop orgs with blank names from submitting
- decomposed/simplified the add sites functionality to be cleaner
- fixed submission behavior so that it brings you back to the admin's general Organizations page after creating a new org (we can eventually point this to the specific page for that org if we're so inclined, but this feels analogous to how we have the Create Record button redirect back to the overall roster after submission)
- fixed field ordering in the site card so that `TABBING` over moves by rows, not by weird column spaces

Should be ready for a solid re-review! Also, for whoever's validating or checking this on local as they're reviewing or whenever, check to see if the success alert that comes up after you create an organization flashes. It's flashing for me on my local but I don't get why; it's the same code that stably shows a revision request on the home page if a user does that, so I also don't know if it's actually an issue or just me.

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -1,48 +1,96 @@
...
+    region: (null as unknown as Region),
+    facilityCode: undefined,
+    licenseNumber: undefined,
+    registryId: undefined,
+    naeycId: undefined,
+  };
+  const [newSites, setNewSites] = useState<Partial<Site>[]>([{ ...emptySite }]);
+  const addNewSite = () => {
+    setNewSites(o => [...o, { ...emptySite }]);
+  }
```
vonbearshark, line 84: style nit nit: 

```
  const addNewSite = () => setNewSites([...newSites, { ...emptySite }]);
```

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -1,48 +1,96 @@
...
+  useEffect(() => {
+    if(newSites.length === 0) {
+      addNewSite();
+    }
+  }, [newSites]);
```
vonbearshark, line 89: Still using the `useEffect`?

client/src/containers/CreateOrg/CreateOrg.scss
```diff
@@ -0,0 +1,3 @@
+.container-with-bottom-margin {
+  margin-bottom: 4rem !important;
```
vonbearshark, line 2: hmm. Do you know where the conflict is happening? This shouldn't be necessary...

client/src/containers/Organizations/Organizations.tsx
```diff
@@ -6,14 +6,24 @@ import { OrganizationSummary } from '../../shared/payloads/OrganizationsResponse
...
+  
+  // We might have a success alert pushed from the create org page
+  // so check whether we do (but filter so that we only show one
+  // if a user created multiple orgs)
+  const [alertElements, setAlerts] = useAlerts();
+  useEffect(() => {
+    setAlerts((_alerts) => [
+      _alerts.find((a) => a?.heading === 'Organization created'),
+    ]);
+  }, []);
```
vonbearshark, line 26: Do we want this to be in a `useEffect`?

client/src/containers/CreateOrg/CreateOrg.tsx
bamader: Can probs change the name, but I learned doing the revision request form that when modifying an array with state in our  app, things behave more stably with an arrow function inside the `setState` call. I have no idea why this is true, but empirically it has been; that's why there's an `o => [...o, { ...emptySite }]`  inside the func.

client/src/containers/CreateOrg/CreateOrg.scss
bamader: Yeah. It's in the `grid-container` style. Nick, Jordan, and I paired on it last Friday and like nothing would override the grid-container's own margin, so this was the cleanest/easiest solution we came up with.

client/src/containers/Organizations/Organizations.tsx
bamader: We do, yeah. Taking away the `useEffect` leads to a glorious infinite re-rendering bug: after pressing submit, the app crashes because it keeps setting alerts over and over again.

client/src/containers/CreateOrg/CreateOrg.tsx
bamader: LOL that's a professional grade brain fart right there.
Me: adds new functionality to replace old formulation
Also me: fails to actually make the replacement 🤦‍♂️

client/src/containers/Organizations/Organizations.tsx
vonbearshark: Yep that's what I was wondering about. Makes sense kind of

client/src/containers/CreateOrg/CreateOrg.scss
vonbearshark: ugh but makes sense to me

client/src/containers/CreateOrg/CreateOrg.tsx
vonbearshark: sure fair enough

vonbearshark (approved):

client/src/containers/Organizations/Organizations.tsx
bamader: Lol I feel that. Half the component display stuff in here is a "well it kind of makes sense 🤷‍♂️ ".

client/src/containers/CreateOrg/getCreateSiteCard.tsx
```diff
@@ -0,0 +1,109 @@
...
+import React from 'react';
+import { Region, Site } from '../../shared/models';
+
+/**
+ * Function that maps the state information in a new site object
+ * the card-contained JSX field style that will show up on the page.
+ * Each card maintains its own state data and we collect it at
+ * the end.
+ */
+export const getNewSiteCard = (card: Partial<Site>, numberOnPage: number) => {
```
jlhogan, line 11: can these be components that are rendered with props in the calling components, as opposed to getter functions?

jlhogan: also, calling the site object a `card` is confusing to me. `newSite`?

src/controllers/sites.ts
```diff
@@ -22,3 +22,17 @@ export async function getSites(
...
+ * free in the DB.
+ * @param siteNames 
+ * @returns 
+ */
+export async function areSiteNamesUnique(siteNames: string[]) {
+  const foundSites = await Promise.all(siteNames.map(
+    async (siteName) => await getManager().findOne(
+      Site, { where: { siteName: siteName } }
+  )));
+  return foundSites.every((s) => !s);
```
jlhogan, line 37: shouldn't the result be empty ? can't we explicitly check for that? would be more straight forward to my brain

jlhogan (approved):
Bye bye component-library "Form" was nice knowin ya

client/src/containers/CreateOrg/getCreateSiteCard.tsx
bamader: Sure, I can change that (I originally liked the `get` basis of it but two other engineers expressing confusion is a "yeah this needs to go" in my book lol). So we want it to be `export const NewSite: React.FC<NewSiteProps> = ({ cardObject, numberOnPage }) => {}` and then rendered on the page with `<NewSite cardObject={thing} numberOnPage={n}>`.

jlhogan: yeah, it can be `<NewSiteCard .... />` or `<NewSiteFormCard ... />` if you prefer the verbosity! (u know i do :slightly_smiling_face: )

src/controllers/sites.ts
jlhogan: `!foundSites?.length` ?

bamader: Oh actually realized this doesn't work. Because we have to make a `findOne` query for each name, the `foundSites` array will be full of either matching `site` objects or `undefined`s (since each find call returns an object or an undefined). So the list will always have a length, we just need to make sure it contains _only_ undefineds.

jlhogan: wait yeah ah lookign at this more closely now, let's definitely _not_ do multiple findOne queries (which do an ugly nested select top 1... i'm pretty sure) and just do a single `find` with `IN` and ensure it returns nothing

bamader: Done!

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -1,48 +1,91 @@
...
+        const alertToSet = getErrorMessage(err);
+        if (alertToSet) setAlerts(alertToSet);
         else throw new Error(err);
       }
     );
   };
 
+  const emptySite: Partial<Site> = {
+    siteName: '',
+    titleI: (null as unknown as boolean),
```
JordanGuinn, line 74: hhhhhhhhhhhhhheh - what's going on here?  Why can't this just be `titleI: null`?

client/src/containers/CreateOrg/getErrorMessage.tsx
```diff
@@ -0,0 +1,23 @@
+import { AlertProps } from "@ctoec/component-library";
+
+export const getErrorMessage = (err: any) => {
+  
+  // 409 Resource conflict error if an org with given name already existed
+  if (err.includes('Organization')) {
+    return [{
```
JordanGuinn, line 7: nit: do we need to specifically cast these responses as `AlertProps[]`?  Can we instead specify the return value in the function signature, like:

```
export const getErrorMessage = (err: any): AlertProps[] | undefined => {
```

JordanGuinn (approved):

nickbristow:
We probably need a remove site button, if you accidentally add an extra site, you have to refresh the page and lost all your info in order to submit it. I added one to my branch for funding space.

client/src/containers/CreateOrg/CreateOrg.tsx
jlhogan: my guess is typescript expects a boolean. B/c this is partial, can't we just leave titleI out of it?

bamader: `titleI` is a field that's required by the DB to create a site (not an optional field), so I don't think we should omit it (right now we force the user to select the field before even sending the request just to avoid any problems). Typescript also does indeed get unhappy with just null cuz it does expect a boolean.

jlhogan: i dont mean omit like never send, just omit from this default "empty site" and add the value when the user adds the value.

bamader: Ah I see. Sure, I can take this approach. Will get rid of it in the partial and only add it once the user makes a selection.

JordanGuinn:
## Validation Strategy

- [ ] Create an organization with one site without issue
   - [ ] Confirm success banner is displayed
   - [ ] Confirm user is navigated back to the main admin view
- [ ] Create an organization with multiple sites without issue
   - [ ] Confirm success banner is displayed
   - [ ] Confirm user is navigated back to the main admin view
- [ ] Confirm an org with one site isn't created under the following conditions:
   - [ ] Site name is missing
   - [ ] TitleI is missing
   - [ ] Region is missing
- [ ] Confirm an org with multiple site isn't created under the following conditions:
   - [ ] Site name is missing from one site
   - [ ] TitleI is missing from one site
   - [ ] Region is missing from one site
   - [ ] Two sites share the same name
- [ ] Confirm a user can remove sites they inadvertently added during org creation
- [ ] Confirm adding a site, then removing it, then saving the org does not include that removed site

JordanGuinn:
After removing the `component-library` form component because of the mysterious element removal it was causing (?!), this is looking good.  The only thing that we found was the success message flashing briefly before disappearing on the list organizations page after creating a new organization.

Otherwise, this is good to go!

JordanGuinn:
Creating a separate ticket for the brief alert flashing on the list org page, as it's not the biggest deal, and actually seems to be a non-trivial amount of work to fix.  That being said, validation on this is complete and is ready for merge!

Issue #1313, opened 2021-05-05
Manually/programmatically add email data to Fawkes users


JordanGuinn:
## Background
While we do currently have an `email` column in the `User` table of ECE Reporter, that column isn't currently populated with any actual data in any of our environments.  The emails of all of our users currently resides solely in Winged Keys.  In order to allow for our [user search functionality from within our OEC admin view](https://app.zenhub.com/workspaces/ece-dev-board-5ff506028d35f30012e0e937/issues/ctoec/data-collection/1310), we're going to need to both manually add that data to all the users that exist in our prod-like environments, as well as update any scripts/automated code responsible for auto-creating users accordingly in our non-prod-like environments.

## Acceptance Criteria
- All users in our `prod` and `devsecure` environments have their emails in the `Users` table in Fawkes updated manually
- All user mutation scripts in the repo are updated to include email address, including:
   - All scripts `src/data/prod`
   - All scripts `src/data/fake`
   - The user creation mechanism in `src/middleware/authenticate.ts`

jlhogan:
alternatively, might it be a better idea to ignore the email column in ece reporter, and have the server make API call to wingedkeys to see if theres a user with that email? can we do that? or can we only look up by uuid? because keeping those emails in sync is an annoying and unnecessary step.... i added email column when we were talking about moving all of auth into ECE reporter and i was in there anyway so i just did it, but i dont think its the rightest way forward

JordanGuinn:
The native IdentityServer 4 `userinfo/connect` endpoint we're currently using supplies user info based on jwt and jwt alone.  I'm sure we _could_ extend the Winged Keys API to include an endpoint to look up users by endpoint, but I'm not sure we want to, namely because:

1. We don't have any precedent for a hand-rolled IdentityServer API to be consumed by a different application, so effort there is kind of an unknown
2. We know WingedKeys is eventually going to be replaced with ForgeRock

Certainly neither options are particularly ideal; I just picked this one because it seemed somewhat simpler and less throwaway.

PR #1327, opened 2021-05-14
Programmatically add emails to Fawkes users

JordanGuinn:
## Background
This PR updates all scripts and code bits responsible for creating/updating user data to now account for emails coming directly from WingedKeys, so we can keep them in Fawkes as well (given WingedKeys will eventually be replaced with ForgeRock).

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1313

## Associated PRs
https://github.com/ctoec/winged-keys/pull/74

## Validation Plan
- [x] Confirm users programmatically created on first login in non-prod environments now have emails added

## Automated Testing

## Additional Tasks
- [x] Add emails to all users in `devsecure`
- [x] Add emails to all users in `prod`


bamader (approved):
God these tokens are ugly 😭

vonbearshark (approved):

src/data/prod/users/upsert.ts
```diff
@@ -81,9 +83,27 @@ export const createUserData = async (
...
         createdCount += 1;
       } else {
-        console.log(
-          `\tUser ${row.userName} with username ${row.email} already exists`
-        );
+        let updatedUser: Partial<User> = {};
+
+        if (!!_firstName && _firstName !== user.firstName)  updatedUser.firstName = _firstName;
+        if (!!_lastName && _lastName !== user.lastName)  updatedUser.lastName = _lastName;
+        if (!!row.email && row.email !== user.email)  updatedUser.email = row.email;
```
vonbearshark, line 90: nit: can we do: `row?.email !== user.email`, etc?

src/middleware/authenticate.ts
```diff
@@ -112,7 +139,7 @@ const getUser = async (wingedKeysId: string) => {
  */
 async function getUserFromWingedKeys(
   bearerToken: string
-): Promise<AxiosResponse<any>> {
+): Promise<AxiosResponse<UserInfoResponse>> {
```
vonbearshark, line 142: 👏🏻

src/middleware/authenticate.ts
```diff
@@ -37,7 +62,9 @@ const decodeClaim = jwt({
  */
 const addUser = passAsyncError(
   async (req: Request, _: Response, next: NextFunction) => {
-    if (req.claims?.sub) {
+    let claims = (req.claims || {}) as TokenClaims;
```
vonbearshark, line 65: nit nit: `const` and `??` (instead of `||`)

vonbearshark (approved):

src/data/prod/users/upsert.ts
JordanGuinn: That would just confirm that `row` is defined, not necessarily that `email` is defined, yeah?  Can you use that conditional `?` check on primitives?

jlhogan:
so the approach here is that we're updating user data in ECE Reporter from a file? 
how are we imagining we manage updating emails in wingedkeys? it appears we can get emails in the tool from the user claim, so is there realy a need to store emails in ECE Reporter? As long as the emails are up to date in wingedkeys DB (with correct id FKs to ECE Reporter user table) replacing ID server with forgerock should still be easily achievable. 

if we can get email from ID Server in the claim, then i think storing emails in the tool just introduces an additional opportunity for things to get out of sync. But can we actually get that custom claim from ID Server?

JordanGuinn:
We can get the _currently authenticated_ user's email from the claims we decode from the JWT, but that's wildly different from getting all emails of all users in WingedKeys.  We _could_ do that, it would just be an unknown amount of unprecedented, throwaway work (because of the ForgeRock shift).

Re the email update question: I haven't put any thought towards non-manually updating emails in WingedKeys.  Is that a thing we're gonna need to provide programmatic capabilities for?  How often would that be necessary?

jlhogan:
duhhhhhhhhhhhhhhhhhhhh wow my brain. that's what i get for early AM reviews.

Re "updating" emails in wingedkeys .. i think what i mean more than that is actually this:
would it be useful to have a process that reconciles emails + ids in wingedkeys with emails + wingedkeysid in ece reporter user table? that way we just make the updates in wingedkeys when we hear them, and then run a script that updates any mismatched emails in the ECE Reporter DB. seems slightly less error prone than:
1. get a list with updated user emails
2. update emails in wingedkeys
3. run process to update emails in ece reporter

because if a user email changes it DEFINITELY needs to be changed in wingedkeys since that's how they sign on/ get new password

JordanGuinn:
Ohhhhhh, I think I follow.  In my head, I had #1 and #2 reversed, where the list of emails used to update Fawkes was only ever exported _directly_ from the WingedKeys database.  I assumed that we weren't going to allow admins to programmatically update users' emails through Fawkes _at all_, so this script would only ever need to be one-way (WingedKeys => Fawkes).

So the process for updating a user's email, should we ever need to do that, would be:

1. Manually update said email in WingedKeys
2. Export appropriate WingedKeys users table columns to file (probably worth documenting)
3. Run script to update Fawkes from said file

Which I think is effectively what you're describing as well?

JordanGuinn:
I now see the script does both WingedKeys user creates AND Fawkes user creates at the same time 😅 Will split the two, so we're able to do one without the other

src/data/prod/users/update.ts
```diff
@@ -0,0 +1,101 @@
...
+  });
+
+  for (const user of usersToUpdate) {
+    try {
+      const wingedKeysUser: BasicWingedKeysUser = wkUsersById[user.id];
+      if (!wingedKeysUser) {
+        throw new Error('This is actually impossible');
+      }
+
+      const [firstName, lastName] = wingedKeysUser.userName.split(' ');
```
jlhogan, line 37: we should not do this. only very fiew wingedkeys users have usernames that are actual username. In most cases it's the email. also, have updated some firstname/lastnames in ECE db and not in wingedkeys db. TLDR; wingedkeys is source of truth for email addresses (which should = usernames, except in a few legacy cases) and ECE reporter DB is source of truth for names.

src/data/prod/users/update.ts
```diff
@@ -0,0 +1,101 @@
...
+
+      let updatedUser: Partial<User> = {};
+
+      if (!!firstName && firstName !== user.firstName)
+        updatedUser.firstName = firstName;
+      if (!!lastName && lastName !== user.lastName)
+        updatedUser.lastName = lastName;
+      if (!!wingedKeysUser.email && wingedKeysUser.email !== user.email)
```
jlhogan, line 45: as mentioned above, imo we should only be checking/updated emails not firstname/lastname

src/data/prod/users/update.ts
JordanGuinn: Works for me - this was just a direct pull from the create script anyway

Issue #1281, opened 2021-04-16
As an OEC Admin User, I can edit another user's first/last name


Derekskylight:
## Background
As an OEC Admin User, I can **edit** another user's personal data, specifically their first and last name. 

## Acceptance Criteria
- Clicking on a user's name in the admin list view takes you to this Edit view
- Admin is able to edit user's first and last name, but **not** email
- User permission management can be ignored for now
   - To be handled in another ticket

## Designs
[Full Figma files can be found here.](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=2%3A2)

#### Edit View
![Screen Shot 2021-05-07 at 9.23.06 AM.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/7e0c9cdf-c7f1-4bfd-a9dc-fd324658b575)

jlhogan:
Currently, changing a users email (an thus their login username) requires the separate ID management system.
What can be edited here is the users name + their association with any orgs/ sites

JordanGuinn:
@Derekskylight @ehuang525  Also, now that I think about it, do we _really_ want to provide OEC admins with the ability to edit _other users'_ first and last names?  I could be off-base here, but that feels odd and kind of intrusive to me

ehuang525:
@JordanGuinn Since providers don't currently have the ability to edit their own information, I think that capability on the admin side is probably better to have than not. I don't see why admins would edit people's information unless there was a real reason to? @jlhogan any thoughts on this?

jlhogan:
somehow just seeing this^ i think the ability to edit user's names is whatever -- won't be abused by staff, probably isn't actually all the useful. More important is the permissions stuff.

Issue #1316, opened 2021-05-06
OEC admin users can access all the various roster and child detail views


JordanGuinn:
## Background
While non-admin users are currently prevented from accessing all admin views in any capacity, the reverse is not actually true currently.  While OEC admins are navigated to their default overview page by default (and don't have access to the roster navbar directly), they're _also_ able to access all provider roster views as well, if they know the direct URL.

It should be noted that both the corresponding non-admin views **and** the non-admin APIs should be appropriately gated by role.

## Repro Steps
- Log in to ECE Reporter **as an OEC admin**
- Manually attempt to navigate to the `Roster` view
   - Appending `/roster` to your URL should do it
- Observe

## Expected Behavior
- User is re-routed to the default admin overview page

## Actual Behavior
- User is able to navigate to the roster view, as well as all the detail/edit child views within

JordanGuinn:
@jlhogan @ehuang525 @Derekskylight Something we uncovered a week or two ago - turns out that admin OEC users can actually access ALL views, including the roster, file upload, child edit, etc.  So they could theoretically edit child/enrollment data on behalf of given providers, whether accidentally or intentionally.

Thoughts on whether or not this is problematic?

Derekskylight:
Ideally, it's probably best to have admin users only access specific views just to remove this possibility, but I wouldn't expect providers to be accidentally/intentionally editing another user's information. 
From what I've seen in usability sessions, users typically focus on the buttons/UX that are right in front of them and data enrollment/viewing, I didn't get the notion that they would be experimenting heavily with ECE Reporter's functionalities

ehuang525:
In the current schema of things, how would the admin user access all the roster views? Does clicking on an organization in the list view take them to the organization's roster? 

If that's the case, we just have to make sure that when the admin user clicks on an organization's page that the roster isn't the most prominent thing. There may be a way we intentionally make roster access a bit more hidden. To my knowledge, the admin's primary goals are to create and edit org info so that's what we should present to them upfront.

JordanGuinn:
If they just append `roster` to the end of their URL, they can get to the roster view without issue.  Unlikely an admin would ever _do_ that, so this likely isn't particular high priority, but did wonder in principle if we just wanted to prevent that across the board.

jlhogan:
sure, let's prevent it. 
But also, since none of the admin are associated with any org or site permissions, I don't think they actually can access data.

JordanGuinn:
[...]

PR #1333, opened 2021-05-19
Edit users name

vonbearshark:
## Background
Added user edit page for name information

## GitHub Issue
#1281 

## Validation Plan
- [x] Can edit user first, last, and middle names and suffix

## Automated Testing

## Additional Context
UI:
Screen Shot 2021-05-18 at 9 19 17 PM

### DIFF
--- a/client/src/containers/Users/EditUser.tsx
+++ b/client/src/containers/Users/EditUser.tsx
@@ -1,0 +1,154 @@
+import React, { useContext, useState, useEffect } from 'react';
+import { Redirect, useHistory, useParams } from 'react-router-dom';
+import {
+  Button,
+  Divider,
+  Form,
+  FormField,
+  FormSubmitButton,
+  LoadingWrapper,
+  TextInputProps,
+  TextInput,
+} from '@ctoec/component-library';
+import { BackButton } from '../../components/BackButton';
+import AuthenticationContext from '../../contexts/AuthenticationContext/AuthenticationContext';
+import UserContext from '../../contexts/UserContext/UserContext';
+import { useAlerts } from '../../hooks/useAlerts';
+import { getH1RefForTitle } from '../../utils/getH1RefForTitle';
+import { apiGet, apiPut } from '../../utils/api';
+import { User } from '../../shared/models/db/User';
+
+const EditUser: React.FC = () => {
+  const [alertElements, setAlerts] = useAlerts();
+  const { accessToken } = useContext(AuthenticationContext);
+  const { user: adminUser } = useContext(UserContext);
+  const history = useHistory();
+  const { id } = useParams() as { id: string };
+  const h1Ref = getH1RefForTitle();
+
+  const [user, setUser] = useState<User | null>(null);
+  const [saving, setSaving] = useState(false);
+
+  useEffect(() => {
+    (async function loadUser() {
+      if (adminUser && accessToken)
+        apiGet(`users/${id}`, accessToken)
+          .then((res: User) => setUser(res))
+          .catch((err) => {
+            console.error(err);
+            history.push({
+              pathname: '/users',
+              state: {
+                alerts: [
+                  {
+                    type: 'error',
+                    text: 'Unable to fetch user information',
+                  },
+                ],
+              },
+            });
+          });
+    })();
+  }, [adminUser, accessToken]);
+
+  return !adminUser?.isAdmin ? (
+    <Redirect to="/home" />
+  ) : (
+    <div className="grid-container">
+      <div className="grid-row grid-gap">
+        <div className="tablet:grid-col-12 margin-top-4 margin-bottom-2">
+          <BackButton text="Back to users" location="/users" />
+          {alertElements}
+          {user ? (
+            <>
+              <h1 ref={h1Ref}>{`${user.lastName}, ${user.firstName}`}</h1>
+              <h2 className="margin-top-4">User Information</h2>
+              <Divider />
+              {user.email && (
+                <>
+                  <h3>Email Address</h3>
+                  <p>{user.email}</p>
+                  <p>
+                    If you need to make changes to user emails, please{' '}
+                    <a
+                      className="usa-button usa-button--unstyled"
+                      href="/support"
+                    >
+                      send us a support request
+                    </a>
+                    .
+                  </p>
+                </>
+              )}
+              <Form<User>
+                id="change-enrollment-form"
+                className="usa-form"
+                data={user}
+                onSubmit={(_user: User) => {
+                  setSaving(true);
+                  apiPut(`users/${user.id}`, _user, { accessToken })
+                    .catch((err) => {
+                      console.error(err);
+                      setAlerts([
+                        {
+                          type: 'error',
+                          text: 'Unable to save user information',
+                        },
+                      ]);
+                    })
+                    .finally(() => setSaving(false));
+                }}
+              >
+                <FormField<User, TextInputProps, string | null>
+                  getValue={(data) => data.at('firstName')}
+                  parseOnChangeEvent={(e) => e.target.value}
+                  inputComponent={TextInput}
+                  type="input"
+                  id="firstName"
+                  label={<h3>First name</h3>}
+                />
+                {user.middleName && (
+                  <FormField<User, TextInputProps, string | null>
+                    getValue={(data) => data.at('middleName')}
+                    parseOnChangeEvent={(e) => e.target.value}
+                    inputComponent={TextInput}
+                    type="input"
+                    id="middleName"
+                    label={<h3>Middle name</h3>}
+                  />
+                )}
+                <FormField<User, TextInputProps, string | null>
+                  getValue={(data) => data.at('lastName')}
+                  parseOnChangeEvent={(e) => e.target.value}
+                  inputComponent={TextInput}
+                  type="input"
+                  id="lastName"
+                  label={<h3>Last name</h3>}
+                />
+                {user.suffix && (
+                  <FormField<User, TextInputProps, string | null>
+                    getValue={(data) => data.at('suffix')}
+                    parseOnChangeEvent={(e) => e.target.value}
+                    inputComponent={TextInput}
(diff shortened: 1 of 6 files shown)


client/src/containers/Users/EditUser.tsx
```diff
@@ -0,0 +1,154 @@
...
+                )}
+                <FormSubmitButton
+                  text={saving ? 'Saving...' : 'Save changes'}
+                  disabled={saving}
+                />
+                <Button text="Cancel" appearance="outline" />
+              </Form>
+            </>
+          ) : (
+            <LoadingWrapper loading={true} />
```
bamader, line 146: I think you can just put the `LoadingWrapper` around the whole block of code (starting with where you had the ternary for `user`) and instead of having the props `loading={true}` you can use `loading={!!user}` or something equivalent.

bamader (approved):
This looks good to me (aside from a somewhat/somewhat not nit around loading wrapper)! I'm in favor of keeping the suffix and middle name fields, since those seem entirely innocuous from a design perspective and are valid things for an admin to want to edit.

client/src/containers/Users/EditUser.tsx
vonbearshark: Yeah, I hear this. I went back and forth. Even if you put `loading={!!user}` you still need to check for `{user && (<>/* etc... */</>)}` within the `LoadingWrapper`. To my eye it makes it seem like there's a state where the loading wrapper isn't loading and also there's no `user`--and what would be shown there? just nothing.. That state doesn't actually exist, practically (I don't think, anyway), so it just feels confusing or clunky.

I saw this pattern elsewhere and went with it because it helps illustrate the state of: either there's a user defined or we're loading to wait for one.

But that's just my eye

bamader: This is totally fine with me then. If we use this pattern elsewhere and it makes more sense to you to separate states out like that, I've got no problems with it. It probably is cleaner in the body of the code anyway.

JordanGuinn (approved):
Looks great to me!

vonbearshark:
Just deployed to e2e and QA if someone wants to jump on validation? @bamader @JordanGuinn ?

client/src/containers/Users/EditUser.tsx
```diff
@@ -0,0 +1,154 @@
...
+                    text: 'Unable to fetch user information',
+                  },
+                ],
+              },
+            });
+          });
+    })();
+  }, [adminUser, accessToken]);
+
+  return !adminUser?.isAdmin ? (
```
jlhogan, line 54: at this point should we wrap all components in some big switch that does this redirect always? not exactly sure how that's implemented, altho i am sure there's some accepted solution out there. And at this point, we shouldn't need to be writing this on every single page b/c that also leaves us open to https://github.com/ctoec/data-collection/issues/1316

client/src/containers/Users/EditUser.tsx
```diff
@@ -0,0 +1,154 @@
...
+                        {
+                          type: 'error',
+                          text: 'Unable to save user information',
+                        },
+                      ]);
+                    })
+                    .finally(() => setSaving(false));
+                }}
+              >
+                <FormField<User, TextInputProps, string | null>
```
jlhogan, line 102: FWIW i would prefer to drop the `FormField` and just use the `mutableUpdateData` or whatever its called from the form context in the normal input component `onChange`

jlhogan (approved):

client/src/containers/Users/EditUser.tsx
vonbearshark: Yeah I thought of something similar. Following convention here, and would prefer this didn't block, but I could definitely see either 1. a boolean field in the router like `admin` (which is what I prefer, I think), or 2. a top-level `<AdminRedirect>` component we could drop in

jlhogan: agreed, not a blocker. will open a ticket to get sprinted for next sprint. 
if we can add flags to the router, let's do that but i'm not sure we can.

vonbearshark: Created a ticket to capture that here #1347

client/src/containers/Users/EditUser.tsx
vonbearshark: Yeah I wasn't sure the exact relationship of the various related `Form` components. That does make sense to me as well. I am going to keep it here in this case since I see this pattern elsewhere and especially since we will eventually replace all `Form` stuff anyway

JordanGuinn:
Looks like a bunch of e2e failed?

vonbearshark:
Yeah it looks like it's from issues I addressed in base awhile back. I mean, it doesn't make sense that any e2e tests would actually fail, right? Like this isn't covered at all and I haven't tweaked anything generic. I think it has to do with the epic branch being off..

JordanGuinn:
It... indeed does not.  Gonna run em again to confirm, then will check things out on `base`

vonbearshark:
`base` is probably fine but `EPIC-oec-admin-view` probably isn't, I mean

JordanGuinn:
Welp, ran em again and they passed!

src/controllers/users.ts
```diff
@@ -41,3 +41,18 @@ export const getUsers = async (): Promise<User[]> =>
...
+  const foundUser = await getManager().findOne(User, id);
+  if (!foundUser) throw new NotFoundError();
+  return foundUser;
+};
+
+export const updateUserName = async (id: string, updatedUser: User) => {
+  const foundUser = await getUserById(id);
+  foundUser.firstName = updatedUser.firstName;
+  foundUser.middleName = updatedUser.middleName;
+  foundUser.lastName = updatedUser.firstName;
```
JordanGuinn, line 55: heh, `lastName` should probably be set to `lastName`, yeah? 🙃

JordanGuinn:
## Validation Plan
- [ ] Confirm I can't access the edit user view as a non-admin user
- [ ] Confirm I can access the edit user view as an admin user
- [ ] Confirm I can update first name
- [ ] Confirm I can update middle name
- [ ] Confirm I can update last name
- [ ] Confirm I can update suffix

JordanGuinn:
Functionally, the only issue I found was the last name is always being set to whatever is specified in the form for first name.  Have a number of design questions though, which I'll capture below.

JordanGuinn:
## Design Findings
- There's no success alert or other indication informing the user that the user has been successfully updated
- `Middle Name` and `Suffix` fields can only be edited if there's already a value present for those fields
   - Which is to say, if you remove a user's middle name, there's no way to add it back
- All form text is the same size as the `User Information` form header text (it's all smaller in designs)
- Nit: `Back to Users` button different from `Users -> Username` breadcrumb path in designs
- Nit: There's no light gray box surrounding the support request text, as in designs

cc @ehuang525, if you have thoughts on any of these (but especially the middle name and suffix bits)

ehuang525:
Design comments:

- Good call out. Upon successfully saving changes, let's have a success message appear at the top of the page that says "Your changes have been saved." Does that seem reasonable for now? I'm happy to provide a mockup but think that we can move forward without one for the time being since we have that success message component available. Will keep these scenarios in mind for other tickets!
- As we discussed over Slack, the middle name and suffix situation should be fine for now. I don't imagine these fields being the primary identifiers for a user anyway.
- Could we reduce the field title size to the same size as what's inside the text field?
- 'Back to users' is fine. If we go with this direction I'd rather us use this approach rather than breadcrumbs for all the other admin designs as well so there is consistency. 
- This is also fine for now. Whenever we have more capacity to do a full styling sweep through the site that might be a better time to standardize visual details.

Thanks Jordan!

JordanGuinn:
WOOT, sounds like a plan - thanks for the super quick turnaround on this!

vonbearshark:
Than you _both_!

1. It makes sense to me to remove the restriction on middle/suffix and have those fields regardless of if they're present. Does that work?
2. re: the gray box: is that a style we already have that you know of?
3. Keeping the back button versus the breadcrumbs? I am partial to this, but no strong opinions at all

Otherwise will address all the issues

JordanGuinn:
1. Yep!
2. I don't think so - we can leave as-is for now
3. Sounds like `Back to Users` is fine

client/src/containers/Users/EditUser.tsx
```diff
@@ -0,0 +1,180 @@
...
+                  label={<span className="text-bold">First name</span>}
+                  status={(data) =>
+                    !data.at('firstName').value
+                      ? {
+                          type: 'error',
+                          id: `status-firstName`,
+                          message: 'First name cannot be empty',
+                        }
+                      : undefined
+                  }
```
vonbearshark, line 128: This seems to work, but it doesn't actually block submission... As you said @JordanGuinn  I don't think this is a necessary fix, but if there was an easy drop-in for making sure all `status`es were okay before submission, I am all ears. (Seems like something that should already be happening...)

client/src/containers/Users/EditUser.tsx
```diff
@@ -0,0 +1,180 @@
...
+                  label={<span className="text-bold">Suffix</span>}
+                />
+                <FormSubmitButton
+                  text={saving ? 'Saving...' : 'Save changes'}
+                  disabled={saving}
+                />
+                <Button
+                  text="Cancel"
+                  appearance="outline"
+                  onClick={() => setCancelToggle(!cancelToggle)}
```
vonbearshark, line 167: Originally I had `=> setUser(user)`, but I straight up could _not_ trigger the `Form` re-render/update. State could revert, but it just won't propagate, given our `Form` architecture. I am open to other thoughts, but this toggle `key` hack _does_ work

vonbearshark (commented):
Should have addressed all feedback. Left comments for tricky bits in the fixes, but mostly good for another review/validation (haven't deployed it myself)

JordanGuinn:
Confirmed all functional and design issues have been addressed, and all previous manual tests are still passing.  I'd say this is ready for merge!

Issue #1310, opened 2021-05-05
Admins can search for users by email from the org create view


JordanGuinn:
## Background
In order for OEC admins to be able to provide users access to the organizations they create, they first need a way to look users up (by **email** specifically).  Let's add a UX that allows for admins to do just that.

Of course, this ticket presupposes that there is actual email data to reference in the `User` table of ECE Reporter, which will be addressed in https://app.zenhub.com/workspaces/ece-dev-board-5ff506028d35f30012e0e937/issues/ctoec/data-collection/1313.

## Acceptance Criteria
- New `Users and permissions` section is added to the admin org create view, matching the designs specified below
- Admin users can look up users by email in the org create view
   - **Admin only** user search API is written
   - Search API requires an **exact** match to return applicable results
- The search response includes a **full count** of all user matches, as well as the **user emails** themselves
- Users can **NOT** yet _add_ users to organizations
   - To be implemented in a later ticket

## Designs
[Comprehensive FIGMA designs can be found here.](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=2%3A2)

#### Search
![Screen Shot 2021-05-05 at 1.01.23 PM.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/8f791509-9922-4f60-9a0e-9dc4e6c0486b)

#### Search Response
![Screen Shot 2021-05-05 at 1.01.34 PM.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/e22bb9f5-68bb-4302-bfe5-ffa8b0399d6a)

JordanGuinn:
@ehuang525 @Derekskylight Do we want the user search UX in the create organization view to search by **exact** email matches, or **partial** email matches?

Derekskylight:
I don't think we've come to a consensus on this topic, but I would argue exact search is the way to go:
1. It allows for an easier lift in dev (checking if strings are equal to one another as opposed to seeing if a string is contained within another string)
[...]

PR #1341, opened 2021-05-19
Add user search to create org page

bamader:
## Background
This PR provides a search function for admin users when viewing the create organization page. Users can be searched by email (only exact matches are returned), and any found results are displayed with Add User buttons beside them. These buttons currently do _not_ add any permissions, as that will come in a later ticket.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1310

## Related PRs
https://github.com/ctoec/component-library/pull/153

## Validation Plan
Created multiple users in my local database and performed searches on them to ensure they successfully appeared. Also tested the UI of when search results should show.

## Automated Testing

## Additional Context
This PR isn't technically blocked by https://github.com/ctoec/data-collection/pull/1326, but since it builds off of a lot of the things already there, that other should be merged in first (it's done with reviews and ready for someone to validate it--I think it's also semi-blocking one of Nick's PRs because I think his funding spaces section work is also building onto the site cards).


vonbearshark:
This is our first search component, right? Do we want to try something like [carbon's search component](https://www.carbondesignsystem.com/components/search/code/)?

bamader:
Mucked around with the search component a bit and I don't actually think it does what we want it to. Their docs aren't very clear about this (or maybe I'm just a lil dense cuz it took me a few read-throughs to comprehend this), but carbon search is about _on-page_ navigation rather than _in-app_ navigation. The examples they give are all for finding and navigating content on a single page (for `global` search) or within a single, space confined component (for `small` search). It's not actually supposed to be a search for querying pieces of data. I additionally realized this when I tried out the only two potential props that might still allow data search functionality, `onChange` and `onKeyDown`. Both activate every time the value in the search bar changes (i.e. a user will get the 0 results message as soon as they start typing, rather than waiting until they hit enter), leading me to conclude this is basically limited to on-page CMD + F. Which is kind of a letdown, but I think a single `TextInput` and a button are still the way to go.

TLDR: Carbon search is for cmd + f on a single page, not for data querying, and it doesn't have props or functionality to support using it like a search engine. Would recommend not using it for this.

vonbearshark:
Ugg. My bad but appreciate the investigation. What do you think about trying to make a `Search` component in the component library at any rate? Even if it's just the existing code

bamader:
No worries! Yeah, we could do that, but I think we'd have to answer some design questions around the component first (just so I don't make executive decisions about what goes where and then we want to change that). Some initial thoughts/Q's that come to my mind:

- How minimalist do we want the component? One easy thing to do would be to have the component just be the TextInput and the Search Button, and then have the caller supply everything else that's needed (the state data that holds the text, the query function that happens when a user hits enter, the state where results will be stored after they're found, etc.). That would get this off the ground quickly, albeit maybe not as "React-y". I think regardless, we'll need to have the caller supply the "onSearch" function since we'll be querying different routes with different uses of a Search component.
- Another option is to have the Search component manage its own text field state. The caller may or may not need to know what the value of the state is,  but if the search component is making the api request to actually _do_ the search on its own, then it may be less of an issue. 
- Handling the search results is a bit more tricky. The way it is now, the api request just directly sets some state on the page, which makes displaying that very easy. If we try to push the result handling and result display into the Search component, that will definitely clean up the caller's code, but then we have to have the component manage those things (although now that I think about it, maybe we could just use the same state toggles for showing/hiding results).
- But, having the Search component handle result showing means we're essentially standardizing the styles that results display with, unless we also do some fancy `cx` styling magic (possible, but will likely require some time just cuz it's not my forte)
- Final question is how to handle when we want to display more things per result than just the result itself. In the current use case, we want to show the email _plus_ an Add User button for each result we find. If other searches return a table or some different configuration of elements, I'm not entirely show how to set that up stylistically so that props passing can make it happen. I'm sure there's a way (maybe having a prop called `perResultDisplay` or something like that--though I don't necessarily know how I'd get all the `grid-row grid-gap` and `grid-col:tablet` splits in there), but one doesn't come to me readily (which may just be a natural consequence of me not having any previous React experience).

Happy to chat about this more in depth whenever! It's definitely something we should thoroughly explore.

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -113,10 +134,59 @@ const CreateOrg: React.FC = () => {
...
+        <Button
+          className="margin-top-2 margin-bottom-4"
+          text="Search"
+          onClick={searchForUsers}
+        />
+        {showSearchResults && (
+          <>
+            <p className="margin-bottom-2 text-bold">
+              {
+                `We found ${foundUsers.length === 0
```
JordanGuinn, line 165: Do we need this ternary here?  I think `pluralize` handles 0 properly just like any other number

JordanGuinn (commented):
LGTM overall 👍🏽👍🏽👍🏽👍🏽 Just a few things

JordanGuinn:
Also, I still don't think I understand why we can't just use the Carbon `Search` component with a `Button` - something like:

```
<Search
  id='new-org-user-search'
  placeHolderText="Search for existing user email"
  onChange={(e: any) => {
      setNewOrgName(e.target.value);
      return e.target.value;
   }}
/>
<Button
   className="margin-top-2 margin-bottom-4"
   text="Search"
   onClick={searchForUsers}
/>
```

From what I can tell, the Carbon `Search` component doesn't appear to have any specific use case, and we only need to use the `onChange` event handler to update the search text (since the actual search itself is triggered by the button).

JordanGuinn:
ALSO also, assuming that the above is actually possible, I do think that this search box/button combination  _would_ be a good candidate for something to add to `component-library`, as it's a combination of multiple components and will _for sure_ be used later (at least in the org search UI on the user admin page).

I'd vote to import this kinda thing directly from Carbon if it was _just_ the Search component, or we didn't have immediate plans of using this anywhere else.

src/controllers/users.ts
```diff
@@ -41,3 +41,13 @@ export const getUsers = async (): Promise<User[]> =>
...
     GROUP BY u.id, u.lastName, u.firstName, u.email`
   );
+
+  /**
+   * Finds all users in the DB whose email is an exact match
+   * for the provided string.
+   */
+export const getUsersByEmail = async (email: string) => {
+  const users = await getManager().find(User, { where: { email: email } });
+  const usersWithValidEmails = (users || []).filter((u) => u.email && u.email !== "");
```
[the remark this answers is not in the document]
bamader: Haha it's def not necessary

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -69,6 +76,20 @@ const CreateOrg: React.FC = () => {
...
+  const searchForUsers = async () => {
+    if (lookupUser === "") return;
+    setShowSearchResults(false);
+    await apiGet(`/users/by-email/${lookupUser}`, accessToken)
+      .then((res) => {
+        setFoundUsers(res);
+        setShowSearchResults(true);
+      })
+      .catch((err) => {
+        console.error(err);
```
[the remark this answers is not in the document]
bamader: Roger that. I'll throw the error instead, like we do for other failed api requests.

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -113,10 +134,59 @@ const CreateOrg: React.FC = () => {
...
+        <TextInput
+          label="Search for user to add to organization"
+          id='new-org-user-search'
+          type="input"
+          onChange={(e: any) => {
+            setLookupUser(e.target.value);
+            return e.target.value;
+          }}
+        />
+        <Button
```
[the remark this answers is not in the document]
bamader: Why yes we can, and that probably looks better with our UI

client/src/containers/CreateOrg/CreateOrg.tsx
bamader: Huh. It totally does. First time I tried it with 0 and pluralize, I got `We found 0 user`. Fixing this now.

bamader:
As requested, I created a search bar component in the component lib which I'll import back into this once that PR gets approved.

src/controllers/users.ts
```diff
@@ -41,3 +41,12 @@ export const getUsers = async (): Promise<User[]> =>
     JOIN organization o on op.organizationId = o.id
     GROUP BY u.id, u.lastName, u.firstName, u.email`
   );
+
+  /**
+   * Finds all users in the DB whose email is an exact match
+   * for the provided string.
+   */
+export const getUsersByEmail = async (email: string) => {
+  const users = await getManager().find(User, { where: { email: email } });
```
jlhogan, line 50: nit: you should be able to just do:
`{ where: { email } }`

src/routes/users.ts
```diff
@@ -29,11 +29,25 @@ usersRouter.get(
       throw new ForbiddenError();
     }
     const users = await controller.getUsers();
-
     res.send(users);
   })
 );
 
+usersRouter.get(
+  '/by-email/:email',
```
jlhogan, line 37: this feels weird to me (as it implies we'd build a separate route for looking up a user by whatever field, which feels like it should be a QS search param), but as i think more about it i dont see any other field we might search by so tldr i guess this is fine and does make sense, just ... feels weird

jlhogan (approved):

src/routes/users.ts
JordanGuinn: Honestly this could all just be baked into the main `GET /users` route, with the specification of an `email` query parameter trigging this particular behavior.  It's generally how I've seen filterable endpoints like this built.

vonbearshark (approved):

JordanGuinn (approved):

JordanGuinn:
@bamader and I have completed manual validation without issue - once e2e is green I'd say this is ready for merge

bamader:
We're all green here!

Issue #1340, opened 2021-05-19
Create new placeholder roster page


JordanGuinn:
## Background
This ticket effectively marks the beginning of the roster redesign work!

First things first - we should do away with the previous roster page entirely, and create a brand new placeholder, containing nothing more than the **organization name**, **site count** and **child count**, both total and broken down by age group (all captured in the designs below).

## Acceptance Criteria
- New roster page is created, **replacing** the current roster view
   - Current roster view is removed
- Roster should **only** include what's captured in the screenshot below
   1. User's **organization** as the header (multi-user support to come later)
   2. Subheader of org **child count** and org **site count**
   3. **Child counts across age groups** in cards (we already have a UI component for this kinda thing)

## Designs
[The full set of designs can be found in Figma.](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=86%3A348)

#### Initial Roster View
![Screen Shot 2021-05-19 at 11.37.55 AM.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/4765bd52-48e5-4095-b02c-7bf02f700e08)

bamader:
Documenting here a conversation I had with Jordan: the goal is to create a _complete_ overhaul/redesign of the roster, so as part of this ticket work I'll be making all the existing Roster stuff called `RosterOld`. That way we can keep it around until we're ready to delete the whole thing, but we're not tempted to pull from the code that's already there (the idea after all is to massively simplify what's going on).

PR #1360, opened 2021-05-28
Create placeholder roster page

bamader:
## Background
This PR kicks off the roster redesign work by making a new placeholder roster page to take the place of the beast that's currently there.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1340

## Validation Plan
Log into the app (should automatically land you on the roster page with Nick's PR). From there:
 - [x] Cards should currently be empty aside from the age groups
 - [x] User's organization is displayed as the page title
 - [x] Should be subheading text displaying that the user has 0 children

Upload a file using the upload wizard. Then, navigate back to the roster page.
 - [x] Cards should display counts of number children in each age group
 - [x] Sum of numbers in cards should equal the number reported in the subheader of "X children across Y sites"

## Automated Testing

## Additional Context
Per the attached Zenhub ticket, Jordan and I discussed just renaming all the old roster stuff RosterOld. So the diff in this PR is gonna look _wildly_ worse than it actually is. That's because there's a ton of stuff that's just being re-named so that we have it around, but are less tempted to pull from it (the idea being that we really are trying to rewrite the roster from the ground up to keep it simple).


client/src/routes.ts
```diff
@@ -10,6 +10,7 @@ import SubmitSuccess from './containers/SubmitSuccess/SubmitSuccess';
 import AddRecord from './containers/CreateRecord/CreateRecord';
 import FundingSourceTimes from './containers/FundingSourceTimes/FundingSourceTimes';
 import Roster from './containers/Roster/Roster';
+// import Roster from './containers/RosterOld/Roster';
```
vonbearshark, line 13: Just remove? Or at least `TODO` this

client/src/shared/payloads/AgeGroupCount.ts
```diff
@@ -0,0 +1,5 @@
+export type AgeGroupCount = {
+  'Infant/toddler': number,
```
vonbearshark, line 2: Are these formal enums somewhere already? Could this type be:

```
{
    [ageEnum]: number
}
```
or something?

src/routes/children.ts
```diff
@@ -37,6 +38,11 @@ childrenRouter.get(
         take,
       }
     );
+    if (count) {
+      const ageGroupCounts = await controller.getChildrenCountByAgeGroup(req.user, organizationIds);
+      res.send(ageGroupCounts);
+      return;
+    }
```
vonbearshark, line 45: Lets not mix counts and children records in this route. Worst-case the `count` should be its own endpoint. But I think the existing `metadata` endpoint is kind of what we're looking for here, maybe?

src/controllers/children/get.ts
```diff
@@ -180,6 +182,21 @@ export const getActiveChildrenCount = async (user: User, organizationIds) => {
   return await qb.getCount();
 };
 
+export const getChildrenCountByAgeGroup = async (user: User, organizationIds) => {
```
vonbearshark, line 185: I think that for this general roster refactor effort we should stay away from this in-memory calculation. I know it's a lift, but let's do it in SQL from the start. Especially since in this case depending on `activeChildrenQuery` will fall apart with actual the actual pagination happening

client/src/containers/Roster/Roster.tsx
```diff
@@ -1,233 +1,95 @@
...
-      childrenWithErrorsAlert,
-    ]);
-  }, [
-    isSubmitted,
-    activeErrorsCount,
-    withdrawnErrorsCount,
-    alertType,
-    activeOrgId,
-  ]);
+  const userOrg = (user?.organizations || [])[0]
```
vonbearshark, line 15: Seems like something we can use the `RosterContext` for

client/src/containers/Roster/Roster.tsx
```diff
@@ -1,233 +1,95 @@
...
-            <p>
-              The Roster allows you to view and update the enrollment data you
-              submitted previously. Using the Roster, you can:
+    <div className="margin-top-4 grid-container">
+      <h1 ref={h1ref}>{userOrg.providerName}</h1>
+      <p>
+        {`Your roster has 
+          ${
+            Object.values(childCounts).reduce(
+              (accumulator: number, value: any) => accumulator += value, 0)
```
vonbearshark, line 60: nit: if the names are just going to be `accumulator` and `value`, nothing worse about `a` and `v`, imo

client/src/containers/Roster/Roster.tsx
```diff
@@ -1,233 +1,95 @@
...
-    } else if (activeOrgId) {
-      await apiPut(`oec-report/${activeOrgId}`, undefined, {
-        accessToken,
-      }).then(() => {
-        setSubmittedModalOpen(true);
-        setIsSubmitted(true);
-      });
+  const [childCounts, setChildCounts] = useState<AgeGroupCount>(
+    Object.values(AgeGroup).reduce((a, v) => {return {...a, [v]: 0}}, {}) as AgeGroupCount
+  );
```
vonbearshark, line 20: nit:

```
const [childCounts, setChildCounts] = useState(
    Object.values(AgeGroup).reduce((a, v) => ({...a, [v]: 0}), {}) as AgeGroupCount
)
```
?

client/src/containers/Roster/Roster.tsx
```diff
@@ -1,233 +1,95 @@
...
+          }
+        children enrolled across ${pluralize('sites', sites.length, true)}.`
+        }
+      </p>
+      <div className="grid-row three-column-layout">
+        <div className="desktop:grid-col-4 three-column-card">
+          <Card className="font-body-lg">
+            <p className="margin-top-0 margin-bottom-0">Infant/toddler</p>
+            <p className="text-bold margin-top-0 margin-bottom-0">
+              {childCounts['Infant/toddler']}
```
vonbearshark, line 70: Can we dynamically create the cards/categories from the age group enums?

vonbearshark (commented):
very exciting

client/src/containers/Roster/Roster.tsx
jlhogan: i am personally a big fan of smaller reusable components to call within components, so we don't ahve to re-write all this card set up multiple times. a la https://github.com/ctoec/data-collection/blob/base/client/src/containers/Overview/SummaryCard.tsx

jlhogan: but also maybe this was JJ's point, which keys should be the enum values, and access should then look like:
```
childCounts[AgeGroup.InfantToddler]
```

src/controllers/children/get.ts
jlhogan: iim on board with that. I'd be on board with just adding an age group filter to the where query and doing the `count` version of the query, 3x (as opposed to writing a custom query that grabs each one. unless we find it's super not performant, and then i guess we write some sql)

src/routes/children.ts
jlhogan: IMO this is the only blocker in this PR

client/src/routes.ts
bamader: Whoops definitely can throw a `TODO` on this

client/src/shared/payloads/AgeGroupCount.ts
bamader: Yeah they are, they're defined for the `AgeGroup` enum. I want the value of `AgeGroup` to be the key of `AgeGroupCount`--is that what your syntax suggestion does?

src/controllers/children/get.ts
bamader: This was gonna be exactly what I did as a first pass solution, just run one query for each age group (since we've already got this well-built query).

client/src/containers/Roster/Roster.tsx
bamader: Good call. This was the first `reduce` call I'd written from scratch so the names were frankly more to just make sure I knew what each thing was lol

client/src/containers/Roster/Roster.tsx
bamader: Yeah I think different things are being said, but they're both valuable. I should refactor out the cards into components the same we did for the upload wizard, and I should make the accessor based off the enum instead of static magic strings.

client/src/shared/payloads/AgeGroupCount.ts
vonbearshark: yep!

src/controllers/children/get.ts
vonbearshark: I'm good with that

client/src/containers/Roster/Roster.tsx
bamader: I wasn't sure if we still needed the roster context. I guess I'd imagine we eventually want the same functionality as the current roster page, but also I wasn't sure if this context was something that we'd keep or try to re-write / write around / write without. Can move the org name retrieval there if others feel strongly, but I don't feel super compelled to put it there / incorporate the context right now.

vonbearshark: Yeah that's fair enough. We might want to rewrite `RosterConext` anyway

vonbearshark (approved):

client/src/containers/Roster/Roster.tsx
```diff
@@ -1,228 +1,83 @@
...
+    if (user && userOrg && accessToken) {
+      (async function getRosterSubHeader() {
+        try {
+          const counts = await apiGet(`children/count?${stringify({
+            organizationId: userOrg.id,
+          })}`, accessToken);
+          setChildCounts(counts);
+        } catch (error) {
+          console.error(error);
+          throw new Error(error);
```
JordanGuinn, line 32: We probably don't need to catch this error if we're just gonna re-throw it anyway, yeah?

client/src/containers/Roster/Roster.tsx
```diff
@@ -1,228 +1,83 @@
...
+  const [sites, setSites] = useState<Site[]>([]);
+  useEffect(() => {
+    if (user && userOrg && accessToken) {
+      (async function getSites() {
+        try {
+          const sites = await apiGet('sites', accessToken);
+          setSites(sites);
+        } catch (error) {
+          console.error(error);
+          throw new Error(error);
```
JordanGuinn, line 47: same here

client/src/containers/Roster/Roster.tsx
```diff
@@ -1,228 +1,83 @@
...
+      <p>
+        {`Your roster has 
+          ${
+            Object.values(childCounts).reduce(
+              (a: number, v: any) => a += v, 0)
+          }
+        children enrolled across ${pluralize('sites', sites.length, true)}.`
+        }
+      </p>
+      <div className="grid-row three-column-layout">
```
JordanGuinn, line 65: 😎😎

client/src/containers/Roster/RosterSummaryCard.tsx
```diff
@@ -0,0 +1,23 @@
+import { Card } from '@ctoec/component-library';
+import React from 'react';
+
+type RosterSummaryCardProps = {
+  ageGroup: string;
```
JordanGuinn, line 5: nit: can we make this parameter type `AgeGroup` instead of type `string`?

client/src/index.scss
```diff
@@ -3,7 +3,7 @@
 @import './containers/Landing/Landing.scss';
 @import './containers/Help/Help.scss';
 @import './containers/EditRecord/EditRecord.scss';
-@import './containers/Roster/Roster.scss';
+@import './containers/RosterOld/Roster.scss';
```
JordanGuinn, line 6: Do we still actually want to import this old CSS at this point?

src/controllers/children/get.ts
```diff
@@ -180,6 +182,25 @@ export const getActiveChildrenCount = async (user: User, organizationIds) => {
   return await qb.getCount();
 };
 
+/**
+ * Counts the number of children in each age group enrolled at sites
+ * associated with all of a user's organizations. No records are
+ * returned, only the counts within each age group.
+ */
+export const getChildrenCountByAgeGroup = async (user: User, organizationIds) => {
```
JordanGuinn, line 190: nit: can we add a return type of `Promise<AgeGroupCount>` to this function signature?

JordanGuinn (approved):
LGTM!  Just a handful of minor things

client/src/containers/Roster/Roster.tsx
bamader: So just no try/catch block here? I've seen us this (granted, weak) pattern before, but I have no strong preference to keep it.

client/src/index.scss
bamader: That is...an excellent observation. Nope, don't think we need it. I'll comment it out with a TODO for deletion like I did with the stuff JJ caught.

client/src/containers/Roster/Roster.tsx
JordanGuinn: Yep!  The whole premise of a catch block is to handle exceptions without breaking the flow of execution.  The fact that we're re-throwing the exact same error here indicates that we don't really know how to handle it, in which case we should just not have a catch block.

JordanGuinn:
## Validation Plan
- [ ] Confirm site count total is properly displayed when no children are enrolled
- [ ] Confirm child count total is properly displayed when no children are enrolled
- [ ] Confirm site count total is properly displayed when children are enrolled
- [ ] Confirm child count total is properly displayed when children children are enrolled
- [ ] Confirm child count by age group cards are properly displayed

Issue #1367, opened 2021-06-03
Roster export does not include SASID/unique ID


jlhogan:
## Background
For whatever reason, the `SASID / unique identifier` column in rosters generated when clicking `Export` from the roster view never _actually_ contains applicable SASID data, regardless of whether or not the corresponding child actually has a SASID.  We should fix this.

**_NOTE:_** This is relatively high priority as is blocking user from being able to confirm their ECE Reporter roster is correct (via comparison with system of record).

## Repro Steps
- Log in to ECE Reporter
- Navigate to a child detail view
- Navigate to the `Child Identifiers` section
- Add a ten digit number to the `SASID` field
- Save your changes
- Navigate back to the roster view
- Click `Export`
- Observe the downloaded roster

## Expected Behavior
- Exported roster includes the new SASID value for the updated child in the `SASID / unique identifier`

## Actual Behavior
- Exported roster has nothing but blank values in the `SASID / unique identifier` column

jlhogan:
@JordanGuinn would be very good to pull this into current sprint if possible

JordanGuinn:
For sure, can do

PR #1368, opened 2021-06-04
Export unique IDs

bamader:
## Background
The export function currently does not export SASIDs or Unique IDs saved on children to the resulting spreadsheet. This PR fixes that.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1367

## Validation Plan
- Log into ECE Reporter
- Upload a child record
- Add or edit the SASID information for that child in the roster, and save the changes
- Export the roster
- Check the resulting spreadsheet for the successfully output SASID/UniqueId

## Automated Testing

## Additional Context
The issue was happening because export works using a `hasOwnProperty` call. Our template has a `sasidUniqueId` field that jointly bundles these two, but `child` objects have only `sasid` and `uniqueId` fields, so the `hasOwnProperty` check was returning false. We just needed a special case to handle a template property that doesn't map to any entity.

### DIFF
--- a/src/controllers/export.ts
+++ b/src/controllers/export.ts
@@ -1,5 +1,5 @@
 import { BookType } from 'xlsx';
-import { ColumnMetadata } from '../../client/src/shared/models';
+import { ColumnMetadata, UniqueIdType } from '../../client/src/shared/models';
 import { Child, Enrollment, Funding } from '../entity';
 import { Response } from 'express';
 import { isMoment } from 'moment';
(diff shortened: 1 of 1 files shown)


src/controllers/export.ts
```diff
@@ -102,9 +102,25 @@ function flattenChild(
...
+            if (child.organization?.uniqueIdType === UniqueIdType.SASID) {
+              childString.push(formatProperty(
+                child[UniqueIdType.SASID.toLowerCase()], UniqueIdType.SASID.toLowerCase())
+              );
+            }
+            else {
+              childString.push(formatProperty(child['uniqueId'], 'uniqueId'));
+            }
+            return true;
+          }
```
vonbearshark, line 119: if the `propertyName === 'sasidUniqueId'` is mutually exclusive to `o && o.hasOwnProperty(propertyName)` they should be `if`/`else if` with each other rather than handling the ID case in with the other

vonbearshark: so like

```
if (o && o.hasOwnProperty(propertyName)) {
    // ... stuff
    return
} else if (propertyName === 'sasidUniqueId') {
    // diff stuff
    return
}
```

bamader: Oh yeah that makes sense

vonbearshark (approved):

JordanGuinn (approved):

JordanGuinn:
## Validation Plan
- [ ] Confirm the `sasid` column of all children is included in the exported roster for organizations with a `uniqueIdType` of `SASID`
- [ ] Confirm the `uniqueId` column of all children is included in the exported roster for organizations with a `uniqueIdType` of `Other`

JordanGuinn:
This looks good to me - once e2e tests pass, I'd say this is ready for merge 👍🏽

Issue #1320, opened 2021-05-07
Admins can search for orgs by name from the user edit view


JordanGuinn:
## Background
As an OEC admin, I should be able to look up organizations by name from the user edit view, in order to provide various users with site-level and organization-level permissions.

This is the same general UX as https://app.zenhub.com/workspaces/ece-dev-board-5ff506028d35f30012e0e937/issues/ctoec/data-collection/1310.

## Acceptance Criteria
- Admin users can look up orgs by email in the user edit view
   - **Admin only** org search API is written
   - Search API requires a **partial** match to return applicable results
- The search response includes a **full count** of all user matches, as well as the **organizations** themselves
- Users can **NOT** yet _add_ organizations to users
   - To be implemented in a separate ticket

## Technical Details
[SQL Server supports the % wildcard character](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/like-transact-sql?view=sql-server-ver15), so while it may or may not be possible with TypeORM to programmatically generate wildcarded LIKE queries, it can definitively be done manually if need be.

## Designs
[Full Figma files can be found here.](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=2%3A2)

#### Organization Search
![f4c9acf6-1805-4792-8b49-92d102c3c7d1 copy.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/679adb21-a0d7-413e-89a0-1fe023f7df28)

PR #1369, opened 2021-06-04
Search org from edit user page

nickbristow:
## Background
Adds form to search for orgs on the edit user page. 

Actual functionality to add the org to the user to come in later ticket

## GitHub Issue
#1320

## Validation Plan
- [ ] search for org by name, org should show in result
- [ ] search for org by partial, org should show
- [ ] search for org partial of multiple orgs, multiple orgs should show

## Automated Testing


nickbristow:
image

src/routes/organizations.ts
```diff
@@ -30,3 +34,20 @@ organizationsRouter.get(
     res.send(organizations);
   })
 );
+
+organizationsRouter.get(
+  '/by-name/:name',
+  passAsyncError(async (req, res) => {
+    try {
+      const name = req.params['name'];
```
JordanGuinn, line 42: Any particular reason we don't just make `name` an optional query parameter on the other GET organizations route, instead of maintaining two entirely separate endpoints?

src/controllers/organizations.ts
```diff
@@ -63,3 +66,14 @@ from organization o left join (
         ) s on o.id = s.organizationId
 group by o.id, o.providername, s.siteCount`
   );
+
+function escapeLikeString(raw: string): string {
+  return raw.replace(/[\\%_]/g, '\\$&');
```
JordanGuinn, line 71: 👀👀

client/src/containers/Users/EditUser.tsx
```diff
@@ -52,6 +55,21 @@ const EditUser: React.FC = () => {
...
+  const [showSearchResults, setShowSearchResults] = useState(false);
+  const [foundOrgs, setFoundOrgs] = useState<Organization[]>([]);
+  const searchForOrgs = async (query: string) => {
+    setShowSearchResults(false);
+    await apiGet(`/organizations/by-name/${query}`, accessToken)
+      .then((res) => {
+        setFoundOrgs(res);
+        setShowSearchResults(true);
+      })
+      .catch((err) => {
```
JordanGuinn, line 67: Do we _really_ need to catch if we're just gonna rethrow the same error?

JordanGuinn (commented):
Looks good overall, just a few little things 👍🏽👍🏽

JordanGuinn:
Also, can the `Add Organization` h2 get turned into `Permissions`, with an `Add Organization` `h3` underneath it?  I think that's how the designs have it as well

src/controllers/organizations.ts
jlhogan: add comment on why this

jlhogan: also,  i think we should replace all whitespace and punct. with `%` for our like query. so that something like `st francis` will match `st. francis`

src/routes/organizations.ts
jlhogan: ^ this please

src/controllers/organizations.ts
nickbristow: from my research on this, typeorm won't escape this Like pattern.

client/src/containers/Users/EditUser.tsx
```diff
@@ -52,6 +55,21 @@ const EditUser: React.FC = () => {
     })();
   }, [adminUser, accessToken]);
 
+  const [showSearchResults, setShowSearchResults] = useState(false);
+  const [foundOrgs, setFoundOrgs] = useState<Organization[]>([]);
+  const searchForOrgs = async (query: string) => {
+    setShowSearchResults(false);
```
vonbearshark, line 61: Do we need both the `showSearchResults` and `foundOrgs` separate state vars? Can't we just use `foundOrgs`?

Here we could `setFoundOrgs([])`, and in the render check we could `foundOrgs.length && ...`, right?

src/routes/organizations.ts
vonbearshark: Could you say more @JordanGuinn @jlhogan ? I think I prefer the explicit interface of `organizations/name/:name`. In the case where we'd want to add an ID-based check, would you just add that to the `/` GET endpoint as well? I would prefer separate, specific, and explicit routes, I think. But would love your thoughts

JordanGuinn: Most list APIs don't provide independent endpoints for each of the filterable parameters supported said API supports

JordanGuinn: In my eyes, it makes the most sense to have a single LIST organizations route that supports all query parameters that we'd like it to, rather than needing to make an explicit endpoint each time we decide to want to query against a different column.  Like this is the _exact reason_ query parameters exist

vonbearshark: Yeah in thinking about it I retract my comment

client/src/containers/Users/EditUser.tsx
nickbristow: @vonbearshark how should we distinguish between finding 0 results?

vonbearshark: We could use `null | []` type?

client/src/containers/Users/EditUser.tsx
```diff
@@ -52,6 +55,14 @@ const EditUser: React.FC = () => {
     })();
   }, [adminUser, accessToken]);
 
+  const [foundOrgs, setFoundOrgs] = useState<Organization[] | null>(null);
+  const searchForOrgs = async (query: string) => {
+    setFoundOrgs([]);
```
vonbearshark, line 60: wouldn't this be `null` here?

nickbristow: yeah, not actually necessary

src/routes/organizations.ts
```diff
@@ -19,14 +23,15 @@ organizationsRouter.post(
       throw new InternalServerError('New organization not created.');
     }
   })
-)
+);
 
 organizationsRouter.get(
-  '/',
+  '/:name',
```
vonbearshark, line 29: I think @JordanGuinn and @jlhogan were pushing for just using `/`, and extracting the params there (which I agree with)

nickbristow: @vonbearshark updated, hows that?

JordanGuinn (approved):

vonbearshark (approved):

JordanGuinn:
## Validation Plan
- [ ] Confirm admin org list view still functions and displays all orgs
- [ ] Confirm exact match of org name search works
- [ ] Confirm partial match of org name search works, and includes all appropriate orgs
- [ ] Confirm I can't perform SQL injection through the search box
- [ ] Confirm I can't click search without typing anything

JordanGuinn:
Validation is complete, no issues found!  So once e2e passes, I'd say this is ready for merge

Issue #1355, opened 2021-05-27
Reintroduce error messaging to upload UI after invalid file uploads


JordanGuinn:
## Background
At some point, the error message that we displayed previously whenever a user upload a file that we were unable to ingest was inadvertently removed.  We should add that error messaging back in, and while we're at it, simplify it by instructing the user to re-download the data template and port their data into said downloaded template.

## Acceptance Criteria
- Error messaging is displayed whenever a user attempts to upload a incorrectly formatted file, or any other type of file that we simply can't parse
   - i.e. whenever the `enrollment-reports/check` route fails with a `400`
- Error message: "**Files must be uploaded in the format of our data template.** Please re-upload your file using the [template provided](https://ece-reporter.ctoec.org/template). If you need further support, send us a [support request](https://ece-reporter.ctoec.org/support)."

PR #1372, opened 2021-06-08
Add error messaging to upload

bamader:
## Background
At some point, we had error messaging that informed a user of an incorrectly formatted file upload. That messaging disappeared, so this PR adds it back in while also simplifying the messaging states/cases to always direct the user to upload data only in one of our templates.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1355

## Validation Plan
- Open and log into app
- Go to file upload page
- [ ] Try to upload a csv or xlsx that has different formatting from the template
  - [ ] Verify error message appears and directs user to the template
  - [ ] Error alert disappears on page refresh or successful file upload
- [ ] Try to upload a file of the completely wrong type (i.e. a .zip, .jpg, etc.)
  - [ ] Verify error message appears and directs user to the template
  - [ ] Error alert disappears on page refresh or successful file upload

## Automated Testing

## Additional Context
The reason for testing on two different file types is that we have two different error cases in the back end depending on the type of error that happens while parsing. Spreadsheet files with improperly formatted columns will trigger an `ApiError` version of a 400, (lines 62-66 in `src/routes/enrollmentReport.ts`), while files that aren't spreadsheets at all will trigger the more general `BadRequestError` type of 400 (lines 67-69). We want this error message to appear for both of these cases, since the ticket's intention in simplifying the messaging was to handle all 400s by having the user use a proper template.


JordanGuinn (approved):

client/src/containers/Upload/Upload.tsx
```diff
@@ -108,6 +112,7 @@ const Upload: React.FC<StepContentProps> = ({
   return (
     <div className="grid-container">
       <BackButton location="/" />
+      {alertElements}
```
vonbearshark, line 115: welp yeap there's my problem...

vonbearshark (approved):

JordanGuinn:
## Validation Plan 
- [ ] Confirm no error messaging is present on valid file uploads
- [ ] Confirm error messaging is present on upload of CSV with improper number of columns
- [ ]  Confirm error messaging is present on upload of CSV with improperly named columns
- [ ]  Confirm error messaging is present on upload of invalid file type (zip)

JordanGuinn:
@bamader Functionally all looks well!  Just got a few minor thoughts on the copy here:

- Could the alert title be updated to something like `We were unable to upload your file`?  I think the title should **first** be indicative that we weren't actually able to ingest their data, while the alert's internal text should direct them to our templates
- Instead of including links directly to the CSV/Excel templates here, could the `templates provided` text instead be a link to the data templates page?  Just cuz (imo) the presence of those additional file download links in the alert make it a pretty chunky element, on an already content-heavy page

bamader:
@JordanGuinn Yes to both changes, and both are implemented. E2e's previously passed, so this is ready to merge whenever.

bamader:
@JordanGuinn Am I cool to merge this? Wanna get it in before the sprint rituals this afternoon.

JordanGuinn:
Do it!  Sorry, thought I responded earlier then got REAL sidetracked

Issue #1363, opened 2021-06-03
Providers should have the ability to specify enrollment exit dates in the future


JordanGuinn:
## Background
Providers often know the enrollment exit dates of children before said dates actually arrive, so we should allow them to specify those future enrollment exit dates accordingly.  Additionally, these children should continue to show up in the provider's roster view until that future exit date has arrived.

This is currently being prevented by an Enrollment entity decorator on the `exit` column server-side, so removing that decorator should just allow for this to happen, honestly.

## Acceptance Criteria
- Users can specify enrollment exit dates in all areas of the application, including while:
   - Uploading a file
   - Withdrawing a child
   - Editing a past enrollment
- Enrollments with an exit date in the future no longer show up in the `Past Enrollments` section of a child detail view
- An enrollment with a future exit date is **considered active until that date is reached**
   - That is to say, they should appear in the roster until the exit date comes

## Technical Specs
There is a `MomentComparison` decorator on the `exit` column on the `Enrollment` entity that currently enforces the validation of this, at least server-side.  So thankfully, simply removing this decorator entirely should theoretically take care of a good portion of the functional work necessary here.

PR #1374, opened 2021-06-10
Allow exit dates in the future

nickbristow:
## Additional Bug Caveat
See comment at the bottom of the validation discussion for dumb withdraw model behavior.

## Background
Exit dates should be allowed for when a user knows the exit date of a child

## GitHub Issue
#1363 

## Validation Plan
- [ ] upload regular data without exit date, expect no error
- [ ] upload with exit date and exit reason, expect no error
- [ ] upload with exit date before start date, expect error
- [ ] withdraw student with future exit date, expect child to show in active

## Automated Testing

## Additional Context


src/controllers/children/get.ts
```diff
@@ -191,12 +191,13 @@ export const getWithdrawnChildren = async (
   // Empty enrollments qualify as "active"
   qb.andWhere('enrollment.entry IS NOT NULL');
   // Subquery: find active enrollments by childId
+  let end = moment().endOf('month').format('YYYY-MM-DD');
```
jlhogan, line 194: why this? why not today?

src/entity/decorators/Enrollment/exitDateAfterEntry.ts
```diff
@@ -0,0 +1,28 @@
+import { registerDecorator, ValidationOptions } from 'class-validator';
```
jlhogan, line 1: <3!

jlhogan (approved):

src/controllers/children/get.ts
nickbristow: So if you look at getActiveChildren, it includes children up to the end of the current month. So I was following that. 

https://github.com/ctoec/data-collection/blob/base/src/controllers/children/get.ts#L162

bamader: That makes sense to me!

bamader (approved):
LGTM!

vonbearshark (approved):

bamader:
## Validation Plan
I'll just leverage Nick's because it covers the cases I can think of.
- [ ] upload regular data without exit date, expect no error
- [ ] upload with exit date and exit reason, expect no error
- [ ] upload with exit date before start date, expect error
- [ ] withdraw student with future exit date, expect child to show in active

bamader:
This one isn't quite ready to merge. There's some sort of bug in the interaction between future exit dates and funding reporting periods, and there's the question of how should enrollments with future exit dates display on the edit record page. Now, granted, we're doing a roster overhaul, which means these questions might not deserve a whole lot of consideration. If that's the case, and if the bugginess around reporting periods blocking withdrawals is something we know about, then I would change my validation stamp and mark this as ready.

bamader:
@nickbristow is this PR ready for re-validation?

bamader:
Just finished re-validation. It took a surprisingly long time to clear up that last bug, but I ended up finding something that this PR wasn't responsible for. The issue I ran into is that if you attempt to withdraw a child using the Withdraw Modal, and your first attempt fails because of an error, _all successive attempts while the modal is open are also errors_. To clarify: click on a child in edit record, then click Withdraw. Set an exit date of 12/12/2021 (in the case of this PR), then choose Moved within CT for exit reason. Hit submit without selecting a last reporting period. The withdraw will fail. Now, try to select a last reporting period (say, June 2021). The withdraw fails again, _even though all information is given_. If you close the withdraw model, then click withdraw again, you can enter in all the same choices (exit date 12/12/2021, Moved within CT, June 2021 last reporting period), click withdraw, and the call will succeed.

This is probably a dumb edge case, but it is a bug that currently exists in our system (verified in staging and QA) and not something this PR introduces. Therefore, I consider this PR validated and ready for merging.

Issue #1342, opened 2021-05-19
Add explicit error UX to create org admin view


Derekskylight:
## Background
As a provider creating new organizations, there are scenarios in which not all required tables are filled out correctly. In such a scenario, I can see error states/messages that show appropriate messaging and will redirect me to completely fill out the correct information. Doing so will increase the correctness of child enrollment data while also reducing the number of customer support requests submitted towards members of the OEC ECE team at Skylight. 

## Acceptance Criteria
- The `Please enter all required information` error text is displayed at the top of the page whenever there are **any** issues within the form to be corrected

#### Sites
- If a user has not properly filled out all fields for a particular site, an error in red text above the forms will be displayed indicating as much 
   - Error text should read `Please enter all required site information`
- The fields that need to be properly filled out should be receive a red border until they're corrected

#### Organization Name
- If user submits an organization whose name is already in the ECE Reporter database, user will receive a notification saying so and will not be able to submit the creation of this organization. 
   - Error text should read `This organization name already exists`
- The `Organization Name` field should receive a red border until corrected

#### Funding Spaces
- If user does not click on a funding space or form in this section, user will receive a notification saying so and will not be able to create a new organization until this has been clicked on
- The fields that need to be properly filled out should be receive a red border until they're corrected

## Designs
[The full set of designs can be found in Figma.](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=86%3A348)

#### Sites Error UX
![Error.PNG](https://images.zenhubusercontent.com/602d78337403e44d7db5c4f3/67c92059-4454-4c7d-b1b7-080bf50601c6)

#### Org and Funding Spaces Error UX
![Error2.PNG](https://images.zenhubusercontent.com/602d78337403e44d7db5c4f3/5f8c5807-2ad9-4857-a8de-682e11477833)

JordanGuinn:
@Derekskylight Do you remember the context for this ticket, by chance?

Derekskylight:
Hey I cleared this up with Emily and this has to do with error states in org creation in the admin view! Editing ticket here accordingly.

PR #1377, opened 2021-06-16
Error UX for create org page

nickbristow:
## Background
Adds basic form field validation on the Create org page. Adds field level errors and error callouts for sites and funding spaces. 

Replaces basic form components with carbon components.

## GitHub Issue
#1342 

## Validation Plan
- [ ] submitting form with no info shows following errors (org name missing, missing info on site top level message, missing info on funding space top level message, errors on fields site name, title, region, funding source, age group, space type, capacity
- [ ] Submitting form with partial info throws errors only on missing fields, top level messages appear
- [ ] Multiple top level alerts appear
- [ ] Submitting form with no missing data completes as expected 

## Automated Testing

## Additional Context


JordanGuinn:
@nickbristow What's the word on this?  I see it's been marked as `Ready for Review`, but the PR description contains zero context and no reviewers have been requested

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -77,24 +79,31 @@ const CreateOrg: React.FC = () => {
...
           text: `One or more requested funding spaces are not unique.`,
         },
       ]);
+      errors = true;
+    }
+
+    if (errors) {
+      setShowErrors(true);
       return;
     }
```
vonbearshark, line 108: nit: can we just `setShowErrors(errors)`?

client/src/containers/CreateOrg/NewFundingSpaceFormCard.tsx
```diff
@@ -32,83 +41,106 @@ export const NewFundingSpaceCard: React.FC<NewFundingSpaceFormCardProps> = ({
     return tmp;
   };
 
+  //  Array of different field booleans so the form updates whenever
+  //  any field is updated.
+  const getMissingInfo = () => {
+    return {
+      source: !newFundingSpace.source,
+      ageGroup: !newFundingSpace.ageGroup,
+      capacity: !newFundingSpace.capacity && newFundingSpace.capacity != 0,
```
vonbearshark, line 50: I think this is redundant, right? If `capacity`  is `0` then `!capacity` should be `true`?

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -196,15 +205,20 @@ const CreateOrg: React.FC = () => {
...
-            return e.target.value;
-          }}
-        />
+        <div className="grid-row">
+          <div className="tablet:grid-col-6">
+            <TextInput
+              labelText="Organization Name"
+              id="new-org-name-input"
+              type="input"
+              invalid={showErrors && (!newOrgName || newOrgName == '')}
```
vonbearshark, line 214: I think this is redundant, right? `!''` is `true`, right?

vonbearshark: And would this be `||`?

client/src/containers/CreateOrg/NewFundingSpaceFormCard.tsx
```diff
@@ -32,83 +41,106 @@ export const NewFundingSpaceCard: React.FC<NewFundingSpaceFormCardProps> = ({
     return tmp;
   };
 
+  //  Array of different field booleans so the form updates whenever
+  //  any field is updated.
+  const getMissingInfo = () => {
```
vonbearshark, line 46: This system of the parent tracking whether an error is shown and the children separate tracking the missing info states is confusing to my eye.

What if the error state was an array of errors, like the alerts, established in the parent and passed to the children?

So in `CreateOrg`: `const [errors, setErrors] = useState([]);` and later `setErrors(current => [...current, newError])`. Then `<NewSiteFormCard ... errors={errors} setErrors={setErrors} />`, etc

And here (and in `NewSiteFormCard`) use that in place of `missingInfo`

vonbearshark: Might be that an object would be better than an array for tracking errors, more akin to this `missingInfo` shape you have right now?

client/src/containers/CreateOrg/CreateOrg.tsx
bamader: I think JJ is right  that `!''` evaluates to `true`. So you should be able to make the second condition just `!newOrgName`. As far as whether the overall conditional test is 'and' or 'or', I _do_ think it's an `&&` here. Because `showErrors` starts out as `false` but `newOrgName` starts out as `''`, if this were an `||`, then all the input fields would be highlighted as invalid before the user even submits anything, which isn't what we want. So I think the `&&` is correct but the interior condition in parentheses can just be reduced to `!newOrgName`.

client/src/containers/CreateOrg/NewFundingSpaceFormCard.tsx
bamader: I think you're right JJ

client/src/containers/CreateOrg/NewFundingSpaceFormCard.tsx
bamader: I might even suggest a different approach altogether, though I do agree that the current setup threw me for a loop at first. What if we defined an enum that had different strings representing the various possible states of a funding card? So, `getMissingInfo` becomes about a series of `if/else` checks that return the appropriate enum(s) to describe the state of a funding card on the page (e.g. a funding card that had no age group or capacity might have a return value of `[NO_AGE_GROUP, NO_CAPACITY]` where each of those is some appropriately defined string). We could then use the returned strings to show the right state of error on the page?

bamader (commented):
Most of the code looks good. I just think we should clear up what our plan is to track error states in the various cards before we proceed. May be worth having a pairing session tomorrow with the three of us to talk about it? Could also just discuss async if that's easier.

client/src/containers/CreateOrg/NewFundingSpaceFormCard.tsx
nickbristow: @bamader @vonbearshark 
So I tried to address some of these issues. The main thing i wanted to do was only validate in one spot. Because we have the ability to add multiple sites and multiple funding spaces, I decided to leave the validation on that component, and removed it from the parent. 

So each card now has an errors property which the parent just checks to see if it has anything on any of them before submitting.

I had to do some weirdness with the state to keep that updating so it would re render and clear errors when they are fixed.

Let me know if this doesn't make sense...

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -30,41 +34,46 @@ const CreateOrg: React.FC = () => {
...
     }
 
-    const allFundingSpaceOkay = newFundingSpaces.every(
-      (nfs) =>
-        !!nfs.source &&
-        !!nfs.ageGroup &&
-        (!!nfs.capacity || nfs.capacity === 0) &&
-        !!nfs.time
+    const allFundingSpaceOkay = !newFundingSpaces.some(
+      (nfs) => nfs.errors && nfs.errors.length > 0
```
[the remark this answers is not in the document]
vonbearshark: Also `every` with `!nfs?.errors?.length`?

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -30,41 +34,46 @@ const CreateOrg: React.FC = () => {
...
-      return;
+      errors = true;
     }
-    const allSitesOkay = newSites.every(
-      (ns) => ns.siteName !== '' && ns.titleI !== undefined && ns.region
+    // const allSitesOkay = newSites.every(
+    //   (ns) => ns.siteName !== '' && ns.titleI !== undefined && ns.region
+    // );
+    const allSitesOkay = !newSites.some(
+      (ns) => ns.errors && ns.errors.length > 0
```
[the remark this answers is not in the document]
vonbearshark: Also `every` with `!ns?.errors?.length`?

client/src/containers/CreateOrg/NewFundingSpaceFormCard.tsx
vonbearshark: More closely reading the code and I am realizing my concerns stem from how we are already managing the data of this component, updating the props by reference rather than passing in state manipulation functions to the child components. Given that this seems fine enough for me, and it's not really worth any optimizations given the current setup

client/src/containers/CreateOrg/CreateOrg.tsx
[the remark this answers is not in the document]
bamader: No I think Nick is right here. His conditional says all sites are okay if it is _not_ the case that _some_ `ns` has `errors` with a length above 0. Love me an opportunity to use DeMorgan's Laws from my freshman year intro to philosophy class 😂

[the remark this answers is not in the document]
bamader: Unless you're actually just saying you'd prefer `allSitesOkay = newSites.every( !ns?.errors?.length )`, in which case ignore me lol

client/src/containers/CreateOrg/CreateOrg.tsx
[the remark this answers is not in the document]
bamader: Same as above

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -126,28 +140,31 @@ const CreateOrg: React.FC = () => {
...
     setNewSites((currentSites) => [...currentSites, { ...emptySite }]);
   };
 
-  const emptyFundingSpace: Partial<FundingSpace> = {
+  const emptyFundingSpace: Partial<FundingSpaceWithErrors> = {
     source: undefined,
     ageGroup: undefined,
     capacity: undefined,
     time: undefined,
+    errors: [],
```
bamader, line 163: Why include the errors property explicitly here with the `FundingSpaceWithErrors` partial but not the `SiteWithErrors` partial?

client/src/containers/CreateOrg/NewFundingSpaceFormCard.tsx
bamader: Nick I think this pedagogically makes sense. Having validations just live in one place, on the child, and then having the parent check its children for their status seems logical. JJ, I also agree with you that it's not really worth trying to optimize the 'react-y-ness' of this anymore so I'm good with this as is.

bamader (approved):

client/src/containers/CreateOrg/NewFundingSpaceFormCard.tsx
```diff
@@ -32,83 +41,127 @@ export const NewFundingSpaceCard: React.FC<NewFundingSpaceFormCardProps> = ({
     return tmp;
   };
 
+  const Errors: FundingSpaceErrors = {
```
JordanGuinn, line 44: Why string values?  Why not just booleans?

JordanGuinn: Also, if we did end up using booleans here, we could just use `Record` and `keyof` and delete the `FundingSpaceErrors` type altogether.  Something like:

```
const Errors: Record<keyof FundingSpace, boolean> = {
    source: false,
    ageGroup: false,
    capacity: false,
    time: false,
  };
```

JordanGuinn: Also also, can errors be lowercase?

client/src/containers/CreateOrg/NewSiteFormCard.tsx
```diff
@@ -25,74 +32,114 @@ export const NewSiteFormCard: React.FC<NewSiteFormCardProps> = ({
   newSite,
   numberOnPage,
   remove,
+  showErrors,
 }) => {
+  const Errors: SiteErrors = {
```
JordanGuinn, line 37: Same feedback here as with the funding spaces component:
- Can we make `Errors` lowercase?
- Can the values be booleans instead of strings?
- Can we use the `Record<keyof Site, boolean>` type and delete `SiteErrors` entirely?

client/src/containers/CreateOrg/NewSiteFormCard.tsx
```diff
@@ -25,74 +32,114 @@ export const NewSiteFormCard: React.FC<NewSiteFormCardProps> = ({
...
+  };
+
+  const getMissingInfo = () => {
+    const tmp: String[] = [];
+    Object.keys(Errors).forEach((k) => {
+      const siteKey = k as keyof SiteWithErrors;
+      const errorKey = k as keyof SiteErrors;
+      if (!newSite[siteKey]) tmp.push(Errors[errorKey]);
+    });
+    return tmp;
```
JordanGuinn, line 50: something of a nit: can all the usages of the variable `tmp` get replaced with something more informative of the data structure?

client/src/containers/CreateOrg/NewFundingSpaceFormCard.tsx
```diff
@@ -32,83 +41,127 @@ export const NewFundingSpaceCard: React.FC<NewFundingSpaceFormCardProps> = ({
...
+    capacity: 'CAPACITY',
+    time: 'TIME',
+  };
+
+  //  Array of different field booleans so the form updates whenever
+  //  any field is updated.
+  const getMissingInfo = () => {
+    const tmp: String[] = [];
+    Object.keys(Errors).forEach((k) => {
+      const spaceKey = k as keyof FundingSpaceWithErrors;
```
JordanGuinn, line 56: If you end up using the `Record` Typescript utility as mentioned above, you should simply be able to case the key to `FundingSpace` once (if you even need to at all), and reference that same key in both objects.  Like:

```
const newKey = k as keyof FundingSpace;
if (!newFundingSpace[spaceKey]) tmp.push(Errors[errorKey]);
```

client/src/containers/CreateOrg/NewSiteFormCard.tsx
```diff
@@ -25,74 +32,114 @@ export const NewSiteFormCard: React.FC<NewSiteFormCardProps> = ({
...
+  const Errors: SiteErrors = {
+    siteName: 'SITE_NAME',
+    titleI: 'TITLE_I',
+    region: 'REGION',
+  };
+
+  const getMissingInfo = () => {
+    const tmp: String[] = [];
+    Object.keys(Errors).forEach((k) => {
+      const siteKey = k as keyof SiteWithErrors;
```
JordanGuinn, line 46: Same comment as above - you should simple be able to cast k to `keyof Site` once and use that key in both places, assuming you use the `Record` utility outlined in the comments above.

client/src/containers/CreateOrg/CreateOrg.tsx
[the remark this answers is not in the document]
vonbearshark: Yeah, just the nit for `allSitesOkay = newSites.every( !ns?.errors?.length )` preference

client/src/containers/CreateOrg/NewFundingSpaceFormCard.tsx
nickbristow: I agree, thats how i had it originally.

nickbristow:
@JordanGuinn @vonbearshark @bamader 
Updated based on your comments.

JordanGuinn (approved):

vonbearshark (approved):

JordanGuinn:
## Validation Plan
### General
- [ ] Confirm no error messaging is displayed until the user attempts to submit the organization to be created
- [ ] User is able to create organization on first try if there are no errors present
- [ ] User is able to create organization after addressing errors outlined in create org form

### Funding Space
- [ ] Alert banner at **top of the page** is displayed if there are any funding space errors
- [ ] Error message is displayed in funding space card if there are _any_ funding space errors
- [ ] Individualized error messaging is displayed for each and every invalid funding space field
- [ ] Individualized error messaging disappears when corresponding funding space field are corrected
- [ ] Error messaging in funding space card disappears when all funding space fields are corrected
- [ ] Alert banner at **top of the page** disappears when all funding space errors are corrected

### Site
- [ ] Alert banner at **top of the page** is displayed if there are any site errors
- [ ] Error message is displayed in site card if there are _any_ site errors
- [ ] Individualized error messaging is displayed for each and every invalid site field
- [ ] Individualized error messaging disappears when corresponding site field are corrected
- [ ] Error messaging in site card disappears when all site fields are corrected
- [ ] Alert banner at **top of the page** disappears when all site errors are corrected

JordanGuinn:
Overall this is generally functional - just found a few relatively minor issues:

- **Limited to dropdowns** - If you select an option from a funding space/site dropdown, then select `Select` again, then submit the form, those fields do not show up as being invalid, even though nothing has been selected
   - Looks like explicitly selecting `Select` sets the field value to the string `true`
- The `TitleI` field, when set to `No`, displays the invalid field error messaging
- (Not necessarily a blocker) When error banners at the top of the page are present, and the user adds an organization when one is not specified, the user is scrolled to the top of the page on blur

JordanGuinn:
Re-validated, and confirmed the two previously blocking issues have been addressed.  I'd say this is ready for merge 👍🏽

Issue #1373, opened 2021-06-10
Remove reporting periods from ECE Reporter


JordanGuinn:
## Background
The `June 2021` reporting period is the **final** reporting period providers will be able to specify as the `first reporting period` for enrollment fundings, which naturally is problematic - it's currently June 2021!

Additionally, given the real lack of any value reporting periods have provided us up to this point, now seems like as good a time as ever to do away with them entirely, and simply allow for all fundings to have their own specific start and end dates.

## Acceptance Criteria
- `Reporting periods` display text in `Funding` cards is replaced with `Funding period` (to indicate range between new start and end date)
- Fundings can be created/edited with a custom start/end date, independent of any previous notion of reporting periods
   - `First reporting period` dropdown is replaced with `Funding start date` datepicker across the application
   - `Last reporting period` dropdown is replaced with `Funding end date` datepicker across the application
- File uploads are updated according to of the following:
   - Map `First funding period` and `Last funding period` columns to these new columns
   - Change `First funding period` and `Last funding period` to `Funding start date` and `Funding end date` in the upload template

## Technical Specs
- Migration is written to add `startDate` and `endDate` columns directly to `Funding` entity
- ETL is written to perform the following for all current fundings:
   - For enrollments with **a single funding**:
      - Copy `enrollment.entry` into `funding.startDate` column
      - Copy `enrollment.exit` into `funding.endDate` column
   - For enrollments **with multiple fundings**:
      - Copy `reportingPeriod.startDate` of **first reporting period** into `funding.startDate` column
      - Copy `reportingPeriod.endDate` of **last reporting period** into `funding.endDate` column
- Follow-up migration is written to remove `firstReportingPeriod` and `lastReportingPeriod` columns from `Funding` entity
- Funding decorators are updated to reference the new `startDate` and `endDate` columns directly

Derekskylight:
Hi! Just to confirm with this, will the Excel file template also be reflected to show these changes? And will file upload still accept the template if there is any missing data under the column 'first funding period?'

PR #1390, opened 2021-06-30
1373 reporting periods

jlhogan:
## Background
Replaces reporting period references on fundings with inlined dates. Includes updating all server + client logic to account for this change, in addition to DML query to populate the new funding start/end dates with values from enrollment entry/exit or reporting period start/end. 

## GitHub Issue
#1373 

## Associated PRs
N/A

## Validation Plan
- [x] Validate DML query is correct
- [x] Funding start/end can be added via file upload
- [x] Funding can be modified to have end via file upload
- [x] Funding can be created in tool to have start
- [x] Funding can be edited to have different start
- [x] Funding can be ended with end date
- [x] Funding can be changed by creating a new funding, adding an end date to existing funding and creating new one with start date
- [x] Record with active funding can be withdrawn, including adding end date to funding
- [x] Confirm updated validations work as expected
    - [x] no overlapping fundings
    - [x] funding exit date after start date
    - [x] funding start date after enrollment start date

## Automated Testing


src/entity/decorators/Enrollment/fundingsDoNotOverlap.ts
```diff
@@ -0,0 +1,46 @@
+import { registerDecorator, ValidationOptions } from 'class-validator';
+import { propertyDateSorter } from '../../../utils/propertyDateSorter';
+import { Funding } from '../../Funding';
+
+// Make sure first reporting period of a funding is after the last reporting period of the previous funding
```
bamader, line 5: DIE COMMENT

src/entity/decorators/Enrollment/fundingsDoNotOverlap.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const fundingPeriodOverlap =
+  'Cannot claim a child twice in a reporting period.';
```
bamader, line 8: Need to update the copy here--perhaps `Cannot claim a child twice in a given time period.`

src/migration/1624989301327-UpdateFundingAddStartEndDate.ts
```diff
@@ -0,0 +1,52 @@
...
+					SET	
+						funding.startDate = (
+							CASE 
+								WHEN isMultiple = 0 THEN entry else firstReportingPeriod.periodStart
+							END
+						),
+						funding.endDate = (
+							CASE
+								WHEN isMultiple = 0 THEN [exit] else lastReportingPeriod.periodEnd
+							END
```
bamader, line 23: Some general thoughts / question about migrations. As I understand them, they're run every time the database is created/re-created. So as this is written now, the DB will still create and store data using reporting periods until it hits this migration, at which point it will move the data and then (in the next migration) delete the reporting periods. Question 1: will there be a problem if this migration is run more than once? As in, if reporting periods get deleted and we're not storing provider data in that form anymore, then will this migration cause errors if it runs again and there's no reporting period column to pull from? Or will it copy over null / inaccurate values? Question 2: I know it's generally not good practice to modify old migrations, but is it worth deleting all records of reporting periods from the migrations after we run this for the first time? That way, we get all the data migrated into its new form, and then we don't have to worry about reporting period column errors going forward? I might be wrong on my counts here, but just voicing the thoughts for discussion.

bamader (approved):
Other than the super weird spacing that I'm surprised `prettier` didn't catch, and some thoughts on the migrations, this seems...good to me? The only reason I question that is because this is a loooooong PR, but everything I'm seeing here is clear, intuitive, and seems like it should just...work. I really don't have any issues. If it's true, then kudos to you, Julia, for such a massive refactor that doesn't seem to break anything! So, as long as we do some _extensive_ manual validations before merging this, this looks good to me.

src/migration/1624989301327-UpdateFundingAddStartEndDate.ts
JordanGuinn: For your first question - if we're running this migration more than once, then it means we're running it against a completely empty database, in which case there will also be no fundings to update, either.  We should validate to confirm, but that shouldn't be an issue.

For your second question - we should **absolutely not** modify the old migrations to remove reporting periods.  In addition to that introducing the potential for enormous cascading effects in other migrations, it would _also_ mean we'd have to re-run **all subsequent migrations** from that point again across all environments (which is flat-out not possible in production, as we have live data).  Much more complicated to rewrite previous history than to just update the present 🙂

client/src/utils/models/getCurrentFunding.ts
```diff
@@ -18,7 +18,7 @@ export const getCurrentFunding = (source: {
   const today = moment.utc();
   return (_enrollment.fundings || []).find(
     (funding) =>
-      !funding.lastReportingPeriod ||
-      funding.lastReportingPeriod.periodEnd.isSameOrAfter(today)
+      !funding.endDate ||
+      funding.endDate.isSameOrAfter(today)
```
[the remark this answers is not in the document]
bamader: I think you're right, Nick. There should be a moment function which should round the result returned by `moment.utc` to be in just day form?

src/entity/decorators/Funding/fundingBeginsAfterEnrollmentEntry.ts
```diff
@@ -14,20 +15,11 @@ export function FundingBeginsAfterEnrollmentEntry(
...
+        validate(startDate: Moment, { object: funding }) {
           const { entry } = (funding as Funding)?.enrollment;
-          const endFirstReportingPeriod = reportingPeriod.periodEnd;
-          if (!entry || !endFirstReportingPeriod) return true;
-          // The end of the first reporting period must be after the enrollment start
-          const isValid = entry.isBefore(endFirstReportingPeriod);
-          // Delete the enrollment from funding bc it was added in the validation process.  See enrollment entity
-          delete (funding as Funding)?.enrollment;
-          return isValid;
+					if (!entry) return true; // Can't do this validation without the enrollment entry
```
vonbearshark, line 20: wouldn't we want to return `false`?

vonbearshark (approved):
spacing is pretty wild throughout. didn't comment everywhere. Code wide looking great!

src/entity/decorators/Funding/fundingBeginsAfterEnrollmentEntry.ts
jlhogan: i think false means there _is_ a validation error? Will confirm

client/src/utils/models/getCurrentFunding.ts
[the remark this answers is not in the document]
jlhogan: on it !

src/entity/decorators/Funding/fundingBeginsAfterEnrollmentEntry.ts
bamader: I think you're correct, Julia. I believe true is "everything is good" and false = borked.

vonbearshark: Yeah realizing this is validating the `Funding` not the `Enrollment`. makes sense to me

client/src/containers/EditRecord/EnrollmentFunding/EditEnrollmentCard.tsx
```diff
@@ -132,11 +133,17 @@ export const EditEnrollmentCard: React.FC<EditEnrollmentCardProps> = ({
         <div className="flex-2">
           <p className="margin-bottom-0">Enrollment dates</p>
           <p className="text-bold margin-top-0">
-            {enrollment.entry && enrollment.entry.isValid()
-              ? enrollment.entry.format('MM/DD/YYYY')
-              : InlineIcon({ icon: 'incomplete' })}{' '}
-            -{' '}
-            {enrollment.exit ? enrollment.exit.format('MM/DD/YYYY') : 'present'}
+            {enrollment.entry
```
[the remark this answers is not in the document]
jlhogan: we should be doing validity checks on the server side only; so here we only need to check if validation errors occur

client/src/containers/EditRecord/EnrollmentFunding/EditEnrollmentCard.tsx
```diff
@@ -132,11 +133,17 @@ export const EditEnrollmentCard: React.FC<EditEnrollmentCardProps> = ({
...
+							? enrollment.entry.format('MM/DD/YYYY') 
+							: InlineIcon({ icon: 'incomplete' })
+						}
+						{' '}-{' '}
+            {hasValidationErrorForField(enrollment, 'exit')
+							? InlineIcon({ icon: 'incomplete' })
+							: !!enrollment.exit 
+							? enrollment.exit.format('MM/DD/YYYY') 
+							: 'present'
+						}
```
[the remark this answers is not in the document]
jlhogan: i like nested ternaries for clear + concise description of flow. if someone else has a deciding opinion, happy to go in whatever direction gets more votes

client/src/containers/EditRecord/EnrollmentFunding/EditFundingCard.tsx
```diff
@@ -135,8 +118,20 @@ export const EditFundingCard: React.FC<EditFundingCardProps> = ({
...
+							: funding.startDate?.format('MM/DD/YYYY')
+						}
+						{' '}-{' '}
+						{hasValidationErrorForField(funding, 'endDate')
+							? InlineIcon({ icon: 'incomplete' })
+							: !!funding.endDate
+							? funding.endDate?.format('MM/DD/YYYY')
+							: 'present'
+						}
+					</p>
```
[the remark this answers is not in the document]
jlhogan: i dont think that's worth the additional complexity required

src/index.ts
```diff
@@ -20,10 +20,10 @@ getConnectionOptions().then((connectionOptions) => {
     .then(async () => {
       console.log('Successfully established TypeORM DB connection');
 
-      if (!isProdLike()) {
-        console.log('Seeding application data...');
-        await initialize();
-      }
+      // if (!isProdLike()) {
```
JordanGuinn, line 23: reminder that this will need to be uncommented out before merge

src/migration/1624998391756-UpdateFundingRemoveReportingPeriod.ts
```diff
@@ -0,0 +1,42 @@
...
+      `ALTER TABLE "funding" ADD "firstReportingPeriodId" int`
+    );
+    await queryRunner.query(
+      `ALTER TABLE "funding" ADD CONSTRAINT "FK_813982d9605de30735ef677f572" FOREIGN KEY ("lastReportingPeriodId") REFERENCES "reporting_period"("id") ON DELETE NO ACTION ON UPDATE NO ACTION`
+    );
+    await queryRunner.query(
+      `ALTER TABLE "funding" ADD CONSTRAINT "FK_8b0c90027c72688977648e07d05" FOREIGN KEY ("firstReportingPeriodId") REFERENCES "reporting_period"("id") ON DELETE NO ACTION ON UPDATE NO ACTION`
+    );
+
+    await queryRunner.query(
```
JordanGuinn, line 38: I think this `CREATE TABLE` command needs to come before the two constraint additions above it, as those directly reference the table being created in this command.

JordanGuinn:
I've updated the contents of the ticket as well, but just to make sure we don't lose sight of a conversation @vonbearshark & @jlhogan recently had around expected data transformations, I'll outline as much here as well:

For enrollments with a **single** funding, we'll want to:
   - Copy `enrollment.entry` into `funding.startDate` column
   - Copy `enrollment.exit` into `funding.endDate` column
 
For enrollments with **multiple** fundings, we'll want to
   - Copy `reportingPeriod.startDate` of **enrollment's first reporting period** into `funding.startDate` column
   - Copy `reportingPeriod.endDate` of **enrollment's last reporting period** into `funding.endDate` column

src/entity/Enrollment.ts
```diff
@@ -86,23 +87,17 @@ export class Enrollment implements EnrollmentInterface {
     cascade: ['update', 'soft-remove'],
     eager: true,
   })
-  @ValidateNested({ each: true })
-  @FundingAgeGroupMatchesEnrollment()
-  @IsNotEmpty()
+  // Add enrollment start date to fundings to enable funding start date validation
+  // (Not a real condition! Just hack to pre-process)
```
bamader, line 91: Do we need to delete this object change after the validations are done processing? I've seen that pattern applied in other places where we've set data as part of validating.

src/entity/Enrollment.ts
```diff
@@ -86,23 +87,17 @@ export class Enrollment implements EnrollmentInterface {
     cascade: ['update', 'soft-remove'],
     eager: true,
   })
-  @ValidateNested({ each: true })
-  @FundingAgeGroupMatchesEnrollment()
-  @IsNotEmpty()
```
bamader, line 91: Do we not want the `IsNotEmpty` validation anymore?

src/migration/1624998391756-UpdateFundingRemoveReportingPeriod.ts
bamader: Isn't the `down` operation usually just the reverse of the `up` though? Not at all saying that that's right, just my observation. And if it just the reverse of some auto-generated stuff, I would certainly _hope_ that typeorm knows enough to not screw itself up 😅

bamader (approved):
Code still looks mostly good to me. I know there's been a lot of changes, but I don't see anything breaking in here, just a few comments throughout. Seems like this isn't quite ready for validation, but let me know when it is and I'll fling myself and my ungodly test suites at it!

JordanGuinn:
# Validation Plan

## ETL

## Upload

## Create Child

## Edit Child

## Batch Edit Child

## Withdraw Child

src/migration/1624998391756-UpdateFundingRemoveReportingPeriod.ts
JordanGuinn: If this was the reverse of the `up`, then the `CREATE TABLE` operation would be the very first command executed, as it's the _last_ command executed in the `up` sequence.  I'm also not sure how much of this migration was auto-generated, but regardless I'm **very** confident that MSSQL will not allow you to reference the column of a table that doesn't exist 😅

At any rate, I'm now one of the folks assigned this ticket, so I'll update myself 😁

src/entity/Enrollment.ts
JordanGuinn: We can certainly add it back, but [according to the docs](https://github.com/typestack/class-validator#validation-decorators) it only appears to check for falsy values - so this may not actually be doing anything given `fundings` is an array?

src/entity/Enrollment.ts
JordanGuinn: I don't quite think I follow?

bamader: Check line 29 in `src/entity/decorators/Funding/fundingBeginsAfterEnrollmentEntry` to see what I mean. I thought this was more widespread (I think at one point it was but maybe we pared it down), but I think there are some schemas / validations that add properties to entities that we then had to manually delete after the validation runs. Might not apply here but wanted to check.

JordanGuinn: Ahhhhh, okay I follow.  I don't think there are any immediate consequences to attaching a funding's enrollment to it - it's _dumb_, but I don't believe it to be particularly problematic

src/index.ts
vonbearshark: While we're leaving notes, we also would like to inform providers of this change

Issue #1319, opened 2021-05-07
Admins can set user site permissions in the user edit view


JordanGuinn:
## Background
Once OEC admins have the ability to look up users by email, they should then be able to add, remove and set site-specific permissions to those same users for any and all organizations they create.

This is pretty much just the inverse of https://app.zenhub.com/workspaces/ece-dev-board-5ff506028d35f30012e0e937/issues/ctoec/data-collection/1311.

_**NOTE:** When All Sites is selected, it should actually add an entry in the OrganizationPermission table (which inherently provides access to all sites within that organization)._

## Acceptance Criteria
- OEC admins can **add** organizations found from org search to users being edited
   - Default permissions should be all sites
- A user/site component is added above the user search component, based on the designs below
- OEC admins can **remove** organizations previously added to users
- OEC admins can specify **specific sites** for **org users** they've added
   - List is just all sites within the organization
- **Only** when the `Save changes` button is pressed are these permissions saved to the DB

## Designs
[Full Figma files can be found here.](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=2%3A2)

#### Edit View
![Screen Shot 2021-05-07 at 9.23.06 AM.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/f4c9acf6-1805-4792-8b49-92d102c3c7d1)

PR #1392, opened 2021-07-06
1319 add perms

bamader:
## Background
This changeset gives admin users the ability to edit user permissions from the edit user page in the admin view.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1319

## Validation Plan
From the admin view, go to edit user. Then:
- Orgs that are already permissioned appear for the user
- Can add an org via the search bar and have it appear
- Can remove an org without disrupting comparable fields for other orgs
- Default permissioning when adding a new org is all sites (to represent org user)
- Can change sites for orgs listed under the permissions section
- Can save changes using the submit form button

## Automated Testing

## Additional Context
The pieces of state that operate the front-end are kind of inelegant. They do what they're supposed to, but it's not pretty. Any thoughts on changing the structure or state manipulation are very welcome. I don't know how much of this is avoidable, given how we've structured our entities and our form components. This all definitely works, it just may not be the most "react-y" way of doing things.


src/controllers/users.ts
```diff
@@ -52,16 +97,66 @@ export const getUsersByEmail = async (email: string) => {
...
+    }
+  });
+
+  await getManager().transaction(async (tManager) => {
+    await tManager.delete(OrganizationPermission, { user: foundUser });
+    await tManager.delete(SitePermission, { user: foundUser });
+    newOrgPerms.forEach(async (newOrgId) => {
+      const newPerm = tManager.create(OrganizationPermission, { user: foundUser, organizationId: newOrgId });
+      await tManager.save(newPerm);
+      foundUser.orgPermissions.push(newPerm);
```
JordanGuinn, line 153: Is there a particular reason we're pushing the org and site permissions directly to the user after writing them to row?  We don't appear to do anything with them (the controller returns a 200 with no response body), so seems like there's no actual purpose for these lines

JordanGuinn: And if THAT'S the case, we could do away with the implicit `create` call before `save`!!

src/routes/organizations.ts
```diff
@@ -34,3 +36,13 @@ organizationsRouter.get(
     res.send(organizations);
   })
 );
+
+organizationsRouter.get(
```
JordanGuinn, line 40: We could probably just bake this logic into the already existing `getOrganizations` API, no?  There really should only ever be a single route for something like `LIST organizations`, and that route itself incorporates all the filterable options we support

src/controllers/users.ts
```diff
@@ -29,6 +63,17 @@ export const addSiteAndOrgDataToUser = async (user: User) => {
   user.organizations = organizations;
 };
 
+/**
```
JordanGuinn, line 66: Would you mind removing this comment and spinning up a ticket for this?  This is a good callout, and I don't want to lose sight of it 👌🏽👌🏽

src/controllers/users.ts
```diff
@@ -1,6 +1,41 @@
...
+import { uniq } from 'underscore';
+
+/**
+ * Fills out the organization and site ID fields on a user object
+ * by mapping any linked permissions to their respective IDs.
+ * SiteIDs are comprehensively generated by supplementing the user's
+ * sitePerms with all sites under the purview of organizations the
+ * user can also access.
+ */
+const mapUserPermsToIds = async (user: User) => {
```
JordanGuinn, line 13: Any chance instead of copying this from the `authenticate.ts` middleware file, we can pull that logic out into it's own function and reference it in both places?  Something something DRY 😅

client/src/containers/Users/OrgSearchBar.tsx
```diff
@@ -0,0 +1,76 @@
...
+              <div className="grid-row grid-gap">
+                <div className="tablet:grid-col-4">{o.providerName}</div>
+                <div className="tablet:grid-col-2">
+                  <Button
+                    appearance="unstyled"
+                    text={user?.organizations?.find(org => org.providerName === o.providerName) ? "User already has access to this organization" : "Add organization"}
+                    onClick={() => {
+                      setUser(u => {
+                        const updatedUser = produce<User>((user || {}) as User, draft => {
+                          console.log(o.sites);
```
JordanGuinn, line 58: nit: console log

client/src/containers/Users/EditUser.tsx
```diff
@@ -121,51 +162,153 @@ const EditUser: React.FC = () => {
...
+                        </div>
+                      </div>
+                      {user.organizations.map((userOrg, idx) => {
+                        // These will always generate a match because it's just the full
+                        // version of the orgs the user can access
+                        // Never _actually_ undefined
+                        const fullOrg = orgsOnPage.find((o) => o.providerName === userOrg.providerName);
+                        const fullSites = (fullOrg?.sites || []).map((s) => ({ ...s, label: s.siteName }));
+                        const initSites = (fullSites).filter((fullSite) => user.sites?.some((uSite) => uSite.siteName === fullSite.siteName));
+                        console.log(initSites);
```
JordanGuinn, line 236: nit another console log

client/src/containers/Users/EditUser.tsx
```diff
@@ -121,51 +162,153 @@ const EditUser: React.FC = () => {
...
+                                  className="marginless-button"
+                                  text="Remove organization"
+                                  onClick={() => {
+                                    setUser(u => {
+                                      const updatedUser = produce<User>((user || {}) as User, draft => {
+                                        draft.organizations = draft.organizations?.filter(
+                                          o => o.providerName !== userOrg.providerName
+                                        );
+                                        draft.sites = draft.sites?.filter((s) => fullSites.every((fs) => fs.siteName !== s.siteName));
+                                        console.log(draft.sites);
```
JordanGuinn, line 289: another console log

client/src/containers/Users/EditUser.tsx
```diff
@@ -121,51 +162,153 @@ const EditUser: React.FC = () => {
...
+                        console.log(initSites);
+                        let updatedUser: User;
+                        return (
+                          <LoadingWrapper loading={!fullSites.length}>
+                            <div className="grid-row grid-gap margin-bottom-1">
+                              <div className="tablet:grid-col-3 site-perm-field">
+                                {userOrg.providerName}
+                              </div>
+                              <div className="tablet:grid-col-6" key={`${userOrg.providerName}-${idx}-multiselect-${Date.now()}`}>
+                                <MultiSelect
```
JordanGuinn, line 245: IT LIVES!!!!

src/controllers/users.ts
bamader: The permission mapping of the user's perms to IDs does actually use these fields on the user object (see the `mapUserPermsToIds` function body on line 17). So I believe we have to handle them the same way we handle other nested entities, which is to say, create them, put references to their object bodies into the original object, and then save the whole thing. It's the same way for e.g. families, we create income determinations in the DB, stuff all those objects into `family.determinations` via `.push(newDet)`, and then save the `family` at the end of it all. So alas, I think you can't get away with demolishing `create` entirely ;)

src/routes/organizations.ts
bamader: Yep we totally could! I just threw in a query string and the equivalent parsing logic to handle whether we get one set of orgs or another (by ID).

src/controllers/users.ts
bamader: Yup sure thing!

src/controllers/users.ts
JordanGuinn: I don't think I follow.

The creation of an `OrganizationPermission` or `SitePermission` row _itself_ **implicitly** defines its relationships with users and organizations, so I'm having a hard time wrapping my head around what including them in the user update as well actually _does_.  Like the only possible scenarios that I could see existing here are:

- Updating a user with permissions included doesn't actually do anything, in which case we shouldn't include them
- Updating a user with permissions included DOES actually update the corresponding permissions, in which case we shouldn't explicitly update the permissions separately at all

Either way, it seems like we're executing an unnecessary step; just not sure which one.

src/controllers/users.ts
bamader: Moved to its own util

src/routes/organizations.ts
JordanGuinn: Yep!  So it should be super easy to just move that check to the default `LIST organizations` route 🙂

bamader: Is that not what I did? I thought I moved it where it was desired.

src/controllers/users.ts
bamader: I don't _think_ it's unnecessary based on the rest of the codebase, but I could be wrong. I think the family and income determinations relationship makes this clear. An `incomeDetermination` entity does have a field for `familyId` which points back to the family its associated with. However, this `family` entity also has a field for `incomeDeterminations` which stores those nested objects. If we were to just create `incomeDeterminations` and assign them the family's ID without storing the object in `family.incomeDeterminations`, then the det will exist but I'm like 90% sure it won't be returned with the `family` object when using the `{ relations: [] }` option in TypeORM. So based on everywhere else, I think we do have to explicitly store the objects on the parent entity to get them to properly retrieve in the future.

src/routes/organizations.ts
JordanGuinn: Ah, no, you did - I commented preemptively.  Don't mind me 😅🤦🏽‍♂️

src/controllers/users.ts
JordanGuinn: My suspicion (and **strong** hope), is that fields like `family.incomeDeterminations` and `user.orgPermissions` don't actually _exist_ anywhere at the data level, but are rather just a reflection of the relationship between those data points, and define the format in which that joined data is supplied when pulled from the associated table.  It makes sense that `family.incomeDeterminations` is omitted when you specify `{ relations: [] }`, as you're _literally_ telling it to do that.

I would be _thoroughly upset and extremely confused_ by TypeORM if this wasn't the case, honestly 😅😅😅

bamader: Prepare to be thoroughly upset and extremely confused then 😭 . I just finished some testing on trying out both of the bullets you listed (i.e. not pushing the perms onto the user but still making/saving them, and pushing perms onto the user and saving the whole user without individually creating the perms) and they fail catastrophically. In the case of the former, I'm flagged with a `cannot insert NULL into column userId` that I can't get around no matter what I do, and in the case of the latter, the operation _appears_ to proceed correctly, but in reality **all** of the user's permissions are just outright deleted. Therefore, 1) saving a user object with perms that haven't been explicitly created doesn't write or save them to the DB, and 2) saving permissions without pushing them onto the user doesn't write the `userId` field into the requisite table. Both of these operations are required to make this work 😕 .

JordanGuinn: Let's pair on this tomorrow

JordanGuinn (approved):

nickbristow (approved):

nickbristow:
It looks the main functionality is all there and working.

Couple of notes from validation:

Screenshot_7_8_21__3_37_PM

Screenshot_7_8_21__3_38_PM
This can probably be styled like text?

- The search for org page sub form should have its own submit button, the whole form submits when you hit enter. Try this https://www.impressivewebs.com/html5-form-attribute/

- It's unclear that the user's changes have been made. We were brainstorming an alert to the user when they try to navigate with unsaved changes. Maybe try`pageHide`

- Cancel button doesn't reset user permissions changes

bamader:
@nickbristow Just clarifying the outstanding steps to make sure I've got my plan for tomorrow:
- Remove the top margin on the Add Organization button to make it line up with the org name
- Change the design of the "User already has access" text--you're thinking plain text here, with no background? I believe Emily looked at this in design review on QA and said it was okay, so just want to be sure I know what you're thinking before changing it
- I'll give that link you attached a try to see if I can make the form not submit when hitting enter; the idea was never to have the org search be its own form, but based on the attached link, _treating_ it like its own form might actually help override that "hit enter to submit" behavior; do you know of anywhere else in the app where we've overridden that behavior?
- As far as an alert, there should definitely be an alert already on the page--when you submitted data in validation, did you not see a green success banner on the page saying "User has been updated" or "Your changes have been saved" or something like that?
- For the cancel button, the desired functionality is to discard all changes, resetting permissions back to what they initially were? What if we just redirected the admin user back to the table of users to edit?

bamader:
Working through these now, but I'm definitely seeing an alert after saving. 
Screen Shot 2021-07-09 at 9 18 06 AM

JordanGuinn:
@bamader Nick is moreso referring to the inverse - like if you make changes _prior_ to saving, it's not immediately clear that those changes are still pending.  So users might be confused if they make changes, _don't_ click `Save`, navigate away, navigate back, and see that their changes weren't applied.

One solution we were considering for this was using the `pageHide` browser event to detect whenever a user is attempting to navigate to a new page and, if they have pending changes, throw up an alert letting them know as much.

JordanGuinn:
Also re the cancel button - right now it currently resets the other user fields when clicked (first name, last name, etc), so that behavior should probably just be extended to the user's permissions as well.

bamader:
Regarding unsaved changes: not sure I fully agree actually. In both the Edit Record and Create Record flow, we don't save any data at all if a section or page is navigated away from without explicitly hitting "Save". Yet, I don't even think we _tell_ the user explicitly that they need to hit Save to persist the changes. I think it's just been an assumption of how pages with a Save button work. If we want to change the functionality and provide an alert when navigating away, I would propose opening up and pointing a new ticket that accomplishes that in _all_ places in the app where it's relevant, since it seems somewhat out of scope here. We could attach it to the roster redesign, if we're so inclined, because that will touch Edit Record and potentially Create, but to me it feels like a non-blocking follow-up feature rather than core to this changeset (especially if we want to get the admin branch merged into prod and deployed live soon).

The Cancel behavior, on the other hand, is **definitely** related to this work 😂 . I'll make it so that hitting cancel resets the user to their base state (as retrieved from the DB).

JordanGuinn:
yeah that tracks, we can follow-up on that

nickbristow:
@bamader I don't think its completely necessary to warn the user that there are unsaved changes, but I do think some kind of visual clue would be a nice to have.

bamader:
All of the finds from Nick's validation should now be handled except the visual cues one (I opened up a ticket to fast follows on that). I initially thought I had a really quick solution, but it doesn't work (or at least isn't immediately obvious) because of the way our 'Form' hides the 'onChange' function. Will probably require more playing around with to make it clean (I'm guessing a 2 pointer, but don't want to preemptively estimate lol) so I'm deferring it. But everything else is good to go for a second pass before merging.

nickbristow:
The enter key on the search org field still submits the whole page, but I don't think its a blocker. 

We can get it in a `Nice to Have` follow up. 

Validation:
- [ ] Search shows orgs
- [ ] current orgs show as "User already has access"
- [ ] Add org adds org to user with all sites selected
- [ ] Individual sites can be selected
- [ ] Cancel button undoes changes done before saving
- [ ] Saving page adds org to user
- [ ] Orgs can be removed from user
- [ ] Site access from an org can be removed from user
- [ ] Selecting all sites shows "all sites selected at this organization"
- [ ] Selecting some sites shows "Some sites selected for this organization"

Validated

Issue #1338, opened 2021-05-19
Create new Carbon roster table


JordanGuinn:
## Background
The heart of the roster view right here.  Instead of a segmented set of children categorized by age group (as we have with our current roster), this new UI should simply be a [basic Carbon table](https://react.carbondesignsystem.com/?path=/story/components-datatable--basic-table) displaying **all children with an active enrollment** within the user's organization.  Additionally, the total count of applicable children should sit right above this new table, as well.

**Please note** that the following **will not be completed** as part of this ticket, but rather in future tickets:
   - [Sortable table columns](https://app.zenhub.com/workspaces/ece-dev-board-5ff506028d35f30012e0e937/issues/ctoec/data-collection/1339)
   - [Table pagination](https://app.zenhub.com/workspaces/ece-dev-board-5ff506028d35f30012e0e937/issues/ctoec/data-collection/1346)
   - Table filters

## Acceptance Criteria
- [Basic Carbon table](https://react.carbondesignsystem.com/?path=/story/components-datatable--basic-table) is created for the roster, containing the following columns:
   - `Name` (last, first)
      - An exclamation icon should sit next to the names of children **missing required info**
      - Clicking the child's name should navigate the user to the edit view of that child
   - `Age Group`
   - `Funding Type`
   - `Space Type`
   - `Site`
   - `Enrollment Start Date`
      - This date should come from the **most recent active enrollment** for the child
- **All children with an active enrollment** are shown in the roster table
- Paginated will be implemented later, so for now **all applicable children** are displayed in the table
- Full count of all children is displayed above the table

## Designs
[The full set of designs can be found in Figma.](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=86%3A348)

#### Roster Table
![Screen Shot 2021-05-19 at 10.08.14 PM.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/15fd81d5-60ea-46aa-b4bb-ff991ba6a413)

PR #1398, opened 2021-07-10
Created roster carbon table

vonbearshark:
## Background
Removed active/withdrawn and (unused) pagination from existing children query, but otherwise leveraging the existing backend, mostly. Then transforming the children data into row format on the client.

## GitHub Issue
#1338 

## Validation Plan
- [ ] Validate design changes
- [ ] Confirm no infinite pagination
- [ ] Confirm Batch Edit flow is unchanged
- [ ] Confirm Create Record flow is unchanged
- [ ] Confirm Edit Record flow is unchanged
- [ ] Confirm Withdraw Record flow is unchanged
- [ ] Confirm Delete Record flow is unchanged
- [ ] Confirm Preview Changes flow is unchanged

## Automated Testing

## Additional Context
Originally I had intended to update the existing children query to pull and transform the data in the shape the client table rows expect. However, given our server memory-based validation that's not possible, since we need to collect all the potential validation errors. 🙄 . It's not as bad as it sounds, really. If we can get the filter/pagination right it's not a big deal that we continue to validate and client-transform manageable chunks.


vonbearshark:
Honestly most of the time spent so far has been on styling. And I am pretty sure I still don't have it right.

I am pretty convinced at this point that the real pain of the local dev build comes from the client and specifically from the Sass building and specifically specifically from the USWDS Sass imports. Straight up it takes ~5 minutes right now. It makes Sass dev prohibitive. And it's unclear what the Tailwind possibilities for this project are..

vonbearshark:
Also also there's a bunch of `prettier` diffs from a fresh start. Something to be mindfull of: seems like we somehow missed that for this branch as is.

client/src/containers/Roster/Roster.tsx
```diff
@@ -76,8 +145,75 @@ const Roster: React.FC = () => {
           count={childCounts[AgeGroup.SchoolAge]}
         />
       </div>
+
+      <h3>{childRows.length} enrollments</h3>
+
+      <div className="margin-top-4">
+        <DataTable
```
bamader, line 152: Nit: Personally, I'd make this roster table its own file to reduce the length of the `Roster.tsx` file. I know we don't use it anywhere else and there isn't really a compelling modularity reason to do so, I just think it helps make things more readable and accessible. But this is a very non-strong opinion so I'm good with whatever!

bamader (approved):
LGTM! I love that we slashed out a massive chunk of a controller that probably didn't need to be there.

client/src/containers/Roster/Roster.tsx
```diff
@@ -1,32 +1,69 @@
...
+  ageGroup?: AgeGroup;
+  enrollmentDate?: string;
+  fundingSource?: string;
+  fundingTime?: FundingTime;
+  id: string;
+  missingInfo: boolean;
+  name: string;
+  site?: Site['siteName'];
+}
+
```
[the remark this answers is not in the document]
vonbearshark: I thought the same thing. That's where it started when I tried to get the server to _return_ that shape (rather than `Child[]`. So the answer is more like not right now but theoretically in the future? And imo in the meantime since this component is the only one with this concept it should live here

client/src/containers/Roster/Roster.tsx
vonbearshark: I think we're right on the border with this.. I would be in favor of keeping it here for now but moving it out if and when we start adding complexity. I would prefer to wait because I am not sure exactly where the complexity will come in--probably in filtering? But worth it to wait until then, I think

client/src/containers/Roster/Roster.tsx
```diff
@@ -36,24 +73,56 @@ const Roster: React.FC = () => {
...
-              true)
-          }
+          {pluralize(
+            'child',
+            Object.values(childCounts).reduce(
+              (a: number, v: any) => (a += v),
+              0
+            ),
+            true
+          )}
```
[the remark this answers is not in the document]
vonbearshark: Prettier was weird for me. Maybe I am goofed locally? But there were a lot of prettier diffs for me in local. Either I am off or the feature branch is..

src/controllers/children/get.ts
```diff
@@ -119,116 +112,36 @@ export const getChildById = async (id: string, user: User): Promise<Child> => {
...
-    {
-      start,
-      end,
-    }
-  );
-
-  if (skip) qb.skip(skip);
-  if (take) qb.take(take);
-
+const getChildrenQuery = async (user: User) => {
```
JordanGuinn, line 117: This is _so very satisfying_ to see 👏🏽👏🏽👏🏽👏🏽

src/controllers/children/get.ts
```diff
@@ -119,116 +112,36 @@ export const getChildById = async (id: string, user: User): Promise<Child> => {
...
-};
-
 /**
  * Counts the number of children in each age group enrolled at sites
  * associated with all of a user's organizations. No records are
  * returned, only the counts within each age group.
  */
-export const getChildrenCountByAgeGroup = async (user: User, organizationIds): Promise<AgeGroupCount> => {
-  const month = moment();
+export const getChildrenCountByAgeGroup = async (
```
JordanGuinn, line 133: Is this age group count still used in other places across the app?  If not, would be simpler to just get the count for the entire query, since that's all the roster really cares about at this point.

JordanGuinn (approved):
This was a REALLY satisfying PR to review 🔥🔥

src/controllers/children/get.ts
bamader: It's used on the roster page for showing the counts by age group. I don't think anywhere else? But I also remember us as deciding to keep this stuff here.

JordanGuinn: Ahhhhhhh that's right - we still show the counts above the roster table in this new UX.  Okay nvm

bamader:
## Validation Plan
Kind of a weird one, in that logging in and just _using_ the app normally should show quickly enough whether or not it works. But, now that my docker is cooperating, here's what I've got.

Create multiple users, each affiliated with a different organization (Hogwarts, Durmstrang, Academie).
- [ ] Hogwarts user's file upload correctly displays in the roster table
- [ ] Durmstrang user's file upload correctly displays in the roster table
- [ ] Academie user's file upload correctly displays in the roster table
- [ ] Hogwarts user cannot see any children associated with Durmstrang or Academie
- [ ] Durmstrang user cannot see any children associated with Hogwarts or Academie
- [ ] Academie user cannot see any children associated with Hogwarts or Durmstrang
- [ ] Deleting a child from an edit record page removes that child from the appropriate roster
- [ ] Withdrawing a child from an edit record page _does not_ remove the child, because we don't yet have that programmed

bamader:
Some sort of absolute monstrosity is happening with network requests in this PR. The roster page is invoking the `children GET` route literally multiple times per second, continuously the whole time the app is live.

vonbearshark:
Oof. My b. Guessing that I probably forgot to `useEffect(, [])` and it's just an infinite loop

vonbearshark:
Er, wait, no I didn't. Hm.. 🤔

bamader:
I thought that but it looks like all your `useEffects` have correct dependency arrays. For example, throwing a console log in the `async getchildRows()` function on the roster page, I see a non-infinite number of correct logs printed out. It looks like the skip parameter is increasing by 100 every half second?

bamader:
WAIT WTF you deleted `skip`

vonbearshark:
We want to delete skip (for now)? We weren't actually using it anyway. And we're not using that old intentional loop pagination hack thing any more, so it shouldn't be a problem just calling for the data once?

bamader:
Right, yeah, totally on board with that. Sorry, I should've been more clear. There is now not a single reference to skip when fetching children data (which is good), yet my network requests show the skip parameter clearly in calls to the api with an ever increasing value (bad). Gonna compose down and clear all the caches and stuff and see if that fixes the issue.

vonbearshark:
Oh, interesting. So if the old `RosterContext` is being called or brought in anywhere in the pipeline it will kick that off. And now that I killed the `skip` on the backend then it will go infinitely, I think. I just don't know how it called be brought in at all if that's all been moved? Or maybe the context manager hasn't?

bamader:
Great call. Turns out _all_ of our operations adjacent to the roster table (uploading a file via the preview screen, edit record, etc.) all invoke the roster context. And the roster context invokes `usePaginatedChildData`. The problem of infinite querying doesn't happen until I try to either upload a file or edit a record or something similar, and then it just takes off.

vonbearshark:
> all of our operations adjacent to the roster table invoke usePaginatedChildData

Yeap, that will do it. We can just get rid of `usePaginatedChildData` entirely now. I can poke around there and see if it's that simple

vonbearshark:
Added a commit that addresses the infinite pagination issue and also expands the scope of this PR a bit.

I moved the old `RosterContext` to `RosterContextOld`. Also moved necessary hooks/code that would otherwise be deleted to `RosterOld` to preserve that. All should be deleted with `RosterOld` after this effort

Ditched infinite `swr` for normal `swr` (that will be independently useful). Removed pagination and querying, including all references to org IDs in remaining roster cases.

client/src/index.tsx
```diff
@@ -16,17 +16,16 @@ import * as serviceWorker from './serviceWorker';
...
-      ReactGA.initialize(gATrackingId);
-      ReactGA.pageview(window.location.pathname + window.location.search);
-    }
-  });
+getConfig('GoogleAnalyticsTrackingId').then((gATrackingId) => {
+  if (!!gATrackingId) {
+    ReactGA.initialize(gATrackingId);
+    ReactGA.pageview(window.location.pathname + window.location.search);
+  }
+});
```
vonbearshark, line 28: (just formatting change)

vonbearshark:
Added a pass at a validation plan. Essentially, we need to make sure all the Roster functionality works as it currently does with the components that remain, since these changes likely touched them.

And this will need another re-review

nickbristow (approved):

bamader:
## Validation Plan

Roster operations and general functionality
- [ ] Validate design changes
- [ ] Confirm no infinite pagination
- [ ] Confirm Batch Edit flow is unchanged
- [ ] Confirm Create Record flow is unchanged
- [ ] Confirm Edit Record flow is unchanged
- [ ] Confirm Withdraw Record flow is unchanged
- [ ] Confirm Delete Record flow is unchanged
- [ ] Confirm Preview Changes flow is unchanged

Permission-specific functions (ensuring that queries still work the way we expect them to work). For this section, we create multiple users, each affiliated with a different organization (Hogwarts, Durmstrang, Academie).
- [ ] Hogwarts user's file upload correctly displays in the roster table
- [ ] Durmstrang user's file upload correctly displays in the roster table
- [ ] Academie user's file upload correctly displays in the roster table
- [ ] Hogwarts user cannot see any children associated with Durmstrang or Academie
- [ ] Durmstrang user cannot see any children associated with Hogwarts or Academie
- [ ] Academie user cannot see any children associated with Hogwarts or Durmstrang
- [ ] Deleting a child from an edit record page removes that child from the appropriate roster

bamader:
Subjected this to another round of validations. Results were mostly positive. The changes look like everything is good to go except Create Record and a bug I found involving deleting fundings. See my validation post above, but I think the Create Record thing is completely fine right now and shouldn't block this, while the funding deletion thing should be addressed (and while we're at it, we should defensively code for any of the table fields missing, if we haven't already). Once that's addressed, I'd call this good from my perspective.

bamader:
Okay some updates on what's winding up being more complicated than I thought:

Resolved bugs:
- Fixed the bug where the roster table crashes if any cell used to pull info from is missing that property
- Found another separate bug where the Enrollments and Funding page of Edit Record would crash when refreshing a page with a child who has an unfunded enrollment, _or_ when first deleting / changing a funding on the page. The error was `uncaught type exception: cannot read property 'fundingSpace' of undefined`. The fix was tricky to find but easy to implement, just changing the nullish coalescing operators in the RosterContext to force the `fundings` array to be "defined", even if it's empty, so that we could extract an `undefined` funding space from an object instead a nullish...well, thing.

Unresolved bugs: 
- Found another bug (or at least something I believe to be a bug): when a user goes to edit record and saves changes to a child, rather than just calling the `RosterContext`'s `childToRow` function once as part of `updateChildRows`, the `childToRow` function is called on _every_ child in the user's roster. This means that when doing EditRecord things for 1 child, API calls are being generated for every other child on the roster. This may not be an issue for small scale rosters that can afford the API calls happening in the background, but for rosters with lots of kids, this will be an issue that slows down EditRecord. To see all the calls happening, just throw a `console.log` at the top of `childToRow` and print out, say, the `site` property. You'll see it change as many times as there are kids in the roster. Desired behavior should be just calling the `childToRow` function only on the child whose EditRecord page we're viewing.

I've got another meeting coming up that I have to read some stuff for, so can't fix that right now, but I'm pushing up my other fixes including a merge in of the local tests blocking `xlsx` package.

client/src/components/Forms/FamilyIncome/Fields/AnnualHouseholdIncome.tsx
```diff
@@ -33,7 +33,7 @@ export const AnnualHouseholdIncomeField: React.FC<HideErrorProps> = ({
       className="usa-label"
       // @ts-ignore
       hideSteppers={true}
-      onChange={(e) => {
+      onChange={(e: any) => {
```
vonbearshark, line 36: interesting this came up now?

client/src/containers/Roster/Roster.tsx
```diff
@@ -154,7 +154,7 @@ const Roster: React.FC = () => {
                             {childRows[i]?.missingInfo && <IncompleteIcon />}
                           </>
                         ) : (
-                          cell.value
+                          cell.value ? cell.value : ""
```
vonbearshark, line 157: nit: `cell?.value ?? ""`

client/src/contexts/RosterContex/RosterContext.tsx
```diff
@@ -34,7 +34,7 @@ export type UpdateCacheParams = {
 
 const childToRow = (child: Child) => {
   const { ageGroup, entry, fundings, site } = child.enrollments?.[0] ?? {};
-  const [{ fundingSpace }] = fundings ?? [{}];
+  const fundingSpace = fundings?.length ? fundings[0].fundingSpace : undefined;
```
vonbearshark, line 37: good catch

vonbearshark:
Appreciate your fixes! I'll see if I can't find the source of that bug. I don't quite see the relation to the changes now..

vonbearshark:
> or at least something I believe to be a bug

So there _did_ actually seem to be a bug in the `mutate` logic, where we weren't actually using the immutability correctly. I am pushing that minor change now.

However, I am not sure what you uncovered is a bug. Essentially, we are transforming the results of the `GET` to `children`, mapping `Child[]` into `ChildRow[]`. Unless I am severely missing something, this is a lightweight op--just a simple transformation, no API calls. As I mentioned in my note, ideally this would just be the response shape from the API, but the validation error and export logic we have kind of prevents this right now.

As we `mutate` and `revalidate` the response data from `useSWR` we'll also be mapping over the child records again, re-transforming them into `childRows`. Given that it's not expensive, I don't think it's really a problem. And also we should only transform what we need to render. So once pagination lands, we would just be transforming the paginated records

Issue #1337, opened 2021-05-19
Add "Upload File" and "Add Record" buttons to roster


JordanGuinn:
## Background
The logic to upload a file as well as create a new child record already exists, so all we need to do here is create the corresponding buttons and have said buttons route the user to the appropriate views.  From there, the user should then be able to add data as they could previously and have that data reflected in the roster accordingly.

## Acceptance Criteria
- `Upload File` and `Add Record` **Carbon** buttons are added to the new roster view
- User is navigated to the appropriate view upon clicking these buttons
- Child rows added through both the file upload and add record UX are reflected in the roster

## Designs
[The full set of designs can be found in Figma.](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=86%3A348)

![image.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/f0389165-fd2a-407c-98ca-074487819448)

PR #1400, opened 2021-07-14
Add buttons

bamader:
## Background
This change set adds in two action buttons to the roster page, the Add Record button and the Upload File button.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1337

## Validation Plan
Log in to the app. Should directly go to the roster page. 
- [x] Confirm that the Add Record button directs a user to the create flow. Completing the flow successfully adds the child to the roster.
- [x] Confirm that the Upload File button directs a user to the upload wizard. Uploading a file adds the children to the roster.

## Additional Context
Nope!


nickbristow (approved):
Seems fine to me, need to update snapshots

JordanGuinn (approved):

vonbearshark (approved):

src/routes/children.ts
```diff
@@ -6,6 +6,7 @@ import {
   ApiError,
 } from '../middleware/error/errors';
 import * as controller from '../controllers/children';
+import { parseQueryString } from '../utils/parseQueryString';
```
vonbearshark, line 9: is this used?

client/src/containers/Roster/Roster.tsx
```diff
@@ -14,20 +14,40 @@ import pluralize from 'pluralize';
...
 import { AgeGroupCount } from '../../shared/payloads/AgeGroupCount';
 import { apiGet } from '../../utils/api';
+import RosterContext from '../../contexts/RosterContex/RosterContext';
 import { getH1RefForTitle } from '../../utils/getH1RefForTitle';
 import { RosterSummaryCard } from './RosterSummaryCard';
+import { Divider, Button } from '@ctoec/component-library';
+import { AddRecordButton } from '../../components/AddRecordButton';
+import { useAlerts } from '../../hooks/useAlerts';
+
+export interface ChildRow extends ObjectWithValidationErrors {
```
vonbearshark, line 33: This type should be in RosterContext now, I think, right? We should just be able to pull it from there

src/routes/children.ts
bamader: Nope! Merging in the base feature branch brought in some conflicts and kruft; thought I cleaned it all up but looks like a few stragglers survived.

client/src/containers/Roster/Roster.tsx
bamader: Yup it is! While I was developing, I had it based on your branch, but it looks like the merge didn't take it out of the file.

bamader:
Okay well I thought I was unblocking this / finishing up some merge conflict resolution, but there ended up being a _real_ fun new issue. We got hit with a package audit on the back end for `xmldom`. Okay, no problem, there's an issue in all versions < 0.7.0 (https://www.npmjs.com/advisories/1769). We'll just upgrade. Wait, version `>= 0.7.0` doesn't exist and isn't listed under the version numbers for available `xmldom` packages (https://www.npmjs.com/package/xmldom). Okay, hmm, let's try checking out the specific github issue the advisory recommends for publishing info on this supposed version. Wait issue #271 doesn't exist either (https://github.com/jindw/xmldom/issues). Wut. 

TLDR: I have no idea how we're supposed to upgrade to a non-existent package version.

JordanGuinn:
Followed the validation plan outlined above by @bamader (which is sufficient given the limited scope of this PR) and found no issues, so I'd say this is ready to merge.

As far as the `xmldom` issue goes, it looks like they published a 0.7.0 release a little over a week ago, but for whatever reason, [haven't been able to publish said version to NPM](https://github.com/xmldom/xmldom/issues/271).  As such, my inclination is to hold off a little bit longer to see if that version eventually does get published, as I'm not immediately enthusiastic on pinning this dep directly through GitHub if we can avoid it.

I also moved the test steps ahead of the audit steps in the `local-tests` CircleCI workflow, so that we can know if our test suites are green at the very least, for the time being.

PR #1418, opened 2021-08-11
Add Weekly Ship SQL Script

JordanGuinn:
## Background
Adds a SQL script that I put together in order to gather various data points used in the weekly ship.  Figured it's time to capture in source control so that I'm no longer the only one that can run it 😅

## GitHub Issue
N/A

## Validation Plan
N/A


bamader (approved):
Lol bout time we had something like this

Issue #1404, opened 2021-07-20
Edit User header isn't updated when user's name is changed


JordanGuinn:
## Background
A very minor issue for sure, but the top level header of the `Edit User` page (which is just the user's first and last name) isn't updated if the name is changed.

## Repro Steps
- Login to ECE Reporter as an admin user
- Navigate to the `Users` tab
- Navigate to the `Edit` view of a user
- Change the user's name
- Save your changes
- Observe the name at the top of the page

## Expected Behavior
The name header is updated to match the new name specified.

## Actual Behavior
The name header remains what it was prior to having changed it.

## Demo
https://images.zenhubusercontent.com/274936723/40136365-932a-41ae-a3e0-db6e4cb15a01/untitled.mov

PR #1420, opened 2021-08-12
Edit User header isn't updated when user's name is changed

lukenewman:
## Background
* Set the user to be the newly-saved user.
* Compile fixes (component library fixed to 1.1.21 and added a type def to `AnnualHoseholdIncome.tsx`

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1404

## Associated PRs
None

## Validation Plan
Followed the repro steps in the issue description and validated the expected behavior.

## Automated Testing


client/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@ctoec/component-library": "^1.1.21",
+    "@ctoec/component-library": "1.1.21",
```
vonbearshark, line 6: Why this change?

vonbearshark: Also there's lots of lockfile diffs looks like?

vonbearshark (commented):
Seems like the actual code change is small. Lots of `prettier` noise, it looks like. Not your fault and happened to others as well. We should investigate at some point. But as long as this style seems reasonable enough then this particular PR is fine there imo.

Would approve but waiting on confirmation for the package/lockfile changes

client/package.json
lukenewman: for context: https://skylight-hq.slack.com/archives/CM7D99AF3/p1628717533010200

JordanGuinn: Was a `yarn install` run here by chance?  Instead of a `yarn install --frozen-lockfile`?

bamader (approved):
Content-wise this does what it's supposed to. I don't know wtf is going on with `AnnualHouseholdIncome` because I swear I've made that exact change in like seven different PRs already.

JordanGuinn (approved):

bamader:
## Validation Plan
- Log into app with an admin user
- Go to edit user page for some dummy user account
- [ ] Can change user's name, and new name updates to what's on the page title
- [ ] Can add/remove orgs and change site perms as normal

vonbearshark (approved):

bamader:
For whatever reason github isn't letting me mark the validation boxes as checked, but this passes all validation tests and is ready to go!

client/package.json
JordanGuinn: Updated the lockfile to match the results of `yarn install --frozen-lockfile`

Issue #1311, opened 2021-05-05
Admins can set user site permissions in the org create view


JordanGuinn:
## Background
Once OEC admins have the ability to look up users by email, they should then be able to add, remove and set site-specific permissions to those same users for any and all organizations they create.

This is pretty much just the inverse of https://app.zenhub.com/workspaces/ece-dev-board-5ff506028d35f30012e0e937/issues/ctoec/data-collection/1319.

## Acceptance Criteria
- OEC admins can **add** users found from email search to organizations being created
   - Default permissions should be all sites
- A user/site component is added above the user search component, based on the designs below
- All users added to the organization are listed in said component
- OEC admins can **remove** users previously added to organizations
- OEC admins can specify **specific sites** for **org users** they've added
   - List of available sites should be based on those added in the `Sites` section of this view
- **Only** when the `Create Organization` button is pressed are these permissions saved to the DB

## Technical Details
The simplest solution probably incorporates the sending of a single blob of all users and their permissions to the API created here, and letting the server take it from there.  It won't have much of an impact here since all of this data is additive, but it will simplify the edit behavior when we get there (i.e. a single API call vs dealing with adding users/editing user perms/removing users all separately).

**_NOTE:_** When `All Sites` is selected, it should actually add an entry in the `OrganizationPermission` table (which inherently provides access to all sites within that organization).

## Designs
[Comprehensive FIGMA designs can be found here.](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=2%3A2)

#### Add Users
![Screen Shot 2021-05-05 at 1.14.10 PM.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/3d81559d-dd78-4207-ab6f-a3c6917fe00e)

#### User Permissions
![Screen Shot 2021-05-05 at 1.14.16 PM.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/6e2ace26-a0ca-477d-94fa-3d32c56fca48)

PR #1423, opened 2021-08-16
Allow admin to add users at org creation

nickbristow:
## Background
Admins can search for and add users to an org at time of creation. All sites are selected by default.

## GitHub Issue
#1311 

## Validation Plan
TBD

## Automated Testing


src/controllers/organizations.ts
```diff
@@ -50,6 +63,46 @@ export async function createOrganization(
   );
   newOrg.sites = createdSites;
 
+  const usersWithOrgPermissions = users.filter((user) => user.allSites);
+  const usersWithSitePermissions = [];
+  users
+    .filter((user) => !user.allSites && user.newSites)
+    .forEach((u) => {
+      u.newSites.forEach((site) => {
+        const matchedSite = createdSites.filter((s) => s.siteName === site)[0];
```
vonbearshark, line 72: nit: `find` over `filter` at `[0]`?

src/controllers/organizations.ts
```diff
@@ -50,6 +63,46 @@ export async function createOrganization(
...
+    )
+  );
+
+  await Promise.all(
+    usersWithSitePermissions.map(
+      async (u) =>
+        await getManager().save(
+          getManager().create(SitePermission, {
+            siteId: u.siteId,
+            userId: u.userId,
```
vonbearshark, line 100: nit: deconstruct the props instead of `u` and just pass them into the object param?

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -182,6 +190,10 @@ const CreateOrg: React.FC = () => {
     setNewSites((current) => removeItem(current, idx));
   };
 
+  const siteUpdate = () => {
+    setNewSites((current) => [...current]);
+  };
```
vonbearshark, line 195: why is this necessary?

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -195,6 +207,34 @@ const CreateOrg: React.FC = () => {
...
+      return 'All sites selected at this organization';
+    }
+    return 'Some sites selected for this organization';
+  };
+
+  const [users, setUsers] = useState<UserWithSites[]>([]);
+
+  useEffect(() => {
+    console.log(users);
+  }, [users]);
```
vonbearshark, line 236: 👁️  rm?

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -257,20 +298,129 @@ const CreateOrg: React.FC = () => {
...
+                      ...s,
+                      label: s.siteName,
+                    }))}
+                    selectionFeedback="fixed"
+                    initialSelectedItems={initSites}
+                    // item is never actually undefined here--will always have a name
+                    onMenuChange={(open: boolean) => {
+                      if (!open && updatedUser) {
+                        setUsers((currentUsers) => {
+                          return currentUsers.map((u) => {
```
vonbearshark, line 354: here and elsewhere: isn't `currentUsers` the same as `users`? What's the benefit to the callback syntax in this hook?

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -257,20 +298,129 @@ const CreateOrg: React.FC = () => {
...
+          const fullSites = (newSites || []).map((s) => ({
+            ...s,
+            label: s.siteName,
+          }));
+          const initSites = fullSites.filter(
+            (fullSite) =>
+              user.allSites ||
+              user.newSites?.some((uSite) => uSite === fullSite.siteName)
+          );
+          let updatedUser: UserWithSites;
```
vonbearshark, line 330: should this be tracked in state?

vonbearshark (approved):

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -257,20 +298,129 @@ const CreateOrg: React.FC = () => {
...
+          return (
+            <>
+              <div className="grid-row grid-gap margin-bottom-2">
+                <div className="tablet:grid-col-3 site-perm-field">
+                  {user.email}
+                </div>
+                <div
+                  className="tablet:grid-col-6"
+                  key={`${newOrgName}-${idx}-multiselect-${Date.now()}`}
```
vonbearshark, line 339: Why the dated key?

client/src/containers/CreateOrg/CreateOrg.tsx
nickbristow: I don't think so, this is the pattern @bamader used, and its not really  needed in state

client/src/containers/CreateOrg/CreateOrg.tsx
nickbristow: Ahh, this was copy pasted, updated.

client/src/containers/CreateOrg/CreateOrg.tsx
nickbristow: Yes, i don't know if there is any benefit of passing a function vs passing the map directly. We've just been doing it this way. Happy to change if you think we should.

client/src/containers/CreateOrg/CreateOrg.tsx
nickbristow: 😳

client/src/containers/CreateOrg/CreateOrg.tsx
nickbristow: I knew this was going to come up. So this is a bit of a hack. The multiselect of site names in the user permissions tables needs to be updated if the user changes the site name. The state of sites is stored in an array. Unless the array changes or the size of the array changes, it won't re render the page. So I have this bound to the blur of the site name which resets the site state and re renders the page. 

Im open to other suggestions if you know how to make this work better.

src/controllers/organizations.ts
nickbristow: We updated this

client/src/containers/CreateOrg/CreateOrg.tsx
vonbearshark: I guess it's fine, it's just a little more verbose and non-standard as is

client/src/containers/CreateOrg/CreateOrg.tsx
vonbearshark: Ah I just noticed we're creating this for each user. I guess if it works then it's okay 🤷🏻

client/src/containers/CreateOrg/CreateOrg.tsx
vonbearshark: I think I follow...

The array should be changing at every `setNewSites` operation already, if we're using the array spreading to create new arrays every time, right? And it looks like we are? (Though I didn't check every call site) Unless you mean there's a separate state dependency? And then we could use `useEffect`?

client/src/containers/CreateOrg/CreateOrg.tsx
vonbearshark: Maybe just double-check it wasn't used. Like, this does do _something_. I don't think it should be doing what it's doing, but it is not exactly a no-op

vonbearshark: It seems like it's essentially forcing a re-render every time it's being considered in React, maybe? Or maybe it's just trying to do that?

client/src/containers/CreateOrg/CreateOrg.tsx
bamader: Yeah no need for state here because the objects wind up just tracking all their own properties and changes. This shouldn't have any problems executing.

client/src/containers/CreateOrg/CreateOrg.tsx
bamader: I could be wrong, but I _think_ you might actually need to do it the way Nick has it set up. React gets funny when you try to modify state that's a list, particularly when it comes to triggering renders and page updates. I think just passing/modifying the map will treat it as the same object, which won't force any updates to other dependent components on the page, whereas returning a callback map will actually generate a new object for state and so that'll re-trigger things. Like I said, could be wrong, but when it comes to modifying properties within state arrays I do typically opt for verbose and forcing.

client/src/containers/CreateOrg/CreateOrg.tsx
```diff
@@ -286,16 +430,30 @@ const CreateOrg: React.FC = () => {
...
                     <Button
                       appearance="unstyled"
-                      text="Add user"
+                      text={
+                        users.some((_u) => _u.email === u.email)
+                          ? 'Added'
+                          : 'Add user'
+                      }
+                      disabled={users.some((_u) => _u.email === u.email)}
                       // TODO: Add functionality to actually add this user to the org in later ticket
```
bamader, line 439: Nit: can delete this comment now!

client/src/shared/payloads/index.ts
```diff
@@ -4,3 +4,4 @@ export * from './ChangeEnrollmentRequest';
 export * from './ChangeFundingRequest';
 export * from './WithdrawRequest';
 export * from './TemplateMetadataResponse';
+export * from './CreateOrganizationRequest';
```
bamader, line 7: +1 for making this a payload 😎

bamader (approved):
This LGTM! I'll put it on my agenda to validate either tonight or first thing tomorrow.

client/src/containers/CreateOrg/CreateOrg.tsx
vonbearshark: > I think just passing/modifying the map will treat it as the same object

Specifically I mean `setUsers([...users.map( // etc... )])`. So still creating a new array

bamader:
## Validation Plan

- Log into app with admin user and navigate to create org page
- [ ] Org create sanity checks are still in place (still need a unique name, titleI, etc.)
- [ ] Can search for users on page
- [ ] Can add a user on page and give them org level permissions
- [ ] Can add a user on page and give them some site level perms
- [ ] Adding/removing users doesn't change the permissions assigned to other added users
- [ ] Adding or updating a site after adding a user properly reflects in the user site perms drop down

bamader:
Validations are complete for this PR, no glitches or errors found. The only thing that might be worth documenting (again not as an error, just as a way the experience flows) is that if you add a site to the create org page after there are users already there, the page adds this site automatically to each user's permissions list. At first I didn't like this approach, but after thinking about it I actually think it makes a lot of sense. If an admin user is going to add users at the org create step, I feel it's most likely that they're adding an org-level user, which means the added user _should_ have permissions everywhere (rather than forcing the admin user to manually click through everything in the dropdown). Unless anyone else objects to this behavior, this gets the ready to go from me.

TLDR: One piece of behavior that I think is good that others may not; otherwise this is ready to merge.

JordanGuinn:
I agree personally - I think that default behavior for sites added to organizations should be all users receive access.  But we can wait and let general usage dictate that 👌

JordanGuinn (approved):

PR #1428, opened 2021-08-18
Allow for application startup from local machine, as well as Docker

JordanGuinn:
## Background
This PR will allow for local application setup either entirely through Docker (as is currently the case) OR on your local machine as well.  I've found that spinning up the React app and Express server directly on my machine has removed a lot of the development struggles others have experienced.

Also, there's no longer a need to explicitly move `winged-keys` to the `data-collection` directory.  Just start it up wherever it is on your machine.

## GitHub Issue
N/A

## Validation Plan
- [x] Confirm I can start the application locally entirely from Docker
- [x] Confirm I can start the application locally, with the client and server on my actual machine
- [x] Confirm I can still deploy to non-prod-like environments without issue
- [x] Confirm I can still deploy to prod-like environments without issue

## Automated Testing


bamader (approved):
Excited to try this outside of docker

JordanGuinn:
@bamader If you try that out, you should **definitely** do a full `docker-compose down` and remove all previous containers you had beforehand

vonbearshark (approved):
LGTM! I was able to test the non-Docker flow and works for me, even on the M1

This makes it much more clear what's going on locally. The frontend build is still taking 5min+ for me pretty consistently (from fresh start, not from live reload). That said, this does kill the possibility of the kind of nondeterministic/race condition-y false start that would silently fail forever. So, definite improvement!

vonbearshark:
Also worth mentioning that I would imagine another fresh M1 would need more docs to get the non-Docker working, since independent of this project Node/Homebrew installs are still pretty borked for M1. I just happened to sort that all out already. And given we still have docker it's not _required_

JordanGuinn:
Agreed - I can add something to the README to capture as much.  Although hopefully by the time anyone else would get to that, those install issues have been addressed 😅

.circleci/config.yml
```diff
@@ -89,8 +89,7 @@ jobs:
       - run:
           name: Run client tests
           working_directory: client
-          # skip pre-flight checks to avoid the parent-directory node_module issues
-          command: SKIP_PREFLIGHT_CHECK=true yarn run test
```
vonbearshark, line 93: Looks like this is failing `local` rn

bamader: Yeah don't we want the skip_preflight_check flag to be true?

JordanGuinn: I thought had fixed this - the issue was that the back-end version of jest didn't match the version used by `react-scripts`, so I just updated them to match (and confirmed it worked locally).  But it now looks like CircleCI has the old version cached somewhere on the build image 😭 So for the sake of getting this unblocked, I'll add it back now and address later

vonbearshark: lol yeah I just noticed that 🙃

bamader: How tf this get cached like that 🤦

JordanGuinn: The CircleCI node orb caches `node_modules` by default so that it doesn't have to go fetch them every time you want to install.  Otherwise each run would take like 45-60 mins 😅

Issue #1334, opened 2021-05-19
Add search to roster


JordanGuinn:
## Background
Users should have the ability to search for child by **first and last name** from within their roster view, and get a subset of children based on that search.  Specifically, the search should operate as a substring match against the child's **entire name** (i.e. the child's first and last name joined together with a space).

Let's say, for example, that the following children exist within in our roster:

- `Alexander Hamilton`
- `James Vanderhaus`
- `Derek Alhambra`
- `Chad Sanders`

Assuming the above, a roster search with the text `der ha` should **only include** `Alexander Hamilton` as an applicable result.

## Acceptance Criteria
- Search bar is added to the roster view, with a clickable magnifying glass button on the end
   - Search button **cannot** be clicked unless **at least one character** has been entered
- [A computed column for `name`](https://docs.microsoft.com/en-us/sql/relational-databases/tables/specify-computed-columns-in-a-table?view=sql-server-ver15) is added to the `Child` table, joining first and last name with a space
   - Should just be a simple DB migration
- When the search button is clicked, search string is included in the roster query and query is re-run
   - _If trivial_, hitting `Enter` should **also** initiate search
- Search should be **case-insensitive**
- Search should be a **substring match against first/last name, joined by a single space and treated as a single string**

## Technical Specs
In order to search on first and last name joined together, we'll need to [create an additional computed column](https://docs.microsoft.com/en-us/sql/relational-databases/tables/specify-computed-columns-in-a-table?view=sql-server-ver15) in the `child` table based on the combination of those two fields (which will require a migration).  From there, we should be able to perform standard SELECT queries with wildcarded `LIKE` clauses directly against that new field.

## Designs
[The full set of designs can be found in Figma.](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=86%3A348)

![image.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/f0389165-fd2a-407c-98ca-074487819448)

JordanGuinn:
@jlhogan @Derekskylight @ehuang525 Got a few questions on how we envision roster search working:
- We only want to search against **first name** and **last name**, correct?
   - If so, how do we envision that working?  Like we search on the first name and last name fields independently of each other?  Or, as an example, if a user searched for `dan Gu`, would we expect `Jordan Guinn` to be in the list of search results?  (The latter is more effort than the former, fwiw)
[...]

PR #1438, opened 2021-08-24
Add search to roster

lukenewman:
## Background
Added filtering functionality to the search bar on the roster redesign

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1334

## Associated PRs
https://github.com/ctoec/data-collection/pull/1460 ~~should be~~ has been merged in first

## Validation Plan
See associated issue/ticket. Filtering should match "first last" name formatting case-insensitively

## Automated Testing

## Additional Context
N/A


bamader (approved):
LGTM!

vonbearshark (approved):

JordanGuinn:
We're unfortunately gonna need search to take place server-side here, as pagination will be coming up as part of this roster redesign relatively shortly 😕

lukenewman:
@JordanGuinn ahhh tfw you didn't even read the AC... 😬

JordanGuinn:
HEH, I've definitely never done THAT before, countless times... 😅😅

client/src/containers/Overview/Overview.tsx
```diff
@@ -6,6 +6,7 @@ import { SiteSummaryTabNav } from './SiteSummaryTabNav';
 import pluralize from 'pluralize';
 import { useAlerts } from '../../hooks/useAlerts';
 import AuthenticationContext from '../../contexts/AuthenticationContext/AuthenticationContext';
+import { apiGet } from '../../utils/api';
```
vonbearshark, line 9: 🤔  how was this working without this..?

src/migration/1630005495640-UpdateChildAddFullName.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export class UpdateChildAddFullName1630005495640 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "child" ADD "fullName" nvarchar(255)`);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`ALTER TABLE "child" DROP COLUMN "fullName"`);
+  }
+}
```
vonbearshark, line 11: Do we want this to be an actual property on the `Child`? Or would we prefer to just derive it on demand? I am not sure how we feel about derivative properties like this, but I have personally avoided them

src/routes/children.ts
```diff
@@ -18,7 +18,8 @@ export const childrenRouter = express.Router();
 childrenRouter.get(
   '/',
   passAsyncError(async (req, res) => {
-    const children = await controller.getChildren(req.user);
+    const name = (req.query as any).name;
+    const children = await controller.getChildren(req.user, name);
```
vonbearshark, line 22: Let's create a `Query` type that we can share between the route and controller and that can include `searchTerm`, `sorting`, `pagination`, etc. And then let's just pass the `req.query` into the `getChildren` fn and pull those props off the object

src/controllers/children/get.ts
```diff
@@ -21,23 +21,32 @@ const whereNotDeleted = (field: string) => `${field}.deletedDate IS NULL`;
...
+      name: `%${name.toLowerCase()}%`,
+    });
+  }
+
+  // Restrict via org access
+  qb.andWhere('Child.organizationId IN (:...orgIds)', {
+    // SQLServer doesn't like queries like "IN ()", so supply
+    // an impossible value for empty arrays
+    orgIds: orgIds.length ? orgIds : ['NULL'],
+  })
```
vonbearshark, line 49: Might as well move this back to where it was? I guess it doesn't really matter, but ideally we filter on the `orgIds` first (even if it's just in the code flow and the generated SQL actions don't actually filter it first)

src/controllers/children/get.ts
```diff
@@ -21,23 +21,32 @@ const whereNotDeleted = (field: string) => `${field}.deletedDate IS NULL`;
...
-    .where('Child.organizationId IN (:...orgIds)', {
-      // SQLServer doesn't like queries like "IN ()", so supply
-      // an impossible value for empty arrays
-      orgIds: orgIds.length ? orgIds : ['NULL'],
-    })
+  var qb = await getManager().createQueryBuilder(Child, 'Child');
+
+  // optional filtering by name
+  if (name) {
+    qb.andWhere('Child.fullName LIKE (:name)', {
```
vonbearshark, line 39: Don't we want `ILike` so that the search is case-insensitive. I know we're converting `name` to lower, but won't `fullName` still be normally cased?

src/routes/children.ts
vonbearshark: We could probably share the type with the client, too, even

client/src/containers/Roster/Roster.tsx
```diff
@@ -101,12 +114,29 @@ const Roster: React.FC = () => {
...
 
+  const [childRows, setChildRows] = useState<ChildRow[]>([]);
+  const [searchTerm, setSearchTerm] = useState<string | undefined>();
+  useEffect(() => {
+    if (user && accessToken) {
+      (async function getRoster() {
+        var path = 'children';
+        if (searchTerm) {
+          path += `?name=${searchTerm}`;
+        }
```
vonbearshark, line 125: We should be able to use `stringify` and `query-builder` for this. If you search for those in the codebase there should be examples

client/src/containers/Roster/Roster.tsx
```diff
@@ -101,12 +114,29 @@ const Roster: React.FC = () => {
...
+        const result = await apiGet(path, accessToken);
+        setChildRows(result.map(childToRow) ?? []);
```
vonbearshark, line 127: FYI the `??` check shouldn't really do anything as is here. If result is undefined it's going try and map and throw an error (which is not caught with `??`), and if it is defined map will return an empty array at worst, which is truth-y. 
 You would want something like `result?.map(childToRow) ?? []`. But I actually don't think you need a check here, since `result` should either be defined or there should be a server error

vonbearshark: Actually, related, we should probably `try`/`catch` the request and handle a server error here

vonbearshark (changes requested):

vonbearshark:
Functionally looking great! Just comments re: cleaning up the Roster interface really

client/src/containers/Overview/Overview.tsx
lukenewman: this was a result of `base` being merged into `EPIC-roster-redesign` and then me merging that into my branch

src/migration/1630005495640-UpdateChildAddFullName.ts
lukenewman: this was taken directly from an acceptance criteria. are you opining that we can fetch firstName and lastName and then match temporary composite within query builder?

src/controllers/children/get.ts
lukenewman: for the life of me i could not get `ILIKE` to work

lukenewman: `fullName` is lowercased in the `updateFullName` fn

src/migration/1630005495640-UpdateChildAddFullName.ts
vonbearshark: > A computed column for name is added to the Child table, joining first and last name with a space. Should just be a simple DB migration

Ah yeah you're right. That is what I was suggesting, but ignore me!

client/src/containers/Overview/Overview.tsx
vonbearshark: Yeah I guess the diff is just confusing to me

src/migration/1630005495640-UpdateChildAddFullName.ts
lukenewman: i will test that out. we definitely need to if the migration query doesn't trigger the `beforeUpdate()` fullname updater

src/routes/children.ts
lukenewman: are you talking about something like this?

```
interface ChildrenQuery {
  searchTerm?: string;
  sorting?: string;
  ...
}
```

and then we map `req.query` onto that and then pass to the controller?

src/controllers/children/get.ts
vonbearshark: > `fullName` is lowercased in the `updateFullName` fn

Oh yeah interesting just noticed. I re: my previous comments I don't fully get with this part of the AC myself, but even still my personal pitch would be if we are going to store it to maintain the casing. Unless we do this for other naming stuff already in the db?

client/src/containers/Roster/Roster.tsx
lukenewman: can you expound on what the `try`/`catch` would include? handle server error as in show an error ui?

src/routes/children.ts
vonbearshark: Yeah exactly like that. We shouldn't need to map anything--we should be able to pass it right in and then deconstruct it in the controller fn itself

client/src/containers/Roster/Roster.tsx
vonbearshark: > as in show an error ui

That's what I am thinking. Here's an example from the `CreateOrg` stuff I am refactoring now https://github.com/ctoec/data-collection/blob/EPIC-roster-redesign/client/src/containers/CreateOrg/CreateOrg.tsx#L111. So, using our `alerts` system, with something like "Failed to load roster data", or something maybe

src/controllers/children/get.ts
vonbearshark: Not sure if anything in this thread is useful https://github.com/typeorm/typeorm/issues/4418. There are some interesting workarounds there. Including maybe using `LOWER` on `Child.fullName`, like `LOWER(Child.fullName)`?

client/src/containers/Roster/Roster.tsx
lukenewman: are you thinking something like this?

```
      (async function getRoster() {
        var path = 'children';
        if (searchTerm) {
          path += `?${stringify({searchTerm: searchTerm})}`;
        }
        const result = await apiGet(path, accessToken);
        setChildRows(result.map(childToRow));
      })().catch((error) => {
        setAlerts({ type: 'error', text: 'Failed to load roster' });
      });
```

src/controllers/children/get.ts
```diff
@@ -21,16 +25,18 @@ const whereNotDeleted = (field: string) => `${field}.deletedDate IS NULL`;
...
   organizationIds?: string[];
+  query?: ChildrenQuery;
 }) => {
   const readOrgIds = await getReadAccessibleOrgIds(user);
   const orgIds = organizationIds?.length
     ? intersection(readOrgIds, organizationIds)
     : readOrgIds;
 
-  const qb = await getManager()
+  var qb = await getManager()
```
vonbearshark, line 39: nit: keep the `const`

client/src/containers/Roster/Roster.tsx
```diff
@@ -41,26 +44,45 @@ import { useAlerts } from '../../hooks/useAlerts';
 import { ReactComponent as Change } from '../../images/change.svg';
 import { ReactComponent as Withdraw } from '../../images/withdraw.svg';
 import { ReactComponent as Trash } from '@ctoec/component-library/dist/assets/images/trashCan.svg';
-import { getStrippedFundingSourceName } from '../../utils/getFundingSpaceDisplayName';
-import { nameFormatter } from '../../utils/formatters';
+import { getErrorMessage } from '../CreateOrg/getErrorMessage';
```
vonbearshark, line 47: nit: unused?

client/src/containers/Roster/Roster.tsx
```diff
@@ -101,12 +115,31 @@ const Roster: React.FC = () => {
...
+      (async function getRoster() {
+        var path = 'children';
+        if (searchTerm) {
+          path += `?${stringify({ searchTerm: searchTerm })}`;
+        }
+        const result = await apiGet(path, accessToken);
+        setChildRows(result.map(childToRow));
+      })().catch((error) => {
+        setAlerts([{ type: 'error', text: 'Failed to load roster' }]);
+      });
```
vonbearshark, line 131: nit:

```
  useEffect(() => {
    if (accessToken) {
      (async function getRoster() {
        try {
          const result = await apiGet(
            `children?${stringify({ searchTerm })}`,
            accessToken
          );
          setChildRows(result.map(childToRow));
        } catch (error) {
          setAlerts((current) => [
            ...current,
            { type: 'error', text: 'Failed to load roster' },
          ]);
        }
      })();
    }
  }, [accessToken, searchTerm]);
```
?

updates `useEffect` deps, shorthands `searchTerm` object and stringification, try/catch for the api call itself, preserves current alerts in update

src/migration/1630005495640-UpdateChildAddFullName.ts
lukenewman: i have confirmed that the latest migration successfully backfilled the `fullName` column in the `devsecure` env

client/src/shared/models/db/Child.ts
```diff
@@ -21,6 +21,7 @@ export interface Child extends ObjectWithValidationErrors {
   firstName?: string;
   middleName?: string;
   lastName?: string;
+  fullName?: string;
```
bamader, line 24: Do we need this field? My initial reaction is "no, because anytime we want to use the `fullName` we can just concatenate `firstName + middleName + lastName`". I know it's _possible_ to use concatenations like that with SQL (and I'd imagine this field exists to make the SQL querying easier), so have we tried doing that before generating a new DB column that will require migrating data? If this is strongly easier then no worries, just wanted to check if we can accomplish the job using what we have first.

client/src/containers/Roster/Roster.tsx
```diff
@@ -101,12 +114,35 @@ const Roster: React.FC = () => {
...
+  const [searchTerm, setSearchTerm] = useState<string | undefined>();
+  useEffect(() => {
+    if (accessToken) {
+      (async function getRoster() {
+        try {
+          const result = await apiGet(
+            `children?${stringify({ searchTerm })}`,
+            accessToken
+          );
+          setChildRows(result.map(childToRow));
```
bamader, line 127: Nit: would prefer we stick to arrow function syntax here. I find it helps clarity and readability when looking at what this code is doing.

client/src/containers/Roster/Roster.tsx
```diff
@@ -101,12 +114,35 @@ const Roster: React.FC = () => {
...
   /// TODO: this is a placeholder! batch actions to be implemented with their own function
   const batchActionClick = (selectedRows: Readonly<DenormalizedRow[]>) => {};
 
   const childToRow = (child: Child) => {
     const { ageGroup, entry, fundings, site } = child.enrollments?.[0] ?? {};
-    const fundingSpace = fundings?.length ? fundings[0].fundingSpace : undefined;
+    const fundingSpace = fundings?.length
+      ? fundings[0].fundingSpace
+      : undefined;
     return {
```
bamader, line 146: Yeah going back to my earlier comment about merging in the `no-cache` code, it looks like there's some duplication; this `childToRow` function is in here twice. If we haven't merged those two together, we should to make sure we isolate duplicates and conflicts and clean them up now while we're still tweaking.

src/controllers/children/get.ts
```diff
@@ -12,6 +12,10 @@ import { AgeGroupCount } from '../../../client/src/shared/payloads/AgeGroupCount
 
 const whereNotDeleted = (field: string) => `${field}.deletedDate IS NULL`;
 
+export interface ChildrenQuery {
+  searchTerm?: string;
+}
+
```
bamader, line 18: Nit: do we need an interface to hold a single variable? Seems like it'd be more straightforward to me to just pass the search term directly into functions where it's needed.

bamader (commented):
Functionally looking okay, just some things to change either stylistically or with regard to the DB.

client/src/shared/models/db/Child.ts
vonbearshark: lol see https://github.com/ctoec/data-collection/pull/1438#discussion_r700328426

client/src/containers/Roster/Roster.tsx
vonbearshark: For `result.map(childToRow)`?

src/controllers/children/get.ts
vonbearshark: This was my suggestion to make the Roster interface more generic for sort and pagination following this PR https://github.com/ctoec/data-collection/pull/1438#discussion_r700330737

src/entity/Child.ts
```diff
@@ -57,6 +57,14 @@ export class Child implements ChildInterface {
   @IsNotEmpty()
   lastName?: string;
 
+  @Column({ select: false })
+  fullName?: string;
```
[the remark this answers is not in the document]
vonbearshark: Same re: https://github.com/ctoec/data-collection/pull/1438#discussion_r701088144. All 👀 on @JordanGuinn to say for sure. But your similar suspicious does reinforce my hunch imo..

client/src/shared/models/db/Child.ts
bamader: LOL okay, I'll take it 😂

client/src/containers/Roster/Roster.tsx
```diff
@@ -41,26 +44,44 @@ import { useAlerts } from '../../hooks/useAlerts';
...
+  missingInfo: boolean;
+  name: string;
+  site?: Site['siteName'];
 }
 
+const childToRow = (child: Child) => {
+  const { ageGroup, entry, fundings, site } = child.enrollments?.[0] ?? {};
+  const fundingSpace = fundings?.length ? fundings[0].fundingSpace : undefined;
+  return {
+    ageGroup,
```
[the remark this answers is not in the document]
bamader: Something to check for sure, especially given the duplicate function code

client/src/containers/Roster/Roster.tsx
bamader: Yeah. I just find an extra clarity boost in `result.map((c) => childToRow(c))`. Not a big deal if others don't mind it.

src/controllers/children/get.ts
bamader: That's fair. No real objection to it, and if you think it'll set up future work I'm all for it.

src/entity/Child.ts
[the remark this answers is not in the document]
bamader: HAHA I also missed that it was straight up part of the AC. Wow I need to up my review game...

client/src/containers/Roster/Roster.tsx
vonbearshark: It's my preference, but readability > preference, so fine to change. jw since I would do it in the future otherwise, so good to know

client/src/containers/Roster/Roster.tsx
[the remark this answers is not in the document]
lukenewman: yeah this pr has the latest from that `no-cache` branch and also from a recent `base`. this dupe slipped past me!

src/entity/Child.ts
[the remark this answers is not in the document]
JordanGuinn: Check out [this Slack thread](https://skylight-hq.slack.com/archives/CM7D99AF3/p1629991490000900) for more details - TLDR; there was initial investigation into setting up as a virtual field, only to uncover that TypeORM doesn't actually support that (even though SQL Server itself does).  So it was decided that an actual field was the next best (or least worst) option

client/src/shared/models/db/Child.ts
JordanGuinn: Biggest concern around that strategy is performance - we'd have to generate that concatenation on each and every search attempt, and I'm not clear on what the implications of that would be

JordanGuinn:
In a nutshell, let's stick with `fullName` as an actual field because TypeORM doesn't really provide any actual support for virtual fields and on-the-fly field generation as a component of each search query doesn't seem like a particularly performant alternative

client/src/containers/Roster/Roster.tsx
lukenewman: it'll have to look like `result.map((c: any) => childToRow(c))

client/src/shared/models/db/Child.ts
bamader: Yeah totally fair

src/entity/Child.ts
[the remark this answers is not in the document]
bamader: This makes sense. No real objections to the actual field, especially if substantially improves our quality of life.

client/src/containers/Roster/Roster.tsx
```diff
@@ -234,6 +248,10 @@ const Roster: React.FC = () => {
                     tabIndex={
                       getBatchActionProps().shouldShowBatchActions ? -1 : 0
                     }
+                    onChange={(input) => {
```
JordanGuinn, line 251: We want this to trigger on click, yes?  Just to avoid spamming the DB a bunch as the user types 😅

src/controllers/children/get.ts
JordanGuinn: Agreed - we'll be adding to this interface _very_ soon

JordanGuinn (approved):
LGTM me overall - just one minor q regarding search init on keypress vs on click/enter

client/src/containers/Roster/Roster.tsx
```diff
@@ -234,6 +250,15 @@ const Roster: React.FC = () => {
...
+                      setSearchTerm(input.target.value);
+                    }}
+                    onKeyDown={(event) => {
+                      if (event.key === 'Enter') {
```
JordanGuinn, line 257: hehehehe 😎

nickbristow:
## Validation Plan
- [ ] Search bar is on roster
- [ ] Search button can't be clicked if no text is entered
- [ ] hitting enter searches
- [ ] search is case insensitive
- [ ] search is fuzzy
- [ ] search matches against first and last name

nickbristow:
Search bar does not have a button, and requires you to hit enter key. Are we ok with this @JordanGuinn?

Also, only accepts first name last name and no commas, which may be confusing considering the table is listed as last name first name.

nickbristow:
@lukenewman Can we add a button and match the look of the mockups? Unless there was some decision made im not aware of?

lukenewman:
> @lukenewman Can we add a button and match the look of the mockups? Unless there was some decision made im not aware of?

@nickbristow we made the decision to roll with carbon's table components in the multi-select work since it has all the stuff we're looking for out-of-the-box. that being said, i'm not sure if there's a "search" button so will need to investigate

JordanGuinn:
Let's just use a native Carbon button, [as we're doing here](https://github.com/ctoec/component-library/blob/base/src/components/SearchBar/SearchBar.tsx).

lukenewman:
@nickbristow this is good to go again. users should be able to search on enter and search button click. clearing the search bar manually or via the X button should automatically refetch the roster

nickbristow:
Design doesn't exactly match, but functionality is good to go. 

Hogwarts_Childcare

JordanGuinn:
Confirmed that designs are good as-is for the time being - so if all else is well functionally, then this is ready for merge!

PR #1462, opened 2021-09-01
Speed up builds by minimizing Carbon Sass imports

JordanGuinn:
## Background
Turns out the most significant culprit for our client build performance woes was the fact that we've been loading and compiling _all of Carbon's Sass_ on start-up, regardless of whether or not any of it pertains to components we actually use.  This PR addresses that, simply [pulling in what we need and nothing more](https://github.com/carbon-design-system/carbon/blob/main/docs/guides/sass.md#optimizing-your-sass-builds)!

## GitHub Issue
N/A

## Associated PRs
- https://github.com/ctoec/component-library/pull/166

## Validation Plan
- [ ] Confirm application can be built and run locally without issue
- [ ] Confirm application can be built and deployed without issue
- [ ] Confirm general application styles look as they did previously
- [ ] Confirm all components look as they did previously

## Automated Testing


client/src/containers/EditRecord/EditRecord.scss
```diff
@@ -1,5 +1,3 @@
-@import '~@ctoec/component-library/dist/assets/styles/index.scss';
```
vonbearshark, line 1: lol

client/src/index.scss
```diff
@@ -1,3 +1,8 @@
+@import '~@ctoec/component-library/dist/assets/styles/index.scss';
```
vonbearshark, line 1: Isn't this also importing all the styles?

vonbearshark (commented):
v hyped

client/src/index.scss
JordanGuinn: All of _our_ styles, yes - there are a number of thematic and color-specific styles pertinent to USWDS that we're pulling in from that `index.scss`, but we're NOT pulling in the entirety of `carbon-components` anymore, which was **by far** the biggest problem

vonbearshark: just noticed the difference 👍🏻

vonbearshark (approved):

client/src/carbon-component-styles.scss
```diff
@@ -0,0 +1,26 @@
...
+// // Optionally include the grid
+// @import 'carbon-components/scss/globals/grid/_grid.scss';
+
+//  ONLY INCLUDE STYLES FOR THE COMPONENTS THAT ARE ACTIVELY IN USE
+//  Sass compilation is a relatively slow and expensive task, performance-wise
+@import 'carbon-components/scss/components/data-table/data-table';
+@import 'carbon-components/scss/components/multi-select/multi-select';
+@import 'carbon-components/scss/components/number-input/number-input';
+@import 'carbon-components/scss/components/text-input/text-input';
+@import 'carbon-components/scss/components/select/select';
```
bamader, line 26: So if/when we want to use other carbon components, will we have to import their `scss` styles here too?

client/src/components/AddRecordButton.tsx
```diff
@@ -33,39 +27,42 @@ export const AddRecordButton: React.FC<AddRecordButtonProps> = ({
       // the same alignment the text is
       <div className="display-inline-block">
         <Button
-          appearance={appearance || 'unstyled'}
+          kind="ghost"
```
bamader, line 30: Wut this do?

bamader: Nvm looked it up cuz it's everywhere. Seems solid.

client/src/containers/EditRecord/EnrollmentFunding/EditEnrollmentCard.tsx
```diff
@@ -93,7 +93,7 @@ export const EditEnrollmentCard: React.FC<EditEnrollmentCardProps> = ({
             setAlerts={setAlerts}
             AdditionalButton={
               <ExpandCard>
-                <Button text="Cancel" appearance="outline" />
+                <Button kind="tertiary">Cancel</Button>
```
bamader, line 96: General question: why do we use `tertiary` for some of the `Cancel` or `Skip` buttons, but `ghost` for others?Seems like maybe we should make that styling consistent?

bamader (approved):
For how comprehensive this was, this refactor must've _sucked_ to implement. Everything looks great though!

client/src/carbon-component-styles.scss
JordanGuinn: Yep!

client/src/containers/EditRecord/EnrollmentFunding/EditEnrollmentCard.tsx
JordanGuinn: It was entirely to maintain parity with the current state of the application.  Based on the [Carbon React style docs](https://react.carbondesignsystem.com/?path=/story/components-button), a [ghost button](https://react.carbondesignsystem.com/?path=/story/components-button--ghost) effectively looks like a link, while a [tertiary button](https://react.carbondesignsystem.com/?path=/story/components-button--tertiary) is transparent save for it's border.

So wherever you see `tertiary` or `ghost` used, it's because it specifically matches the current application UI.

JordanGuinn: There may very well be inconsistencies in how we style various cancellable buttons generally in the application, but that is certainly beyond the scope of this already ballooned PR.

client/src/components/AddRecordButton.tsx
JordanGuinn: Yep!  Just another native Carbon thing 🙂

client/src/carbon-component-styles.scss
JordanGuinn: Er, rather if you want to style a given component meaningfully with a lot of the out-of-the-box CSS that Carbon provides.  It will become **very** obvious if it's not imported, so forgetting to do so won't be an issue

client/src/_variables.scss
```diff
@@ -0,0 +1 @@
+$button-color-on-hover: #4c3d69;
```
vonbearshark, line 1: Is this a convention, or should we just add this var into the `globals`?

client/src/_globals.scss
```diff
@@ -0,0 +1,173 @@
+@import './variables';
```
vonbearshark, line 1: Should we add any of this into the component lib instead?

client/src/_globals.scss
```diff
@@ -0,0 +1,173 @@
...
+
+/*********************************************************************
+ *** Button
+ *********************************************************************/
+.usa-button {
```
vonbearshark, line 6: I see a lot of removals of `usa-button`--is it still used?

vonbearshark (approved):
What an effort 🦾 

lgtm? But lmk if there's any specific pieces you want to more closely scrutinize

bamader (approved):
I think it all looks good, but imma be real, reviewing CSS changes is the worst so I'll probably need validation to help verify things.

client/src/_variables.scss
JordanGuinn: I've seen it as a convention a bunch, though given we only have one variable at this point, probably seems most reasonable to throw it into `globals`

JordanGuinn: Though it _would_ mean that we now have to import `globals` anywhere else that we'd like to reference variables like this, which I'm not really a fan of given how much else is in there

client/src/_globals.scss
JordanGuinn: We're extending it with the carbon classes, though I'm not sure if it's necessary at this point.  I'll play around with it and see

client/src/_variables.scss
vonbearshark: Yeah, fair. Maybe we want to go in the opposite direction, even, pulling more declarations in `globals` into the `variables`

but fine as is for this PR

Issue #1357, opened 2021-05-27
As an OEC Provider, I can see an automated e-mail notifying me to update my roster


Derekskylight:
## Background
As an OEC Provider, I should receive an automated e-mail notifying me to submit my monthly data. This will allow all providers within ECE Reporter to have their data accurate on a monthly cadence in one place rather than disparate sources. With updated data, users will be able to understand their data better, and this could also lay the groundwork for admins to see it or monthly reports to be created in the future.

## Acceptance Criteria
- Submission reminder email for **each month** is sent out on **the first of the following month**
   - This can be accomplished by setting up a [cron job](https://www.digitalocean.com/community/tutorials/nodejs-cron-jobs-by-examples) that utilizes SendGrid's API, an email provider we already have provisioned
- Email is sent to **all providers in ECE Reporter**
- Email contents match specs captured below

## Technical Details
While using something like [Serverless](https://www.serverless.com/)/[Lambda](https://aws.amazon.com/lambda/) is probably the most appropriate technical solution for something like this, it would require additional infrastructural investment (Lambda spin-up, API Gateway spin-up, Serverless configuration and permissioning, etc) that simply isn't worth the effort at this moment, _especially_ given the uncertainty around whether or not we'll even continue with AWS in the foreseeable future.  So, as an interim stopgap, let's just [set up a cron job on our Express server](https://www.digitalocean.com/community/tutorials/nodejs-cron-jobs-by-examples) to get this done.

## Designs
![Email.PNG](https://images.zenhubusercontent.com/602d78337403e44d7db5c4f3/e72b8992-3c5f-4cae-848c-83b5915627a7)

JordanGuinn:
@jlhogan @Derekskylight @ehuang525 Wondering if this is actually worth an engineering ticket at this point?  It looks like we can automate recurring batch emails entirely through the SendGrid UI, which would assuredly be quicker than whatever tech solution we'd roll out.

JordanGuinn:
Update - spoke in person, and decided that because we would still need to continually update the list of users that emails get sent to, we might as well build a technical solution here.

JordanGuinn:
@JordanGuinn FYI https://www.digitalocean.com/community/tutorials/nodejs-cron-jobs-by-examples

bamader:
Structure question: where's the best place to put this code? I feel like we should probably give  it a new file somewhere within `src/` but I'm not entirely sure where.

PR #1465, opened 2021-09-06
Automated Monthly Emails

bamader:
## Background
This PR sets up automated monthly reminder emails to go out to all providers on the first of each month, reminding them to update their roster.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1357

## Associated PRs
- https://github.com/skylight-hq/ctoec-devops/pull/220

## Validation Plan
- Start up the app
- Change the target email address, name, and time signature to match the validator and the day (e.g. to test whether you can receive an email scheduled for 2:35 PM on September 10th, you would use `35 14 10 9 *` as the cron signature in the first param of `schedule`).
- Wait for the time to pass
- Check inbox for email

## Automated Testing

## Additional Context
Before merging this PR in, we need to change two things:
1.  The actual provider emails need to be put into the  `TO` field for the outgoing emails, and
2. The time signature of the cron job scheduler needs to be set to the start of the day on the first of each month (`0 0 1 * *`).
These aren't set right now because obviously we don't want to bombard providers with dummy emails while testing.


JordanGuinn:
Changing the base branch here to `RELEASE-monthly-submission-alerts`, cuz we're not actually quite ready to release this yet

src/index.ts
```diff
@@ -61,6 +65,76 @@ getConnectionOptions().then((connectionOptions) => {
       }
 
       const port = process.env.PORT;
+
+      cron.schedule('47 11 9 9 *', async function() {
+
+        const allUsers = await getManager().find(User);
```
JordanGuinn, line 71: Maybe not _all_ users?  Probably don't want to send this to admin users, right?

src/index.ts
```diff
@@ -10,6 +10,10 @@ import { QueryLogger } from './loggers/QueryLogger';
 import { isProdLike } from './utils/isProdLike';
 import { requestLogger } from './loggers/RequestLogger';
 import { dateReviver } from '../client/src/shared/dateReviver';
+import moment from 'moment';
+import { User } from './entity';
+const cron = require('node-cron');
```
JordanGuinn, line 15: nit: can these not be brought in using the `import` syntax?  Like:

`import * as cron from 'node-cron';`

JordanGuinn (approved):
Love it!  Just a few minor things

vonbearshark (approved):
minus the user values stuff lgtm!

src/index.ts
bamader: It doesn't like that. I tried it for both of them and got hit with a "cannot find declaration file".

src/index.ts
bamader: Yeah very fair

src/index.ts
JordanGuinn: ah, it's looking for TS models I imagine, smh

src/index.ts
```diff
@@ -61,6 +65,78 @@ getConnectionOptions().then((connectionOptions) => {
...
+          const userEmail = u.email;
+
+          client.setApiKey(process.env.SENDGRID_API_KEY);
+          const request = {
+            method: 'POST',
+            url: '/v3/mail/send',
+            body: {},
+          };
+
+          // PLUG IN ACTUAL USER VALUES BEFORE MERGING
```
lukenewman, line 88: before merging into the temp release branch?

src/index.ts
```diff
@@ -61,6 +65,78 @@ getConnectionOptions().then((connectionOptions) => {
       }
 
       const port = process.env.PORT;
+
+      cron.schedule('47 11 9 9 *', async function () {
```
lukenewman, line 69: will this schedule meet the AC of sending on the first of every month? from the pr description it looks like this would only run on Sep 9th at 11:47

lukenewman, line 69: will this schedule meet the AC of sending on the first of every month? from the pr description it looks like this would only run on Sep 9th at 11:47

lukenewman (commented):
seems like there might be a few things that need to be updated before merge, but otherwise lgtm

src/index.ts
JordanGuinn: Yep - this was put in temporarily to actually validate behavior, and will need to be updated prior to merge.

JordanGuinn:
## Validation Plan
- [ ] Confirm email is sent to all non-admin users in ECE Reporter DB
- [ ] Confirm emails are sent at the time specified in the cron job schedule
- [ ] Confirm email contents match what's provided in the designs

JordanGuinn:
After a few tweaks, confirmed cron behavior works and appropriate emails are sent as expected.  This is ready for merge!

Issue #1280, opened 2021-04-16
Draft: As an OEC Admin User, I can edit an organization's information


Derekskylight:
## Background
As an OEC Admin User, I can edit an organization's details. This includes basic information, as well as funding spaces and necessary individual users and their permissions

## Acceptance Criteria
**As an OEC Admin User, I am able to edit an organization's:**

**Details:**
- An organization's name
- A site's name
- Ability to add more sites within an organization

**Funding spaces:**
- Number of Infant/Toddlers
- Number of Preschoolers
- Number of School-Aged
- Add a new type of funding space (CDC/SR/SS)
    - TBD

**Individual Users and Permissions** 
TBD

## Designs
[Create Org Figma File](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=2%3A2)

JordanGuinn:
@Derekskylight @ehuang525 Are there any final designs for this?  Or are we gonna be repurposing the Create Organization designs here?

ehuang525:
@JordanGuinn this will need to be updated! we will be repurposing most of the Create Organization page here

JordanGuinn:
Sounds like a plan - thanks Emily!

ehuang525:
@JordanGuinn The mockup here has been updated with some revisions coming out of the OEC interviews. Let's discuss and reestimate/reorganize how we want to approach this work before pulling this into the next sprint!

PR #1477, opened 2021-09-09
Vonbearshark/1280

vonbearshark:
## Background
- Created an Edit org flow, including UI and backend
- Abstracted out the existing Create code to be leveraged by both Create and Edit
- Made Title I and Region optional fields on sites
- Reorganized the /organizations route

## GitHub Issue
#1280 

## Validation Plan
TBD but will need to test Edit, Create, and any interactions with the `/organizations` backend route

## Automated Testing

## Additional Context
I wouldn't say this is WIP, but I am still validating and discovering new edge cases locally

I tried to group related changes to be reviewed commit-by-commit, if that's helpful

I was wondering about the relationship between org and site permissions: are the mutually exclusive, or can some users be both? How does that work right now?


src/utils/getReadAccessibleOrgIds.ts
```diff
@@ -16,7 +16,7 @@ export const getReadAccessibleOrgIds = async (
   ).map((site) => site.organizationId);
 
   // combine and deduplicate site org Ids + org Ids
-  return Array.from(new Set([...siteOrgIds, ...user.organizationIds])).map(
-    (id) => `${id}`
-  );
+  return Array.from(
+    new Set([...siteOrgIds, ...(user?.organizationIds ?? [])])
```
vonbearshark, line 20: just adding `?` `??` checks in an attempt to stop an unrelated local dev bug I came across

src/controllers/users.ts
```diff
@@ -27,7 +27,7 @@ export const addSiteAndOrgDataToUser = async (user: User) => {
   // Need to append organizations for which the user can just
   // access sites, too, so that they can be edited in the front-end
   const orgIds = uniq(
-    (user.organizationIds || []).concat(sites.map((s) => s.organizationId))
+    (user?.organizationIds ?? []).concat(sites.map((s) => s.organizationId))
```
vonbearshark, line 30: just adding ? ?? checks in an attempt to stop an unrelated local dev bug I came across

src/entity/Site.ts
```diff
@@ -44,6 +45,9 @@ export class Site implements SiteInterface {
   @Column()
   organizationId: number;
 
+  @OneToMany((type) => SitePermission, (perm) => perm.site)
+  sitePermissions?: Array<SitePermission>;
+
```
vonbearshark, line 50: Added relationship here to facilitate the query to find the list of users to show in the Edit page

src/entity/Organization.ts
```diff
@@ -32,6 +33,9 @@ export class Organization implements OrganizationInterface {
   @Column({ unique: true })
   providerName: string;
 
+  @OneToMany((type) => OrganizationPermission, (perm) => perm.organization)
+  orgPermissions?: Array<OrganizationPermission>;
+
```
vonbearshark, line 38: Added relationship here to facilitate the query to find the list of users to show in the Edit page

client/src/containers/Organization/NewSiteFormCard.tsx
```diff
@@ -29,12 +30,17 @@ export const NewSiteFormCard: React.FC<NewSiteFormCardProps> = ({
...
+  fundingSpaces,
 }) => {
   const getMissingInfo = () => {
     const errors: Partial<Record<keyof Site, boolean>> = {
+      ...(fundingSpaces.some((fs) => fs?.source === FundingSource['CDC'])
+        ? {
+            titleI: false,
+            region: false,
+          }
+        : {}),
```
vonbearshark, line 42: Confirming this behavior: we want to run a client check to require these fields only if there are some CDC funding spaces?

client/src/containers/Organization/UpdateOrg.tsx
```diff
@@ -39,7 +51,49 @@ const CreateOrg: React.FC = () => {
   const [showErrors, setShowErrors] = useState(false);
   const history = useHistory();
 
-  const createNewOrg = async () => {
+  // Get deduped users based on permissions: either for the org itself or its sites
```
vonbearshark, line 54: re: my q from the description: do we actually want to dedupe here? What does the overlap look like between the org and site permissioned users?

src/controllers/organizations.ts
```diff
@@ -17,51 +74,82 @@ async function doesOrgNameExist(name: string) {
...
+  if (id) {
+    organization.providerName = name;
+    await getManager().save(organization);
+  }
+
+  const upsertedSites: Site[] = await Promise.all(
+    sites.map(async (s: Partial<Site>) => {
+      // Either find existing by ID or create
+      if (s.id) {
+        // Thanks, typeorm!
```
vonbearshark, line 124: open to alternatives here?

client/src/containers/Organization/CreateOrg.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import UpdateOrg from './UpdateOrg';
+
+const CreateOrg: React.FC = () => (
+  <UpdateOrg
```
JordanGuinn, line 6: Already liking this a LOT 😎

client/src/containers/Organization/NewSiteFormCard.tsx
JordanGuinn: Correct!  `TitleI` and `Region` are now required only for orgs that have CDC funding spaces, whereas `Site Name` and `Registry ID` are now required across the board.

client/src/containers/Organization/UpdateOrg.tsx
JordanGuinn: I'm not sure I quite follow the question, but I will say that users with entries in the `OrganizationPermission` table should not also have entries in the `SitePermission` table for sites that correspond to the same org, and vice versa.

src/controllers/organizations.ts
```diff
@@ -1,14 +1,70 @@
...
+    orgIds,
+    name,
+  }: { orgIds?: Organization['id'][]; name?: Organization['providerName'] },
+  user
+): Promise<Organization[]> => {
+  const ids = orgIds ? [orgIds] : await getReadAccessibleOrgIds(user);
+
+  const qb = await getManager()
+    .createQueryBuilder(Organization, 'Organization')
+    .where('Organization.id IN (:...ids)', {
```
JordanGuinn, line 25: nit: can we instead just throw this `where` clause in an `if` block?  Checking against a `NULL` array seems kinda wonky

src/controllers/organizations.ts
```diff
@@ -1,14 +1,70 @@
...
+  if (name) {
+    qb.andWhere('LOWER(Organization.providerName) Like :providerName', {
+      providerName: `%${name
+        .toLocaleLowerCase()
+        .replace(/[.,\/#!$%\^&\*;:{}=\-_`~()\s\\]/g, '%')}%`,
+    });
+  }
+
+  return await qb
+    .leftJoinAndSelect('Organization.fundingSpaces', 'fundingSpaces')
```
JordanGuinn, line 40: 🔥🔥🔥

JordanGuinn (approved):
Big fan of this refactor overall 👏🏽👏🏽

client/src/containers/Organization/UpdateOrg.tsx
vonbearshark: should like it definitely doesn't/can't, or should like it shouldn't? My local env might be a little goofed by now, but I thought I saw users in both

src/controllers/organizations.ts
vonbearshark: Yeah agreed. Actually, this is pulled from the roster children get https://github.com/ctoec/data-collection/blob/base/src/controllers/children/get.ts#L41.
1. If it's wonk there, too, should I change it?
2. I'm not 💯 if typeorm will let you gate the first `where` (versus the subsequent `andWhere`s), but I can give it a shot
3. I think I did see it throw an error when it tried to query against that `NULL` array in the roster children now that you mention it. (It just hasn't hit the condition in my query in my tests.) Would be interested learning why we did that originally

client/src/containers/Organization/UpdateOrg.tsx
JordanGuinn: I imagine whether or not it can happen depends on what we do here, but there are no instances of this in production at all right now.  There are two users that currently have both org-level and site-level permissions, but the sites they have site-level permissions **do not** belong to any of the orgs they also have org-level permissions for.

vonbearshark: > I imagine whether or not it can happen depends on what we do here

Yeah fair point, and I do see this check here that should prevent the overlap https://github.com/ctoec/data-collection/pull/1477/files#diff-4bf785c65eece578685169bd64a75b00ddc3a7c818308c1f13f906f87e4d77c9R150

src/controllers/organizations.ts
JordanGuinn: Are you not able to just call `save` with the partial Site `s` in both cases?  [According to the TypeORM docs](https://typeorm.io/#/repository-api), `save` should act as an upsert, so I'd think that if we gave it a site _with_ an ID, it would update the existing site accordingly.  And if we gave it one _without_, it would create a new site.

vonbearshark: > Are you not able to just call save with the partial Site `s` in both cases

I would _love_ that. My original code was something like `getManager().save(Site, { ...s })`. I think the issue is that you need an entity instance from typeorm passed into `save`. So you need to retrieve it yourself. And then since it's an object instance you have to also use the individual prop setting to update the values.

vonbearshark: > Also supports partial updating since all undefined properties are skipped

from the docs is either not accurate or I haven't been able to swing it. From looking at the codebase it seems like we've run into this issue before?

vonbearshark: Although looking at the doc maybe `preload` can help load the entity with the partial in both create and update cases?

lukenewman (approved):

src/controllers/organizations.ts
vonbearshark: Yeah, based on the docs we should be able to do

```
  const organization = await getManager().preload(Organization, {
    id,
    providerName: name,
  });
 ```
 
 or at least
 
 ```
 const organization = await getManager().preload(Organization, {
    ...(id && { id }),
    providerName: name,
  });
 
  ```
  
  Since
  
>  **Creates** a new entity from the given plain javascript object. **If the entity already exists in the database**, then it loads it (and everything related to it), replaces all values with the new ones from the given object, and returns the new entity.
  
 This does work on edit (with an ID), but fails on create (no ID provided)
 > Error creating organization:  Given object does not have a primary column, cannot transform it to database entity

Comment (vonbearshark) [reply src/controllers/organizations.ts:123]: In fairness, the docs _do_ say

> Note that given entity-like object must have an entity id / primary key to find entity by. Returns undefined if entity with given id was not found.

but unclear how that would work on create flows? And isn't that the whole point of the command?

Comment (vonbearshark) [reply src/controllers/organizations.ts:123]: At the very least it _does_ work with edit flows, which was the main issue with this syntax. So will update with this caveat

Comment (vonbearshark) [reply src/controllers/organizations.ts]: Still annoyingly need to branch on the presence of ID in latest push, but at least it's a little more flexible/less verbose in the prop setting

**Review by vonbearshark — COMMENTED** [review]
Was able to address the existing feedback and fix local issues. This should be ready for validation

**Comment by nickbristow** [comment]
## Validation
- [ ] Organization Overview page functions as expected

**Comment by vonbearshark** [comment]
Validation plan for Create and Edit flow (extended from older PRs)
 
 - [ ] Organizations are created successfully when name is available
    - [ ] Success messaging indicates as much
- [ ] Organizations are NOT created successfully when name is unavailable
    - [ ] Error messaging indicates as much/
 - [ ] Organizations are edited successfully when name is available
    - [ ] Success messaging indicates as much
- [ ] Create an organization with one site without issue
   - [ ] Confirm success banner is displayed
   - [ ] Confirm user is navigated back to the main admin view
- [ ] Edit an organization with one site without issue
   - [ ] Confirm success banner is displayed
- [ ] Create an organization with multiple sites without issue
   - [ ] Confirm success banner is displayed
   - [ ] Confirm user is navigated back to the main admin view
- [ ] Edit an organization with multiple sites without issue
   - [ ] Confirm success banner is displayed
- [ ] Confirm an org with one site **isn't** created under the following conditions:
   - [ ] Site name is missing
   - [ ] Registry ID is missing
   - [ ] There is a CDC funding and Title I is missing
   - [ ] There is a CDC funding and Region is missing
- [ ] Confirm an org with one site **isn't** edited under the following conditions:
   - [ ] Site name is missing
   - [ ] Registry ID is missing
   - [ ] There is a CDC funding and Title I is missing
   - [ ] There is a CDC funding and Region is missing
- [ ] Confirm an org with multiple site **isn't** created under the following conditions:
   - [ ] Site name is missing for one site
   - [ ] Registry ID is missing for one site
   - [ ] There is a CDC funding and Title I is missing for one site
   - [ ] There is a CDC funding and Region is missing for one site
   - [ ] Two sites share the same name
- [ ] Confirm an org with multiple site **isn't** edited under the following conditions:
   - [ ] Site name is missing for one site
   - [ ] Registry ID is missing for one site
   - [ ] There is a CDC funding and Title I is missing for one site
   - [ ] There is a CDC funding and Region is missing for one site
   - [ ] Two sites share the same name
- [ ] Confirm a user can remove sites they inadvertently added during org creation
- [ ] Confirm a user can remove sites they inadvertently added during org editing
- [ ] Confirm adding a site, then removing it, then saving the org does not include that removed site
- [ ] Confirm users can be assigned to orgs
- [ ] Confirm users can be assigned to sites
- [ ] Confirm users can be unassigned to orgs
- [ ] Confirm users can be unassigned to sites

--- 

Completed, and should not need to be re-validated based on iteration:

Validation plan for Overview/Summary changes
- [x] Confirm counts of:
    - [x] Organizations
    - [x] Sites
    - [x] Enrollments
    - [x] Users

File: src/controllers/organizations.ts (hunk 7 above)
Comment (JordanGuinn) [reply src/controllers/organizations.ts]: Wow, the idea that you have to provide a primary key to look up by, which obvs you don't yet have if you're trying to create an entity 🤦🏽‍♂️ Doesn't seem like TypeORM thought through that logical sequence too well

Comment (JordanGuinn) [reply src/controllers/organizations.ts]: Agreed though - what's here is a bit more flexible, if nothing else

**Review by vonbearshark — COMMENTED** [review]
Form stuff. Always a joy. Latest commit should fix the issues we saw @nickbristow . Just updated from `base` and should be good for re-validation for anyone who has a sec. Created a validation plan but happy to talk through as well

**Comment by nickbristow** [comment]
- Organizations are created successfully when name is available
   - Success messaging indicates as much `Message is displayed and then page refreshes immediately`
- On edit
  - Confirm properties like Registry ID is both required and set `Registry ID is not set on edit form, Age group on funding space not set on edit form`

There is a CDC funding and Title I is missing `always required`
 There is a CDC funding and Region is missing `always required`

`Deleting the registry ID allows you to create the org`

`Able to edit an org and give sites the same name`

`Can't remove site from edit view`

`Getting an error when trying to access users orgs`

File: src/routes/organizations.ts [hunk 11]
```diff
@@ -6,46 +6,53 @@ import {
...
-import { getManager } from 'typeorm';
-import { Organization } from '../entity';
-import { parseQueryString } from '../utils/parseQueryString';
 
 export const organizationsRouter = express.Router();
 
+// TODO: Should integrate into top-level routing
+const adminCheck = (req) => {
+  if (!req.user?.isAdmin) throw new ForbiddenError();
+};
```
bamader, line 15: I definitely wrote a PR that did this at the routing level--that's why all of our other `isAdmin` checks disappeared. Can we not re-use that here?

bamader (approved):
A lot of stuff happening in this PR, but from everything I can see it all looks good. I'm so-so on the refactoring parts, but also my brain is still at like 80% so once I get back into working I bet things clear up. TLDR, looks good!

src/routes/organizations.ts
vonbearshark: For the server or for the client?

bamader: Ah it was client side. It's in `client/src/containers/App/MakeRouteWithSubroute.tsx` but you're right that doesn't cover the server-side error handling we want if access is forbidden.

vonbearshark:
Fixed
- `Registry ID not set on update form`
- `Title I and Region required on CDC funding`
- `Removing sites/funding spaces on update form`
- `Unsetting optional site fields`

On prod/couldn't repro (probably fixed with other changes)/unrelated:
- `Success messaging indicates as much Message is displayed and then page refreshes immediately`
- `Getting an error when trying to access users orgs (User > user name)`
-` Deleting the registry ID allows you to create the org`
- `Empty sites in Site Permissions field for user (unnamed 1x in UI)`
- `Able to edit an org and give sites the same name`

bamader (approved):

vonbearshark:
Given the CI issues, closing and creating a new PR here https://github.com/ctoec/data-collection/pull/1499

PR #1478, opened 2021-09-10
Specify env vars on application deploy

JordanGuinn:
## Background
In order to customize our environments with the necessary secrets/configuration values without storing them in Terraform state, we're now specifying all of our back-end environment variables on application deploy, using [env-cmd](https://github.com/toddbluhm/env-cmd).

Additionally, customizing environment variables for local development can now be done safely, without the potential concern of committing them to git.  In order to do so, simply perform the following steps:

#### Developing with Docker
1. Create a `.env.docker` file at the root of the project.
2. Specify all the env variables you'd like to override in this new file.  Syntax is as follows:
 ```
ENV_VAR_1=ENV_VAR_1_VALUE
ENV_VAR_2=ENV_VAR_2_VALUE
```

#### Developing without Docker
1. Create a `.env.local` file at the root of the project.
2. Specify all the env variables you'd like to override in this new file.  Syntax is as follows:
 ```
ENV_VAR_1=ENV_VAR_1_VALUE
ENV_VAR_2=ENV_VAR_2_VALUE
```

## GitHub Issue
N/A

## Associated PRs
- https://github.com/skylight-hq/ctoec-devops/pull/221

## Validation Plan
- [x] Confirm application is functional locally
- [x] Confirm application is functional through Docker
- [x] Confirm application is functional when deployed to an AWS environment
- [x] Confirm conditional environment variables are set appropriately on AWS deploys

## Automated Testing


nickbristow (approved):

vonbearshark (approved):

.env
```diff
@@ -1,9 +0,0 @@
-CLIENT_URL='http://localhost:3000'
```
vonbearshark, line 1: Do we want to keep this file for reference and just rm the values?

JordanGuinn: My hesitation there is that it still leaves the potential for inadvertently adding secrets to git.  Would be easy enough to c/p these values into your local `.env.local` or `.env.docker` though, for personal ref

vonbearshark: Can we rm the values then add it to `.gitignore`? It's pretty handy to do/add to the README: `cp .env .env.local`, etc and have the variables and syntax populated. But agreed re: the risk and makes sense as is as well

JordanGuinn: Adding the file to the `.gitignore` will pull the file out of source control entirely, it won't keep any changes around.  Like there's no scenario in which `.env` is added to the `.gitignore` and `.env` isn't _also_ removed from source as part of that changeset (er, no good scenarios that is).

Is there a particular need for this that I'm missing?  The application will still just work for local/docker development, whether or not you have anything specified in either `.env` file.

vonbearshark: > The application will still just work for local/docker development, whether or not you have anything specified in either .env file.

Ah gotcha, so the variables in the env file are just overrides and it falls back to useable defaults without any env files?

JordanGuinn: Yes, it's purely for convenience.  If you want to turn on SQL logging or customize the DB pw used locally, you'd specify as much in your `.env` files.  The application will still work just fine if either file is empty/missing though.

JordanGuinn: I'll make that clear in the README

vonbearshark: Ahh yeah that is both great and then makes total sense for this change. You wouldn't really want all the vars populated in the env file in that case anyway 👍🏻

Issue #1394, opened 2021-07-09
Family income fields can't be changed during child create UX


JordanGuinn:
## Background
If a user is creating a new child record, and accidentally forgets to provide an annual household income for the child, or wants to update the household size after they've already specified one, the form flat-out prevents them from doing so.  It appears that editing any fields in the `Family Income` section of the child creation form after values have already been specified simply results in those fields reverting to their previous values.

## Repro Steps
- Log in to ECE Reporter
- Click the `Add a record` button
- Fill out the first three sections as normal
- Get to the `Family Income` section
- Fill out all fields except `Annual household income` which should be left blank
- Click `Save`
- Notice the error message for the empty field, functioning as appropriate
- Try to update any of the fields
- Click `Save` again
- Observe

## Expected Behavior
Fields are updated based on the new values specified.

## Actual Behavior
Fields revert to their previous values, preventing you from moving forward in the creation process.

PR #1479, opened 2021-09-13
Bug Fix: Couldn't save income determination after form error on create record

nickbristow:
## Background
In the add record flow, if you failed to put all information on the income determination section, correcting the form would not allow you to save the form. 

## GitHub Issue
#1394 

## Validation Plan

## Automated Testing


client/src/components/Forms/FamilyIncome/Form.tsx
```diff
@@ -128,12 +123,22 @@ export const FamilyIncomeForm: React.FC<FamilyIncomeFormProps> = ({
...
+
+    const saveData =
+      determination.id && !redetermine
+        ? updateDetermination
+        : createDetermination;
+
     saveData(updatedData)
       .then(afterSaveSuccess)
       .catch((err) => {
-        console.error(err);
```
vonbearshark, line 134: Do we want to rm these? Genuinely don't know. I do see it used elsewhere, and keeping console _error_ logging seems fine to me

vonbearshark (approved):

client/src/components/Forms/FamilyIncome/Form.tsx
JordanGuinn: Eh, I don't mind keeping them around personally

client/src/components/Forms/FamilyIncome/Form.tsx
```diff
@@ -128,12 +123,22 @@ export const FamilyIncomeForm: React.FC<FamilyIncomeFormProps> = ({
 
   const onSubmit = (updatedData: IncomeDetermination) => {
     setLoading(true);
+    if (!determination.id && inCreateFlow && dets[0])
```
JordanGuinn, line 126: What exactly is the fix here?

client/src/components/Forms/FamilyIncome/Form.tsx
nickbristow: I thought i added that in my debugging, adding back in

client/src/components/Forms/FamilyIncome/Form.tsx
nickbristow: if we don't have an id on the determination object and we are in the create flow and we have a determination on the child, it means the child has already had one created, so use that id.

JordanGuinn (approved):

JordanGuinn:
## Validation Plan
- [ ] Confirm I can **add** a new income determination to a child from the edit view
- [ ] Confirm I can **edit** an existing determination for a child from the edit view
- [ ] Confirm I can add a new income determination to a child from the create view, under the following conditions
   - [ ] When all fields are present
   - [ ] When certain fields are missing

JordanGuinn:
All looks good to me - this is ready

Issue #1380, opened 2021-06-17
Redesign initial upload page


JordanGuinn:
## Background
As part of our file upload redesign, the initial landing page for the upload process could use a little bit of love.  We should update both the copy and the styles of this page based on the [new designs developed and captured in Figma](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=21%3A0).

## Acceptance Criteria
- Initial file upload page is updated to match the designs captured below
   - Progress indicator is **removed**
   - `Back to Roster` button is added to the top of the page
- User is still able to select a file and go through the upload process without issue

## Designs
[The full set of Figma designs can be found here](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=21%3A0).

#### Current
![Screen Shot 2021-06-17 at 2.41.11 PM.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/e49712d3-b492-49f0-b203-6a68378e9b6b)

#### New
![Screen Shot 2021-06-17 at 2.41.39 PM.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/3fc7ea7f-63c0-4874-8d78-fbce14cfe2bf)

PR #1505, opened 2021-10-01
Redesign initial upload page

lukenewman:
## Background
This PR redesigns the initial upload page.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1380

## Associated PRs
https://github.com/ctoec/component-library/pull/170

## Validation Plan
* head to the upload page
* ensure designs match design specs
* ensure file upload still works

## Automated Testing

## Additional Context
N/A


client/src/containers/Upload/Upload.tsx
```diff
@@ -108,15 +107,15 @@ const Upload: React.FC<StepContentProps> = ({
   }
 
   return (
-    <div className="grid-container">
-      <BackButton location="/" />
+    <div className="upload-container">
```
bamader, line 110: All for the defining of a custom CSS class here to capture design changes, but I _think_ `grid-container` has specific attributes and properties for line heights, indents, spacing, etc. that makes it play nicely with grid rows later on. How  do we feel about having `div className="grid-container upload-container"` in the same line?

client/src/containers/Upload/UploadWizard.tsx
```diff
@@ -49,35 +50,30 @@ const UploadWizard: React.FC = () => {
...
-            &nbsp;to enter all of the required state-funded enrollment data.
-            When you’re finished, you can upload the complete template below.
-            After your data is in the system, you can also{' '}
-            <Link to="/roster">
-              use the Roster to make changes to your data
-            </Link>
-            .
-          </p>
+          <div className="instructions-header">
+            <div className="margin-right-10">
```
bamader, line 57: This seems excessive? We've never had to use a margin of 10 anywhere in the app. Have you tried putting this all in a  `grid-container` div? That carries some automatic margining with it.

client/src/containers/Upload/UploadWizard.tsx
```diff
@@ -49,35 +50,30 @@ const UploadWizard: React.FC = () => {
...
+              <h1 ref={h1Ref} className="margin-bottom-4 margin-top-6">
+                Upload your enrollment data
+              </h1>
+              <p>
+                Once you've filled out the{' '}
+                <Link to="/template">data collection template</Link>
+                &nbsp;— with all of the state funded enrollment required — you
```
bamader, line 64: Nit, typo: we want "...with all of the state-funded enrollment _data_ required" (data highlighted for emphasis, as it's currently missing).

bamader (approved):
LGTM aside from some minor margin things!

vonbearshark (approved):
Agreed with @bamader's feedback, but otherwise lgtm!

JordanGuinn:
Looks like tests are failing?  I imagine the snapshots just need to be updated

client/src/containers/Upload/UploadWizard.tsx
JordanGuinn: Agreed - some combination of [grid rows and grid columns](https://designsystem.digital.gov/utilities/layout-grid/) should be able to accomplish this same thing, while also maintaining a reasonable layout for the page.

JordanGuinn (approved):
Agreed with the sentiments in others' feedback, but beyond that LGTM 💯

client/src/containers/Upload/UploadWizard.tsx
lukenewman: hmmmm not sure how to get inter-column margins right

lukenewman:
looks like circleci can't install yarn packages?

JordanGuinn:
Try updating both of your branches

lukenewman:
updating how?

JordanGuinn:
Merge in `base` everywhere, then repin `component-library` here.  Your lockfile looks to be out of date

bamader:
I think he means pulling in the latest version of base, doing fresh yarn installs based on that, etc.

bamader:
Looks like the only remaining thing that's failing are snapshot tests--almost there!

lukenewman:
so weird i already did `yarn test -u`

bamader:
I would guess pulling in base and re-pinning component library brought in some back-end stylistic things that we can't see but that are technically different in rendering terms.

bamader:
Imma validate this bad boi now

bamader:
## Validation Plan
- [ ] Progress indicator removed from upload page
- [ ] Back to roster button present
- [ ] Designs generally match new specs, including image
- [ ] User can upload correct file without issue
- [ ] User can upload file of wrong type (not a csv) and it's handled elegantly
- [ ] User can upload incomplete/missing info file and it's handled elegantly

bamader:
Validations complete. I think the only thing off is the margin around the header text on the upload page itself ("Upload your Enrollment Data"). Looks like it's currently set to `margin-top-6` and `margin-bottom-4`, but comparing this to the rest of the app I think that breaks with our existing format. Elsewhere on pages with this kind of header layout, we put `margin-top-4` in the div that has the `grid-container` containing the header. The header itself we give `margin-top-0` so that it doesn't overly stack spacing. Neither one gets any `margin-bottom` styling. And then below that, the default `<p>` tag has a top margin of 1  which is a clean line break. Not a blocker by any means but it was pretty noticeable to me so wanted to point it out!

lukenewman:
@bamader i think the difference here is that the "Back to Roster" button is inside the `grid-container`, so we need to add that margin in between that and the `h4`. i played around with those guidelines you outlined but couldn't get it right without adding in some margins

bamader:
Gotcha, that makes sense. How does it look if we change the `margin-top-6` to a `top-4` and the `margin-bottom-4` to a `bottom-2`? Just trying to artificially create the same aesthetic as on other pages. If these number changes don't help then we can just use it as is.

lukenewman:
Screen Shot 2021-10-01 at 4 07 38 PM

lukenewman:
now i'm thinking the icon should be sized-down a bit

bamader:
Yup agreed! That spacing looks good to me so let's shrink the image a bit.

lukenewman:
eh?

Screen Shot 2021-10-01 at 4 23 36 PM

bamader:
That looks good to me! @JordanGuinn Any thoughts on the various design screenshots?

JordanGuinn:
Looks perfect to me! 👍 👍

bamader:
Great! Sounds like with the new design updates, this should be good to go!

PR #1515, opened 2021-10-14
Add site filter

bamader:
## Background
This PR adds site filtering to the roster page, following the same structure and pattern so helpfully laid out by Lucas.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1474

## Associated PRs
None

## Validation Plan
- [x] Confirm site filter works for one or more sites selected
- [x] Confirm that selecting all sites shows whole roster
- [x] Confirm that site filter cooperates with age group filter
- [x] Confirm that site filter cooperates with space type filter
- [x] Confirm that site filter cooperates with funding type filter
- [x] Confirm that a multi-org user only sees sites associated with the roster org page they're on

## Automated Testing
N/A

## Additional Context
Waiting on 1348 to land before marking this ready for review, as I'll need to make some changes so that the site filter only shows sites at the current org. But at least putting this up now in case people have thoughts on manually setting the sites in what's otherwise a const (which I don't love, but is the easiest way to get the info there).


client/src/containers/Roster/Roster.tsx
```diff
@@ -138,6 +138,10 @@ const Roster: React.FC = () => {
     }
   }, [user, userOrg, accessToken]);
 
+  // Dynamically set the sites that appear in the filter
+  // drop down, based on the current org
+  FILTERS[4].items = sites.map((s) => { return { id: s.siteName, label: s.siteName } });
```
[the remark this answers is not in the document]
nickbristow: my prettier likes the `(s)` but i removed the return!

client/src/containers/Roster/Roster.tsx
```diff
@@ -133,11 +133,15 @@ const Roster: React.FC = () => {
     if (user && userOrg && accessToken) {
       (async function getSites() {
         const sites = await apiGet('sites', accessToken);
-        setSites(sites);
+        setSites(sites.filter((s: Site) => s.organizationId === userOrg.id));
```
JordanGuinn, line 136: Can we instead just give `userOrg.id` to the `sites` API?  I always prefer permissions filtering like this to exist server-side over client-side; plus that API already has support for an `organizationId` parameter anyway

JordanGuinn: @nickbristow ^ping 🙂

PR #1525, opened 2021-10-22
Back-End Unit Testing Additions

JordanGuinn:
## Background
This PR includes, among other things, a handful of unit tests demonstrating how we might use Jest to add test coverage to all **functional units** we have in our application.  Moving forward, if there are any utilities we come across that are difficult to write unit tests for, that should prompt us to immediately ask questions like:
- Is this doing too many things?
- Can this be decomposed into smaller pieces?
- Are there other functional units that already exist we can take advantage of instead?

Very much curious to hear other folks' thoughts and opinions on unit testing strategy moving forward, but in general, I believe our code base could very much benefit from more coverage like this, and I _know_ it's doable.

Additionally, this PR also includes:
- Consolidating a few utilities that were duplicated across client and server applications
- Renaming a few utility files to encompass multiple functions of similar type
   - We should _really_ stop the previous practice of creating a brand new file for each utility function we write
- Replacing a few hand-rolled utilities with their equivalents in `lodash` and/or `underscore`

## GitHub Issue
N/A

## Validation Plan
TBD

## Automated Testing


bamader (approved):
My first reaction when I saw this was abject terror, but I guess it doesn't bring back nightwatch so it can't be all bad, right? Right??

JordanGuinn:
@bamader Oh we will **never** be bringing nightwatch back 😅 This is purely unit testing additions, which we implement with [jest](https://jestjs.io/) (a framework I'm generally a _really_ big fan of)

lukenewman (approved):

bamader:
@JordanGuinn Given that this PR is basically its own validation via unit tests, I'm not really sure there's anything left for external validation? It has two approving reviews, so I'd advocate for just merging this in.

JordanGuinn:
Yeah, that's fair.  The majority of the non-unit test additions here are really just me moving stuff around anyway, the likes of which would be caught during the build process.

Will merge in.

Issue #1481, opened 2021-09-14
DRAFT - Add top-level server check for all admin routes


JordanGuinn:
## Background
As we do in the client currently, we should build some sort of top-level admin user check in all routes that should only be available to administrative users.  It should also not be a one-off check that needs to be applied to each and every route.

## Acceptance Criteria
Please supply the full set of predefined requirements that must be met in order for this ticket to be completed.

## Technical Specs
Specify any **technical** requirements that need to be met in in order to mark this ticket as done.

PR #1528, opened 2021-10-26
Create top-level admin only routing utility

JordanGuinn:
## Background
Was in this area briefly today, and figured I might as well unify our back-end admin routing in all the places that it's currently being used.  Now, we can simply include this new utility function when specifying the callbacks for a given route, instead of needing to include it directly as part of the actual business logic.

## GitHub Issue
- https://github.com/ctoec/data-collection/issues/1481

## Associated PRs
N/A

## Validation Plan
- [x] Confirm I can continue accessing the application as an admin user
- [x] Confirm I can continue accessing the application as a standard user
- [x] Confirm admin-level routes throw 403 errors when inappropriately accessed
- [x] Confirm other HTTP errors are thrown as they were previously

## Automated Testing


bamader (approved):
This makes sense to me!

src/middleware/error/handleError.ts
```diff
@@ -1,30 +1,20 @@
...
+import { Request, Response, NextFunction, ErrorRequestHandler } from 'express';
 
 /**
  * Error handling middleware to log errors, and send the appropriate
  * error response.
  */
-export const handleError = (
-  err: Error,
+export const handleError: ErrorRequestHandler = (
+  err: any,
```
jessethach, line 8: Would it help to extend the `Error` type? Like so:

```
interface ResponseError extends Error {
  status?: number;
}
```
Then:

```
...
err: ResponseError,
...
```

jessethach (approved):
Looks good! Just had a comment on a type.

src/middleware/error/handleError.ts
JordanGuinn: I only hesitate to do that because the [function signature of Express' `ErrorRequestHandler`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/507e9d3c8e48b2db81d3e071d270a2a297eed151/types/express-serve-static-core/index.d.ts#L32) explicitly has the `err` parameter defined as type `any`.  Which I _think_ makes sense, as there _could_ be potential error states that we may encounter that we're not currently catching and propagating up with an HTTP error response.

jessethach: Sounds good!

bamader:
## Validation Plan
- [ ] Confirm I can continue accessing the application as an admin user
- [ ] Confirm I can continue accessing the application as a standard user
- [ ] Confirm admin-level routes throw 403 errors when inappropriately accessed
- [ ] Confirm other HTTP errors are thrown as they were previously

JordanGuinn:
> So, for none of the admin-only routes are any errors being thrown. I'm logged in as Voldemort, and when I type any of the three above admin routes into the URL, I get a 200 network response status code and I'm just redirected back to /home. Is that desired? If not, I'm not seeing any errors ever get thrown.

That is to be expected - the React app is redirecting the non-admin user to a page they have access to prior to kicking off any server requests, so we don't even get a chance for the route to fail.  We could `curl` those APIs directly, or temporarily tweak the client code slightly to allow for non-admins to access admin views, but if we can't access them through the app, that honestly might be acceptable enough for testing purposes.  Until we bring automated API tests into the app, that is 😎

bamader:
In that case I'm satisfied this works as intended. We can revisit after we bring API testing in, but I feel like curling the URLs is probs excessive if there's no way someone could actually get there lol.

Issue #1533, opened 2021-10-29
Certain boolean/birth certificate values are not accepted during file upload


JordanGuinn:
## Background
This was brought to our attention as part of a recent customer support request.  It looks like some of the boolean fields we allow for in file uploads don't support all of the potential values that are said to be supported on the data requirements page (namely `Y` and `N`).  Additionally, the birth certificate field doesn't actually support `US` during upload like the application says; instead, users must explicitly type out `US Birth Certificate` when uploading.

We should ensure that our file upload API actually supports all the potential values that our application claims we do.

## Acceptance Criteria
- Each boolean field available during file upload accepts all values shown on the data requirements page
   - `Y`, `Yes`, `N`, `No` and `Not collected`
- Birth certificate type accepts all values shown on the data requirements page
   - `US`, `Non-US` and `Unavailable`
- Unit tests are written capturing the behavior of all scenarios above
- Uploads can still be submitted successfully

PR #1534, opened 2021-10-29
Fix boolean/birth certificate field support in file upload

JordanGuinn:
## Background
This PR ensures that all boolean and birth certificate fields accept all values that are displayed in the data requirements page of the application, and additionally adds unit tests confirming as much.

## GitHub Issue
- https://github.com/ctoec/data-collection/issues/1533

## Associated PRs
N/A

## Validation Plan
TBD (Will need to be thorough)

## Automated Testing
- [x] If applicable, unit tests have been added to cover this changeset.


jessethach (approved):
Great work! Perfect use of unit testing.

src/controllers/enrollmentReports/parse/parseUtils.ts
```diff
@@ -94,74 +103,83 @@ export function getSheetType(sheet: WorkSheet): 'xlxs' | 'csv' {
...
-export function getSpecialProperties() {
-  const booleanProperties: string[] = [];
-  const dateProperties: string[] = [];
-  Object.entries(new EnrollmentReportRow()).forEach(([prop, value]) => {
-    if (typeof value === 'boolean') {
-      booleanProperties.push(prop);
-    }
-    if (moment.isMoment(value)) {
-      dateProperties.push(prop);
+export function parseEnrollmentReportRow(
```
bamader, line 106: nit: can we give this bad boy some docstrings?

src/controllers/enrollmentReports/parse/parseUtils.ts
```diff
@@ -94,74 +103,83 @@ export function getSheetType(sheet: WorkSheet): 'xlxs' | 'csv' {
...
+  enrollment: EnrollmentFromUpload
+): EnrollmentReportRow {
+  Object.entries(enrollment).forEach(([key, value]) => {
+    if (ENROLLMENT_BOOLEAN_KEYS.includes(key)) {
+      enrollment[key] = getBoolean(value);
```
bamader, line 111: _BIG_ fan of these decomposed utilities. This function is hella readable _because_ of them.

src/template/EnrollmentReportRow.ts
```diff
@@ -27,6 +28,57 @@ import {
...
+  homelessness: null,
+  incomeNotDisclosed: null,
+  nativeHawaiianOrPacificIslander: null,
+  raceNotDisclosed: null,
+  receivesDisabilityServices: null,
+  white: null,
+};
+export const ENROLLMENT_BOOLEAN_KEYS: string[] = Object.keys(
+  ENROLLMENT_BOOLEAN_FIELDS
+);
```
bamader, line 49: Curious on the reason for setting up both of these. Any reason to not just manually create the string array containing the keys we want? Since we have to "hardcode" the variables anyway in `ENROLLMENT_BOOLEAN_FIELDS` it doesn't really seem to save us anything.

src/utils/polyfills.ts
```diff
@@ -0,0 +1,4 @@
+//  https://github.com/microsoft/TypeScript/issues/16350
+export type KeysOfType<T, U> = {
+  [P in keyof T]: T[P] extends U ? P : never;
+}[keyof T];
```
bamader, line 4: It may be because I'm still exhausted from DASH, but _wat_. Can we slap a comment here or something?

src/controllers/enrollmentReports/parse/parseUtils.test.ts
```diff
@@ -0,0 +1,386 @@
...
+      expect(getBoolean('n')).toEqual(false);
+      expect(getBoolean('  n    ')).toEqual(false);
+    });
+
+    it('converts "Yes" to "true"', () => {
+      expect(getBoolean('Yes')).toEqual(true);
+      expect(getBoolean('YES')).toEqual(true);
+      expect(getBoolean('yes')).toEqual(true);
+      expect(getBoolean('yES')).toEqual(true);
+      expect(getBoolean('YEs')).toEqual(true);
```
bamader, line 286: How fun was it typing all these permutations 😂

bamader (approved):

src/controllers/enrollmentReports/parse/parseUtils.test.ts
JordanGuinn: SO MUCH FUN 😭

src/template/EnrollmentReportRow.ts
JordanGuinn: It gives us type safety - because the `EnrollmentBooleanFields` type is based on every boolean or `UndefinableBoolean` attribute on the `EnrollmentReportRow`, the `ENROLLMENT_BOOLEAN_FIELDS` variable will throw an error if any boolean attribute is ever added or removed from `EnrollmentReportRow`.  Which would _force_ us to update this accordingly with potential schema changes, something we don't get if we simply hardcode all the boolean keys.

src/utils/polyfills.ts
JordanGuinn: HEH - sure thing!  It's just a generic for creating a TS interface based on the all of the attributes of a given `type` - i.e. if you want an interface for all the `string` fields on the `Child` interface.

The current comment points to the GH link from which I pulled this, but I'm happy to spell it out directly as well

lukenewman (approved):

bamader:
## Validation Plan

File upload properly handles each of the following birth certificate formatting cases:
- [ ] birth cert US, all three US fields supplied
- [ ] birth cert US, missing birth cert ID
- [ ] birth cert US, missing birth town
- [ ] birth cert US, missing birth state
- [ ] birth cert US birth certificate, all three US fields supplied
- [ ] birth cert US birth certificate, missing birth cert ID
- [ ] birth cert US birth certificate, missing birth town
- [ ] birth cert US birth certificate, missing birth state
- [ ] birth cert Non-US, ID supplied
- [ ] birth cert Non-US, town or state supplied
- [ ] birth cert Non-US, no other birth cert fields supplied
- [ ] birth cert Non-US birth certificate, ID supplied
- [ ] birth cert Non-US birth certificate, town or state supplied
- [ ] birth cert Non-US birth certificate, no other birth cert fields supplied
- [ ] birth cert Unavailable
- [ ] birth cert field omitted completely

File upload properly handles each of the following boolean columns for all 5 variations of input (Yes, Y, No, N, Not collected):
- [ ] americanIndianOrAlaskaNative
- [ ] asian
- [ ] blackOrAfricanAmerican
- [ ] dualLanguageLearner
- [ ] foster
- [ ] hispanicOrLatinxEthnicity
- [ ] homelessness
- [ ] incomeNotDisclosed
- [ ] nativeHawaiianOrPacificIslander
- [ ] raceNotDisclosed
- [ ] receivesDisabilityServices
- [ ] white

bamader:
Okay found something that needs further investigation. There are four fields that do not accept the new boolean parsing: foster, receivesDisabilityServices, hispanicOrLatinxEthnicity, and dualLanguageLearner. In the case of these 4, the values for truthiness (Yes, Y) do _not_ correctly save. Instead, they're parsed as missing from the sheet. I did some digging and I found that in `src/controllers/enrollmentReports/map/entities/child` lines 30-56, these are the _only_ four boolean values that receive special case handling for the Undefinable Boolean Enum. I don't know exactly what's causing the load to fail, but I guarantee there's a connection there. Jordan, for your revision passes / subsequent testing, you can use this spreadsheet (which enumerates all combinatorial possibilities of the new booleans and birth cert values you enabled).

[Field_Support_test.xlsx](https://github.com/ctoec/data-collection/files/7477521/Field_Support_test.xlsx)

JordanGuinn:
@bamader Ahhhhh, good find.  Turns out that we intentionally have to handle booleans differently from `UndefinableBooleans`, as the latter are actually strings. 😅 

Fix pushed, and unit tests updated.

bamader:
Everything is green on my end -- this is ready to merge!

Issue #1365, opened 2021-06-03
Add batch delete to roster view


JordanGuinn:
## Background
Providers often take actions on multiple records, particularly at the end of the school year.  Including batch delete for multiple records would allow users to complete their tasks more efficiently.  Thankfully, we should be able to repurpose a good deal of the single record UX we already have in place for this type of action here as well.

## Acceptance Criteria
- `Delete` button is added to the `# of children selected` banner, whenever **1 or more child rows are selected**
- Clicking the `Delete` button pops a modal similar to the delete modal currently available when editing a single child record
   - Title should simply replace **child's name** with **# of child rows to be deleted**
   - `Record` should become `Records` throughout the language of the modal
- **Batch delete API** is written to handle multiple deletes simultaneously
- Clicking `Yes, delete records` in the batch delete modal invokes this new API and successfully removes all selected records
- On success, alert is displayed in the roster, indicating the number of child rows successfully deleted
- If there are issues deleting specific children selected within the batch, all other children are still deleted
   - User should see an alert indicating which children were not successfully deleted

## Designs
[Figma designs can be found here](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=275%3A34).

#### Current Child Delete Modal
![Screen Shot 2021-06-03 at 9.48.37 AM.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/4bfacbc8-a711-4d2f-8d10-6929373582fe)

#### Roster View
![Roster - Batch Actions.png](https://images.zenhubusercontent.com/6050f59bb12f6b5f4da9aeb1/07061ff2-b798-4192-94be-363a5cabbe0e)

PR #1536, opened 2021-10-29
Add batch delete to roster view

lukenewman:
## Background
This PR adds the batch delete action to the roster redesign.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1365

## Associated PRs
N/A

## Validation Plan

## Automated Testing

## Additional Context
N/A


client/src/containers/Roster/BatchDelete.tsx
```diff
@@ -0,0 +1,121 @@
...
+          text={`Delete ${pluralize('record', children.length)}`}
+          Icon={Trash}
+          iconSide={'right'}
+        />
+      </Button>
+      <Modal
+        isOpen={isOpen}
+        onModalClose={toggleIsOpen}
+        // Shorten to accomodate modal close button in a
+        // nice looking way
```
bamader, line 85: Unrelated to this PR, but feels like we can delete this comment now, since it addressed a designer's feedback from like a year and a half ago that's now completely irrelevant lol

bamader (approved):
Looks good!

client/src/containers/Roster/BatchDelete.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const { accessToken } = useContext(AuthenticationContext);
+  const [isDeleting, setIsDeleting] = useState(false);
+
+  function deleteRecords() {
+    setIsDeleting(true);
+    const query = stringify({ childrenIds: children.map((c) => c.id) });
+    apiDelete(`children?${query}`, { accessToken, jsonParse: true })
+      .then((res) => {
+        toggleIsOpen();
+        if (res.length) {
```
jessethach, line 37: Might want the check to be more than 0 since the body returns an empty array by default.

jessethach (approved):
Looks good! Just had a couple of comments.

client/src/index.scss
```diff
@@ -40,6 +26,21 @@ $css--plex: true;
 @import 'carbon-components/scss/components/pagination/pagination';
 @import 'carbon-components/scss/components/ui-shell/ui-shell';
 
+/*********************************************************************
+ *** Custom Application-Specific Component Styles
+ *********************************************************************/
+@import './components';
```
JordanGuinn, line 32: Note that this could have **ALL KINDS** of stylistic implications, for whoever tests this

src/controllers/children/delete.ts
```diff
@@ -21,15 +24,25 @@ export const deleteChild = async (id: string, user: User) => {
...
   }
 
-  await getManager().softRemove(child);
-  // For now, also remove family since we have a 1-to-1 child-family relationship
-  // TODO: remove this when we do family look ups
-  await getManager().softRemove(child.family);
+  let errorChildren = [];
+  for (const child of children) {
+    try {
+      await getManager().softRemove(child);
```
JordanGuinn, line 37: Can we throw both of these `softRemove` calls into a single transaction per child?  This seems like it should be an all-or-nothing type of call

src/controllers/children/delete.ts
```diff
@@ -4,14 +4,17 @@ import { User, Child } from '../../entity';
...
 
 /**
- * Delete child record, if user has access
- * @param id
+ * Delete childrens' records, if user has access
+ * @param ids
  * @param user
  */
-export const deleteChild = async (id: string, user: User) => {
+export const deleteChildren = async (
```
JordanGuinn, line 11: Is this backwards compatible with the `Delete child` action, currently available from the `Edit Record` view for a given child?

src/routes/children.ts
```diff
@@ -68,13 +68,19 @@ childrenRouter.get(
...
+ *
+ * Returns an array of ids of children that could not be deleted
  */
 childrenRouter.delete(
-  '/:childId',
+  '/',
   passAsyncError(async (req, res) => {
-    await controller.deleteChild(req.params['childId'], req.user);
-    res.sendStatus(200);
+    const childrenIDs = req.query['childrenIds'] as string[];
```
JordanGuinn, line 78: We should probably handle the case that `childrenIds` is `null` or `''` here, and throw a 400

client/src/containers/Roster/Roster.tsx
```diff
@@ -143,6 +145,7 @@ const Roster: React.FC = () => {
   // drop down, based on the current org
   FILTERS[4].items = sites.map((s) => ({ id: s.siteName, label: s.siteName }));
 
+  const [children, setChildren] = useState<RosterChild[]>([]);
```
JordanGuinn, line 148: Why do we need to maintain two separate states of all the children in the roster?

client/src/containers/Roster/Roster.tsx
```diff
@@ -57,9 +57,10 @@ import {
 
 import { ReactComponent as Change } from '../../images/change.svg';
 import { ReactComponent as Withdraw } from '../../images/withdraw.svg';
-import { ReactComponent as Trash } from '@ctoec/component-library/dist/assets/images/trashCan.svg';
+import { BatchDelete } from './BatchDelete';
 
 interface ChildRow extends ObjectWithValidationErrors {
+  child: RosterChild;
```
JordanGuinn, line 63: Why does a `ChildRow` include an instance of the `RosterChild` it was generated from?

src/controllers/children/delete.ts
lukenewman: hmmm would this need to be raw SQL?

JordanGuinn: Not at all - TypeORM's EntityManager has a `transaction` function you can call, which takes a callback containing all the commands you'd like to operate within said transaction.  So something like:

```
await getManager().transaction(async (transaction) => {
   await transaction.softRemove(something);
   await transaction.softRemove(somethingElse);
});
```

JordanGuinn: That'll commit both statements if they both succeed, or roll them both back if either throw an error.

JordanGuinn (approved):

nickbristow:
## Validation 

- [ ] Selecting one child and selecting delete brings up delete modal
  - [ ] Delete modal has name of child
  - [ ] Selecting `No, cancel` does not remove child
  - [ ] Selecting `Yes, delete record` removes child
  - [ ] Success alert displayed on removal
- [ ] Selecting multiple children and selecting delete brings up delete modal
  - [ ] Delete modal has number of selected children
  - [ ] Selecting `No, cancel` does not remove records
  - [ ] Selecting `Yes, delete records` removes children
  - [ ] Success alert displayed on removal

nickbristow:
@lukenewman 
Couple of notes from validation
- When selecting a single child, it displays `Name Missing` on the modal and success alert
- After deletion, the multiselect options are still up with 0 selected allowing you to try and delete 0 records, which results in an error

lukenewman:
@nickbristow this is ready for re-validation

nickbristow:
🚀 G2G
Validated

PR #1500, opened 2021-09-29
1348 multi org landing page roster redesign

nickbristow:
## Background
Adds multi org landing page. Limits roster view to a single org.

## GitHub Issue
#1348 

## Associated PRs
If applicable, replace this text with links to additional pull requests associated with this work.

## Validation Plan
Please outline the steps you've already taken to validate this body of work, along with any additional steps that need to be taken in order to complete testing.

## Automated Testing
- [ ] All unit tests are passing.
- [ ] If applicable, unit tests have been added to cover this changeset.

## Additional Context
If there's anything additional information that those reviewing or validating this work should be aware of, please specify it here.

Issue #1275, opened 2021-04-13
Users cannot have varying access levels across different organizations


jlhogan:
## Background
Confirmed this is still an issue in `EPIC-roster-redesign` - permissions are not currently respected for users that have both organization level and site level permissions across multiple sites.

## Repro Steps
- Log into ECE Reporter
- Create a new user (or edit an existing user)
- Provide this user with **organization-level permission** to one organization
- Provider this user with **site-level permissions** to a **subset** of sites at **another** organization
- Log in to ECE Reporter as this new user
- Observe

## Expected Behavior
- Org list view shows roster counts limited to **only the sites the user has access to**
- Roster for org-level permissions includes enrollments for all sites at that org
- Roster for site-level permissions **only includes enrollments for the sites the user has access to** at that org

## Actual Behavior
- Org list view shows roster counts across all sites per org
- Roster for org-level permissions includes enrollments for all sites at that org
- Roster for site-level permissions includes enrollments for all sites at that org

## Acceptance Criteria
- Site-level access is respected for **all** users, regardless of how many organizations those site permissions span
- Organization-level access is respected for **all** users, regardless of how many organizations they have access to
- Site-level access of **one** organization and organization-level access of **another** organization are respected simultaneously

JordanGuinn:
@JordanGuinn Be sure to account for this in our upcoming new roster world

PR #1538, opened 2021-11-01
Remove condition that captures accessType flag

jessethach:
- The accessType property was gating the call from filtering siteIds.

## Background
Sites restrictions across various orgs were not respected after Admin user sets site level permissions. Also added a property on to the User type since it was missing.

## GitHub Issue
#1275 

## Validation Plan

- [ ] Log into ECE Reporter
- [ ] Create a new user (or edit an existing user)
- [ ] Provide this user with organization-level permission to one organization
- [ ] Provider this user with site-level permissions to a subset of sites at another organization
- [ ] Log in to ECE Reporter as this new user

## Automated Testing

## Additional Context
Related to changes introduced in #1500


client/src/shared/models/db/User.ts
```diff
@@ -12,6 +12,7 @@ export interface User {
   email?: string;
   isAdmin: boolean;
   sites?: Array<Site>;
+  sitePermissions?: Array<SitePermission>;
```
bamader, line 15: Do we not need an affiliated variable in the entity schema in `src/`?

bamader (approved):
LGTM. As a general nit type of comment, can you try to make PR titles more indicative of the issue the PR is solving? It could be as simple as just using all or part of the header given to the associated ticket. It's a very small thing, but I think it's helpful when first opening up a PR for review to know exactly what the PR is going to do and why it matters without having to read extra context.

jessethach:
@bamader No prob! Will make more descriptive PR titles moving forward.

client/src/shared/models/db/User.ts
jessethach: I believe it's already there. May have to check though, I may be missing nuance here.

JordanGuinn (approved):

lukenewman (approved):

bamader:
## Validation Plan
Create an ECE Reporter user with org-level permissions at one organization, and site-level permissions at another. Log in as this user.

- [ ] App allows login of user with varied permissions across orgs
- [ ] Roster shows correct view depending on selected org (all sites for org-access org, some sites for site-access org)
- [ ] User can perform all org-level operations at the org-access organization, for all sites
   - [ ] Create children
   - [ ] Upload sheets
   - [ ] Delete children
   - [ ] Change enrollments
   - [ ] Withdraw children
   - [ ] Update records
- [ ] User can perform all site-level operations at the site-access organization, for the sites that the user has access to
   - [ ] Create children
   - [ ] Upload sheets
   - [ ] Delete children
   - [ ] Change enrollments
   - [ ] Withdraw children
   - [ ] Update records

bamader:
Validations for all functionality in this PR are good to go. However, I found two other bugs/weird interactions which I think we should open tickets for. Jordan, I'll let you make the final call on whether it's worth addressing those, so I'll just leave them in my validation comments. After you check them out, this should be good to merge.

JordanGuinn:
Both of those issues are certainly worth tracking in Zenhub, and then we can prioritize them from there.  Will create tickets for em both, and in the meantime we can merge this

JordanGuinn:
> I noticed that in the create flow (and potentially anywhere there are dates in the app), we require a format of MM/DD/YYYY (emphasis on the two digits for MM and DD). However, nowhere is this actually indicated to a user. I realized this when I tried to type in 1/1/2021 for a birthdate when creating a child, and I got the "Unable to save child identifiers" error with no indication that it was the date format that had blocked the save. We should either indicate required formatting, accept more date formats, or provide more specific error copy so that this case doesn't confuse users who don't understand why their data isn't being submitted.

It looks like we're already indicating this to me?
Screen Shot 2021-11-06 at 1 44 39 PM

JordanGuinn:
> It looks like withdrawn children are handled kind of weirdly in the roster redesign branch (see screenshot for example). They appear in on the roster page, but with no information other than their name listed in the roster table. They count towards the bolded heading for total number of enrollments, but don't count for the values in the cards at the top, leading these quantities to differ when they should be the same. Finally, when you click on them in the roster, they show up as having an Enrollment and Funding error with missing info. Just seems like we need to make some design decisions around how to more consistently handle withdrawn kids.

Tickets have been created accordingly.  Turns out the second issue is just us no longer displaying any past enrollments on the child detail view anymore.

- #1554 
- #1555

Issue #1554, opened 2021-11-06
Withdrawing children doesn't remove them from the current roster


JordanGuinn:
## Background
It looks like the withdraw action is resulting in some strange behavior in the roster list.  On successful withdrawal, it looks like the withdrawn child actually remains in the current roster (which should only be displaying children with **active** enrollments), but all fields except `name` have been removed.

It should be noted that all of the other fields we display in the roster list view currently exist on the `Enrollment` entity, not the `Child` entity, so this may be because we're returning children in the roster list **whether or not they have an active enrollment**.

## Repro Steps
- Navigate to your roster
- Go to a child detail view
- Click the `Withdraw` button
- Fill out all the fields in the modal, and submit the withdrawal
- Get navigated back to the roster and observe

## Expected Behavior
- The withdrawn child is removed from the current roster, and the total count is updated

## Actual Behavior
- The withdrawn child is kept in the current roster with all fields outside of name removed, and the total count is not updated

## Demo
https://images.zenhubusercontent.com/274936723/97c02d31-2cac-4fe3-bb34-5a5a4939fdcf/screen_recording_2021_11_06_at_1_50_13_pm.mov

PR #1560, opened 2021-11-12
BUG: Past enrollments not displaying on edit enrollments page

nickbristow:
## Background
On roster redesign, past enrollments not displaying on child edit enrollments page

## GitHub Issue
#1555 

## Associated PRs
If applicable, replace this text with links to additional pull requests associated with this work.

## Validation Plan
Please outline the steps you've already taken to validate this body of work, along with any additional steps that need to be taken in order to complete testing.

## Tech Specs
- [ ] If any TypeORM entities were created/updated/removed, the necessary migration(s) were [correctly auto-generated](https://github.com/typeorm/typeorm/blob/master/docs/using-cli.md#generate-a-migration-from-existing-table-schema) and **not manually written**.

## Automated Testing
- [ ] All unit tests are passing.
- [ ] Unit tests have been added to cover this changeset.

## Additional Context
If there's any additional information that those reviewing or validating this work should be aware of, please specify it here.

PR #1562, opened 2021-11-15
Adjust query builder to optionally include enrollments with exit date.

jessethach:
- Add onlyEnrolled flag to queryBuilderBuilder to query by enrollment exit optionally.
- The withdraw component calls getChildById that leverages the query builder. This page on the ui does not need to filter by exit date.

## Background
It looks like the withdraw action is resulting in some strange behavior in the roster list. On successful withdrawal, it looks like the withdrawn child actually remains in the current roster (which should only be displaying children with active enrollments), but all fields except name have been removed.

It should be noted that all of the other fields we display in the roster list view currently exist on the Enrollment entity, not the Child entity, so this may be because we're returning children in the roster list whether or not they have an active enrollment.

## GitHub Issue
#1554 

## Validation Plan

**Make sure to upload mock roster data in order to follow validation steps.**

- [ ] Navigate to Roster page and select a child. Click Withdraw on the top right corner, triggering a modal. Fill out all fields and submit. 
- [ ] Confirm in the network tab that GET child by id endpoint does not throw 500.
- [ ] Confirm that the withdraw endpoint is successful.
- [ ] Navigate back to roster list and confirm the number of children enrolled matches what is expected (-1).
- [ ] Confirm that the withdrawn child is not showing up in roster list after withdrawal.

## Automated Testing


JordanGuinn (approved):

JordanGuinn:
LGTM!  Though we are probably gonna wanna unify on how we handle the retrieval of previous enrollments, as the [child detail view bugfix PR](https://github.com/ctoec/data-collection/pull/1560) has something very similar implemented slightly differently

nickbristow (approved):

nickbristow:
## Validation
- [ ] Navigate to Roster page and select a child. Click Withdraw on the top right corner, triggering a modal. Fill out all fields and submit.
- [ ] Confirm in the network tab that GET child by id endpoint does not throw 500.
- [ ] Confirm that the withdraw endpoint is successful.
- [ ] Navigate back to roster list and confirm the number of children enrolled matches what is expected (-1).
- [ ] Confirm that the withdrawn child is not showing up in roster list after withdrawal.

nickbristow:
🚀 good to go

lukenewman (approved):

Issue #1364, opened 2021-06-03
Add batch withdraw to roster view


JordanGuinn:
## Background
Providers often take actions on multiple records, particularly at the end of the school year.  Including batch withdraw for multiple records would allow users to complete their tasks more efficiently.  Thankfully, we should be able to repurpose a good deal of the single record UX we already have in place for this type of action here as well.

## Acceptance Criteria
- `Withdraw` button is added to the `# of children selected` banner, whenever **1 or more child rows are selected**
- Clicking the `Withdraw` button pops a modal similar to the equivalent modal currently available when withdrawing a single child record
   - Title replaces child's name with # of children to be withdrawn
   - All enrollment data is removed from the top of the modal
- **Batch withdraw API** is written to handle multiple withdrawals simultaneously
- Clicking `Withdraw` in the batch withdraw modal invokes this new API and successfully withdraws all selected records
- On success, alert is displayed in the roster, indicating the number of child rows successfully withdrawn
- If there are issues withdrawing specific children selected within the batch, all other children are **still withdrawn**
   - User should see an alert indicating which children were not successfully withdrawn

## Designs
[Figma designs can be found here](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=275%3A34).

#### Current Withdraw Modal
![Screen Shot 2021-06-03 at 10.04.10 AM.png](https://images.zenhubusercontent.com/5ff504ac246966652b569927/158bd87a-39fc-4e26-b179-2a529d5a8092)

#### Roster View
![Roster - Batch Actions.png](https://images.zenhubusercontent.com/6050f59bb12f6b5f4da9aeb1/07061ff2-b798-4192-94be-363a5cabbe0e)

PR #1565, opened 2021-11-16
Add batch withdraw to roster view

lukenewman:
## Background
This PR adds the final batch action, Withdraw, to the roster redesign.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1364

## Associated PRs
[This PR](https://github.com/ctoec/data-collection/pull/1562) may need to be merged in before this one. Once withdrawn, records are still showing up in the roster but without any enrollment data.

## Validation Plan
* [ ] - on the roster page, select one record and click Withdraw, fill out the form, hit Withdraw, record should be withdrawn
* [ ] - on the roster page, select multiple record and click Withdraw, fill out the form, hit Withdraw, records should be withdrawn

## Tech Specs

## Automated Testing

## Additional Context
N/A


lukenewman:
merged in jesse's recent enrollment pr, and the old problem has been fixed (children showing up in roster without enrollment data). however, now when selecting multiple children, not all of them are being withdrawn. investigating...

nickbristow (approved):
LGTM 

Edit: just need to fix unused var alerts

lukenewman:
this is good to go

client/src/containers/Roster/BatchDelete.tsx
```diff
@@ -96,8 +96,8 @@ export const BatchDelete: React.FC<DeleteProps> = ({
         header={
           <h2>{`Delete enrollment for ${
             isOneChild
-              ? Object.values(selectedChildren)[0]
-              : Object.keys(selectedChildren).length + ' records'
+              ? Object.values(children)[0]
+              : Object.keys(children).length + ' records'
```
jessethach, line 100: Nit: Any reason to not place in template string like the others?

client/src/containers/Roster/BatchWithdraw.tsx
```diff
@@ -0,0 +1,141 @@
...
+        content={
+          <>
+            <Form<WithdrawRequest>
+              onSubmit={onSubmit}
+              data={{} as WithdrawRequest}
+              className="usa-form"
+            >
+              <ExitReasonField<WithdrawRequest> />
+              <EnrollmentEndDateField<WithdrawRequest> />
+              {/* does this need to still be here? */}
```
jessethach, line 122: Ever find out if this needs to be here?

client/src/containers/Roster/Roster.tsx
```diff
@@ -374,17 +371,31 @@ const Roster: React.FC = () => {
                     setAlerts={setAlerts}
                     fetchRoster={fetchRoster}
                   />
-                  <TableBatchAction
+                  <BatchWithdraw
+                    children={(() => {
+                      let c: ChildEnrollmentIdToNameMap = {};
```
jessethach, line 376: Might be a good use case for `.reduce`:

```
return selectedRows.reduce<ChildEnrollmentIdToNameMap>((acc, { id: rowId }) => {
  const child = childRows.find(
    (child) => child.id === rowId
  );
  if (child?.enrollmentId) {
    acc[child.enrollmentId] = child?.name;
  }
  return acc;
}, {});
```

src/routes/enrollments.ts
```diff
@@ -62,12 +62,24 @@ enrollmentsRouter.post(
...
   passAsyncError(async (req, res) => {
-    const enrollmentId = parseInt(req.params['enrollmentId']);
+    const query = req.query['enrollmentIds'] as string[];
+    const enrollmentIds = query.map(Number);
+    if (enrollmentIds === null || enrollmentIds.length === 0) {
+      throw new HttpErrors.BadRequest(
+        'null or empty enrollmentIds array provided.'
+      );
+    }
+    console.log('enrollmentIds', enrollmentIds);
```
jessethach, line 80: Leftover log?

jessethach (approved):
Just optional nits. Great work!

client/src/containers/Roster/BatchWithdraw.tsx
lukenewman: haha -- nope @JordanGuinn do we need to keep that field around?

client/src/containers/Roster/BatchDelete.tsx
lukenewman: like what others?

jessethach: Like line 75 and 76 above

lukenewman: ah. i mean, it's basically two ways of doing the same thing. they both use templates...

bamader (approved):

bamader:
## Validation Plan

- [ ] Can batch withdraw single child
- [ ] Can successfully batch withdraw multiple children with no failures
- [ ] Can batch withdraw children and have some of them fail

bamader:
Validations come back green on this PR, but I did find a pair of other things that could be worth attention as bugs.

Bug 1: for some reason, using the batch actions does not seem to clear alerts from the roster page, so they wind up accumulating at the top (even with separate/different actions like uploads, successful withdraws, failed withdraws, etc.). See image:
Screen Shot 2021-11-17 at 2 40 13 PM
 
Bug 2: After withdrawing some children, the counts displayed on the roster don't update to reflect withdrawn numbers. You can see that after withdrawals, the number of enrollments go down (because children with active enrollments decreased), but we still report this as the total number of children and give age-group breakdowns that don't account for withdrawn records.
Screen Shot 2021-11-17 at 2 40 28 PM

src/routes/enrollments.ts
JordanGuinn: Reminder to pull this out

JordanGuinn (approved):

JordanGuinn:
That first bug seems like it's own thing, and should be addressed on it's own.  Wondering about the second bug though - does this happen with any other batch actions?  If so, then I'd say it should _also_ be it's own bug.

@bamader @lukenewman

bamader:
@JordanGuinn I'll give bug 2 a test with batch delete to see if it correctly handles counts (which I'd imagine it does). I think we've got a batch change enrollment too so I'll test that. But yeah first bug is probs its own thing.

bamader:
Looks like batch delete still counts the kids you just deleted until the app is refreshed. Batch change age group seems unaffected by this (outside of problems displaying historical enrollments, but that's a separate PR).

lukenewman:
@bamader was there a conclusion reached as to whether this is good to go in? i'm not seeing batch delete still counting the kids you just deleted. also not seeing that on this batch withdraw work either. however, the alerts are persisting pretty far. i can get rid of the `...current` to most likely fix that

bamader:
@lukenewman Hmm odd that you can't repro the bug. Let me give it another try. The only thing I was seeing with batch delete was that you had to refresh the page to get the app to update the counts. The alerts being persistent I think we decided should be a separate bug ticket rather than a blocker to this, so I'd say either you or Jordan can write a ticket for it and then once we clear up the counts thing we'll be ready.

bamader:
Screen Shot 2021-11-19 at 1 23 47 PM

Okay, so it looks like the counting issue is only happening in the header section. The number in the roster section is correct, but the numbers up top (13 children across 8 sites) is wrong, as are the counts of enrollments in each age group (because it doesn't include those who were withdrawn).

I think this was also a problem in some stuff Nick was working on, and I believe he said it was happening because the endpoint was retrieving too many rows and counting all of them rather than filtering out the withdrawn children. @nickbristow Have you already addressed this in your PR, or is it something Lucas needs to work on here?

lukenewman:
@bamader that child count bug should be fixed

GordonFarrell:
Child count fix is working properly, batch withdrawal happy path also working properly.

I am seeing a bit of a bug though. If a batch withdrawal is attempted but fails due to not having selected a reason or dates, even after adding the missing data it will still give an "Exit date and exit reason are required" bad request failure unless you completely close the modal then reopen and try again.

lukenewman:
@GordonFarrell can you expand on those repro steps? couldn't get that bug to repro

GordonFarrell:
Sure thing!

1. Select several students and open withdrawal modal

2. Leave reason empty, fill two valid dates

3. Click withdraw and get an error

4.Add a reason and click withdraw again

For me the error is persisting until I close the modal entirely and try again from the start with all valid info for each field.

bamader:
If I'm not mistaken that bug exists in the normal withdraw functionality, not introduced by this PR's usage of the withdraw modal. We can definitely fix it here (because it makes sense to finalize withdraw functionality), but if we want to get this shipped sooner rather than later we can open a separate ticket for that.

bamader:
Confirmed, the exit reason bug exists independent of this PR. @JordanGuinn What should be the call here? Fix it up as part of this PR, or create its own bug ticket (in the interest of giving us closed tickets to report this week)?

JordanGuinn:
@bamader If it's not a super quick fix, let's create a separate ticket and track it independently

bamader:
Got it. I don't think it will be a quick fix (it's likely a 2 or 4 pointer, just guessing) so I'm in favor of separate tracking. Once you create that ticket, this PR will be good to go!

JordanGuinn:
Ticket created - https://app.zenhub.com/workspaces/ece-dev-board-5ff506028d35f30012e0e937/issues/ctoec/data-collection/1480

Issue #1335, opened 2021-05-19
Rebuild roster export functionality


JordanGuinn:
## Background
Now that we've begun implementing paginating and some filters within the new roster view that we're building, we should take a moment and revisit the `Export` functionality that we had previously.  With this new roster UX, users should now have the ability to export their roster data based on _the filters they have specified_, not necessarily just their entire dataset.

**TL;DR** An exported Excel/CSV file should include **all children that exist within the current set of filters**, **independent of page size/count**.

Additionally, now is probably an ideal time to revisit the export logic collectively, as pieces of it could certainly benefit from some refactoring.

## Acceptance Criteria
- User can export their filtered roster data through an `Export` button on the roster page
- Enrollment data in exported CSV/spreadsheet is based on **filters applied to roster view**
- Enrollment data in exported CSV/spreadsheet is **not** limited by page count (i.e. it includes all pages)

## Technical Specs
Our previous export implementation is pretty convoluted, and could certainly benefit from a bit of a fundamental rework.  Some things to consider while working on this ticket:
   - Modularizing the logic responsible for **transforming child data into an exportable format**
   - If possible, the resulting CSV/spreadsheet should streamed directly to the HTTP response
      - Instead of first written to the EC2 filesystem, as it was previously

## Designs
[Figma View](https://www.figma.com/file/GcYwZ05CjnG6QZNYlsCS1X/ECE-Reporter-%E2%80%93-Final-Designs?node-id=86%3A348)
![Roster - Filter View.PNG](https://images.zenhubusercontent.com/602d78337403e44d7db5c4f3/d7199e20-067d-450a-95c9-1113c914a304)

PR #1566, opened 2021-11-16
Export button should include filtered roster children

jessethach:
- Adjust api to endpoint to take in query params
- Adjust CSVExcelDownloadButton component to include path with query params and hideIcon boolean
- Include export button on UI

## Background
Users should now have the ability to export their roster data based on the filters they have specified, not necessarily just their entire dataset

## GitHub Issue
#1335 

## Validation Plan

**Sing in as a provider level user and upload roster data from csv file**

- [ ] Navigate to roster page. Choose multiple filters and click `Export roster` button. Confirm that the filtered list in table roster is reflected in downloaded csv file.
- [ ] Select a table column to sort order. Confirm that order is persisted in the csv file.

## Automated Testing


lukenewman (approved):

nickbristow (approved):

bamader (approved):

bamader:
## Validation Plan
Sing in as a provider level user and upload roster data from csv file. Then:
- [ ] Navigate to roster page. Choose multiple filters and click `Export roster` button. Confirm that the filtered list in table roster is reflected in downloaded csv file.
- [ ] Select a table column to sort order. Confirm that order is persisted in the csv file.
- [ ] (since we're touching the child get endpoint) Edit a child's information in the roster and ensure the changes are reflected on the roster page
- [ ] (since we're touching pagination) ensure that roster pagination works with and without filters

bamader:
The validations I defined all succeeded, but I found two bugs while testing, one of which is confirmed introduced by this PR. 

Bug 1 (not introduced by this PR): text in our roster table for sure does not wrap to the next line. I know we had a discussion about the list of user role types wrapping, but I think we may want to more generally institute wrapping because for long child names, this is also a problem. See here: 
Screen Shot 2021-11-17 at 11 40 31 AM

Bug 2 (introduced in this PR; tested on staging and QA and this bug is not present): applying multiple filters to the roster results in an empty table. The number of children reported in the header text is correct, and exporting the roster still outputs the correct children, but the table itself shows no enrolled children. See image:
Screen Shot 2021-11-17 at 11 41 32 AM

jessethach:
Thanks @bamader - I am attempting to repro the issue. Results shown in the video below:

https://user-images.githubusercontent.com/10701395/142275483-61b867f8-2547-4e33-8cd7-91379c5d3f5c.mov

I was able to successfully add multiple filters. Maybe there is a step here that needs to added in order to repro?

bamader:
Hmm. Let me see if I can give more precise steps.

bamader:
Oh wow I figured out what it was. It's _waaaaaay_ more nuanced of a bug (potentially worth some larger investigation). 

Okay, so you're correct that just using the filters is totally fine. The issue comes in with pagination. I think what's happening is that if you apply filtering on anything past the first page of the roster, you wind up with rendering errors. Specifically, the pagination tool will tell you you're on page 1 of your query, but it'll _show_ the page that you left before filtering. Here's what I did to create the bug, which should also give a clearer example of what I mean:

1. upload this sheet to the roster 
[ECE Data Collection Template (v3).xlsx](https://github.com/ctoec/data-collection/files/7557866/ECE.Data.Collection.Template.v3.xlsx)
 
2. upload this other sheet to the roster (this is to create a large enough roster where the problem manifests):
[Field_Support_test.xlsx](https://github.com/ctoec/data-collection/files/7557869/Field_Support_test.xlsx)

3. with no filters applied, go to page 3 of the roster
4. apply filters for Age Group - Infant/Toddler and for Funding Space - CDC
5. observe what happens. You'll see the 2 enrollments text with the pagination bar saying you're on page 1 of 1, but you won't see any enrollments shown in the roster. You'll also be unable to navigate using pagination. 

So the issue may be some combination of filtering only the information shown on the current page or some carbon component weirdness.

bamader:
Whoops @jessethach ^^^

jessethach:
@bamader wow great catch. Can you confirm that this behavior isn't repro'd in QA/Staging?

jessethach:
> @bamader wow great catch. Can you confirm that this behavior isn't repro'd in QA/Staging?

@bamader Looks like this issue can be repro'd in QA:

Screen Shot 2021-11-17 at 1 40 18 PM

We should probably write a separate bug for this issue. Again - good catch!

bamader:
@jessethach Thanks for validating that across environments! Jordan, given what Jesse has found here, once we get bug tickets up for the two things I found, this PR is officially good to go.

JordanGuinn (approved):

Issue #1550, opened 2021-11-06
Add role columns/tables to ECE Reporter DB


JordanGuinn:
## Background
As part of the School Readiness Liaison initiative, admins are going to need the ability to:
   - View all roles of all users within the application
   - Specify distinct roles for different users, some which may have multiple
   - Specify organization/site permissions **per role** for users that have multiple roles

With this in mind, we're going to need to expand our role permissioning in ECE Reporter itself in order to make that type of RBAC more easily configurable (given our minimal usage with IdentityServer, currently).  We can start laying the foundation for this by adding a `Role` table to the DB, and then creating the corresponding relationships between entities that need it.

## Acceptance Criteria
- A new `Role` table is created, defining all possible user roles, which are currently:
   - `Admin`
   - `Provider`
   - `School Readiness Liaision`
- A `role` column is added to the `SitePermission` and `OrganizationPermission` tables
- A join table is created outlining the Many-to-Many relationship between users and roles
- All of these migrations are **not handcrafted**, but rather **auto-generated** from TypeORM entity changes

## Technical Specs
First, we'll want to create a new `Role` table with the following:
   - `id` (primary key)
   - `role` (string, based on possible enum values outlined above)

Then, we'll want to add a `roleId` foreign key column to both the `SitePermission` and `OrganizationPermission` tables, in order to understand the context of each permission set.  This can be done by adding the following to each of the aforementioned entities:

```
@ManyToOne(() => Role)
role: Role;
```

And then finally, we'll want a junction table between users and roles (as that is a many-to-many relationship), which we can also have TypeORM handle for us with something similar being added to the `User` entity:

```
@ManyToMany(() => Role)
@JoinTable()
role: Role;
```

PR #1574, opened 2021-11-29
Add role columns tables 1550

gordonfarrell:
## Background
Setup new user role table and create relationships between roles and permissions. My first time using typeorm so please take a close look and make sure I don't bork anything with these migrations. Thanks!

## GitHub Issue
[#1550](https://github.com/ctoec/data-collection/issues/1550)

## Validation Plan
Migrations seem to run properly for me, so confirmation from others that the DB is ending up in the state we want it should do it.

## Tech Specs
- [X] If any TypeORM entities were created/updated/removed, the necessary migration(s) were [correctly auto-generated](https://github.com/typeorm/typeorm/blob/master/docs/using-cli.md#generate-a-migration-from-existing-table-schema) and **not manually written**.

## Automated Testing
- [X] All unit tests are passing.
- [N/A] Unit tests have been added to cover this changeset.


client/src/shared/models/RoleEnum.ts
```diff
@@ -0,0 +1,5 @@
+export enum RoleEnum {
```
lukenewman, line 1: can we call this just `Role` instead, or `RoleType`?

lukenewman (commented):
I'm a bit confused on the technical specs of this ticket. Can't we just have a `RoleType` (currently `RoleEnum`) relationship on the `User`? I don't understand why the `Role` table is required with an `id` and all...

client/src/shared/models/RoleEnum.ts
gordonfarrell: Sure, renaming sounds good, though I'll change it to roleType. I tried using just role as the name and my IDE was unhappy with it, I think just `role` may be a reserved word

lukenewman (approved):

jessethach (approved):

jessethach:
I can validate this PR.

jessethach:
I have validated using the steps below:

**Log in with admin/provider level user**

- [ ] Successfully log into app
- [ ] Check db for new `user_role_role` table. Expected to be empty.
- [ ] Check db permissions tables to contain `roleid` column.

bamader:
@jessethach looks like you have checks for all your tests. Does this PR pass validation in your opinion? (this is just something we do for history/documentation purposes--when satisfied, the validator leaves a comment that the PR is good to merge and tags either Nick or me to let us know we can merge to base).

jessethach:
@bamader Good to know! I will do that moving forward. This is ready to merge!

PR #1578, opened 2021-12-02
Reintroduce user creation/invitation scripts

JordanGuinn:
## Background
This PR reintroduces the user creation and user invitation scripts Julia initially wrote a year ago.  Hopefully we never need them again, but who knows.

## GitHub Issue
N/A

## Validation Plan
N/A

## Tech Specs

## Automated Testing


gordonfarrell (approved):

gordonfarrell:
I think this build is failing due to our TS version being below 4.X

https://github.com/puppeteer/puppeteer/issues/6982

Will try bumping TS version this afternoon and see if it works, if so I'll open a separate PR for that version change

bamader (approved):
Ugh I remember these scripts. I worked on something for like a month that ended up being dropped, and this whole thing was a _massive_ pain. I think it's probably worth preserving this code for archival reasons, maybe in a different directory than `src/data/prod` (which implies its current use), but I sincerely hope that we a) never ever need to run this, and b) never ever have to write anything like it.

jessethach:
@bamader Mind giving us a brief overview of the pain points you experienced with these scripts? May be helpful to document.

bamader:
Sure! The biggest experience I had with these scripts was that working in excel, or with anything to do with spreadsheets, is _incredibly brittle_. Like, changing _anything_ with these scripts made them absolutely implode. At the time, our SSH tunnels and DB parameter interaction was really finicky as well, so I'm sure that didn't help. But changing directories/filepaths in the settings, or renaming variables to something excel didn't expect, or even working with the promises that excel sent back but didn't always resolve was just an absolute nightmare. I don't remember the task I was assigned before it was dropped, but I think it had something to do with creating like a reverse-engineered spreadsheet: from the DB (or maybe an input spreadsheet), transform information into some new combination of rows and columns so that the info was easier to visually see. But working with those objects is just really difficult and prone to failure.

jessethach:
@nickbristow Reminder - we plan to close this PR and reference it in the documentation.

PR #1593, opened 2021-12-16
Address crit vulnerability identified by dependabot

gordonfarrell:
## Background
An upgrade to react-dev-utils specifically targeting our critical vulnerability in the immer sub-package. This may also knock out a couple lower severity ones.

Many of the other highs and several lower severity issues are related to the react-scripts package and attempting to update this to 5.0.0 for a fix breaks our previously automatic polyfills as well as a few more minor bugs. I had been planning to use a library called CRACO (Create React App Config Override) to gain access to webpack configs without having to eject our CRA app. In the end I've decided to wait on doing this as it seems like the folks working on CRA are actively working on a solution within CRA itself [in this issue](https://github.com/facebook/create-react-app/pull/11764). I'll check in on this over the next week or so and see if a solution materializes before getting hacky.

Wanted to at least get in the fix for the crit one while the react-scripts upgrade is cooking. 🥘

## GitHub Issue
[Dependabot link](https://github.com/ctoec/data-collection/security/dependabot/client/yarn.lock/immer/open)

## Validation Plan
App has been spun up and tested for normal functionality, would be good to have someone else do the same

## Automated Testing


jessethach (approved):
Looks good!

nickbristow (approved):

nickbristow:
Validation plan:
Smoke test on QA

Good to go

Issue #1569, opened 2021-11-22
Admin users can specify negative capacity values


JordanGuinn:
## Background
Currently, OEC admin users are able to create/update sites with negative capacity values, which doesn't make sense and could lead to potential confusion.  We should ensure that negative values cannot be provided for any site's capacity.

## Repro Steps
- Log in to ECE Reporter as an admin user
- Navigate to the Edit Organization view for a given organization
- Specify a negative value for the capacity of a given site
- Observe

## Expected Behavior
- Application shows a validation error, preventing the user from specifying a negative value for site capacity

## Actual Behavior
- User is able to save a negative capacity value without issue

PR #1596, opened 2021-12-17
Fix admin users can specify negative capacity values 

jessethach:
## Background
- Add minimum value to disallow negative numbers in input.

## GitHub Issue
#1569

## Associated PRs
None

## Validation Plan

**Login as Admin**
- [ ] Navigate to Organization page and enter negative value. Confirm that the field does not allow negative values and will throw up error if submitted.

## Tech Specs
N/A

## Automated Testing

## Additional Context
N/A

  https://user-images.githubusercontent.com/10701395/146615627-a7e512fe-e7cf-447b-a0dc-911a8b8d87f8.mov


nickbristow (approved):

lukenewman (approved):
nice

bamader (approved):
best kind of PR there is

bamader:
## Validation Plan
Log into app as admin, navigate to edit org page for an organization.

- [ ] can edit and save positive/zero capacities without issue
- [ ] cannot save with a negative number entered in a capacity field
   - This test is _technically_ passing, but not in a way that I think matches expected or desirable behavior. If a user types in a negative number, a red stop sign warning label is displayed in the field. However, once a user clicks out of the field (either somewhere else on the page or another button), the value is changed to a 0 and the warning disappears. This means that the field is now valid again and so can be saved. If a user is moving quickly through the page, I think this change is subtle enough so as to not be detected (happened to me first time, actually). A better approach would be to highlight the field red (or leave the warning sign in there if a negative number is entered), but then deliberately _don't_ change the value to 0 when a user clicks away. If a save attempt is made while one or more fields is negative, we block the request from sending and instead show a banner `alert` at the top of the page indicating that one or more sites has a negative capacity, which is prohibited.

jessethach:
@bamader I will work on displaying the validation message you mentioned. I may need to strip that min prop if that's the case because I think there is some default Carbon behavior with min/max values. I'll keep you posted.

jessethach:
@bamader Changed the logic so that the `-` character is restricted on key press. Let me know if you agree with this pattern. I did not personally find it too invasive as a user. Others may feel different!

bamader:
@jessethach Yeah I don't think I find this too intrusive. My only suggestion would be to change the copy that comes up with a blank capacity field to "Non-negative capacity number is required" instead of "Valid capacity...". That way we emphasize/explain why the '-' key is disabled.

jessethach:
@bamader Copy is changed! Thanks for looking it over.

gordonfarrell:
Ran this through validation, I like the total inability to even put in a - . Nice! That's working as intended, good fix for this.

Only small issue I'm seeing is that on clicking to add a new funding space when the space pops up its already showing a validation error requiring the capacity before a user even touches the field. Assuming we want to default to no error until submission?
Screen Shot 2022-01-04 at 2 46 42 PM

Other than that looks great! 👍

jessethach:
@gordonfarrell Good catch! I will look into possibly defaulting the value to `0` or allowing it to be empty, at least until submission.

gordonfarrell (approved):

Issue #1480, opened 2021-09-14
User can't withdraw child if they initially miss a field in the modal


JordanGuinn:
## Background
It looks like if a user is missing fields in the withdraw modal and we prevent submission (resulting in error UX in the withdrawal modal), said user can’t re-submit until you close/reopen the modal.

## Repro Steps
- Log in to ECE Reporter
- Navigate to a child detail view, for a child that has all data filled in
- Click `Withdraw`
- Omit one or more fields
- Submit
- Observe failure UX within the withdraw modal
- Attempt to address errors by filling in all fields
- Submit
- Observe

## Expected Behavior
User is able to fix their issues and withdraw the child without issue.

## Actual Behavior
User receives another withdrawal error, even though they've addressed all fields in the modal.

PR #1597, opened 2021-12-20
Not great but working solution

bamader:
## Background
When attempting to withdraw a child, if a user has an error during any part of the modal, fixing the error doesn't allow a user to submit. Instead, they have to close the modal, refresh the app, and then reenter everything correctly. This is a really annoying user workflow that's blocked some users from inputting their changes, so this changeset fixes that (albeit in a not ideal way--but see context below).

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1480

## Associated PRs
N/A

## Validation Plan
Because this change touches several places in the app, validation has to hit the following areas:
- withdraw child modal
- change enrollment modal
- editing an enrollment via edit record
- editing a funding via edit record
- creating a child using the Add Record flow

## Tech Specs
N/A

## Automated Testing
N/A

## Additional Context
So this PR is...non-ideal. The code definitely works in that the withdraw bug is solved, as are a couple other (potentially related? maybe not?) bugs that were annoyances in the app. But the way I solved things was just using very basic state in the same way that we do in several other parts of the app (notably, the admin view). I think it's not great to combine state manipulation with form fields (others can disagree though). I think the real issue is ultimately the way that form fields were originally designed and how we currently access them. Drillers make this almost _impossible_ to debug or correct. The console log statements are the proof I left in of that. To see what I mean, run the app with my fix in place and observe the logs; you'll notice that the appropriate fields are created and maintained throughout the lifecycle of the modal being open. However, to see prior behavior, comment out the part of the `ExitReasonField` that leverages the new state code (i.e. it'll turn back into its old form by removing the optional parameter). Now check out the logs. You'll see the exit reason field **completely disappears after clicking the submit button but before data ever reaches the back end**. I have absolutely no explanation for this, and after two days of trying to track it down, I've got no plausible reason why a click should do that if no data is transferred and when the same thing doesn't happen to the date field. So this PR is in essence a timeboxed solution to that problem, which raises the larger issue: _we need to spend several months just freaking cleaning up this app code_. Starting with the component library and form fields. 

TLDR: Putting up a non-ideal solution that fixes this bug and several others, to:
1. get other eyes on the fix
2. to illustrate a problem with current form fields that I can't easily resolve in a way that respect ticket points
3. to advocate for a serious refactor of our form and component library pieces


client/src/components/Forms/Enrollment/Fields/EnrollmentEndDate.tsx
```diff
@@ -1,49 +1,67 @@
...
+  hideStatus = true,
+}: EnrollmentEndDateProps) => {
+
+  const {
+    data,
+    dataDriller,
+    updateData,
+  } = useGenericContext<T>(FormContext);
+
+  console.log(data);
```
jessethach, line 36: Console log here.

client/src/components/Forms/Enrollment/Fields/ExitReason.tsx
```diff
@@ -1,38 +1,86 @@
...
       id="exit-reason"
-      label={<span className="text-bold">Exit reason</span>}
-      options={Object.values(ExitReason).map((reason) => ({
-        text: reason,
-        value: reason,
-      }))}
+      labelText={<span className="text-bold">Exit reason</span>}
       onChange={(e) => {
-        immutableUpdateData(dataDriller.at('exitReason'), e.target.value);
+        // const updatedData = produce<T>(
```
jessethach, line 50: Need commented out code here anymore?

client/src/containers/EditRecord/WithdrawRecord/WithdrawRecord.tsx
```diff
@@ -45,8 +45,28 @@ export const WithdrawRecord: React.FC<WithdrawProps> = ({
   const history = useHistory();
 
   const { query, updateChildRecords } = useContext(RosterContext);
+  const [showErrors, setShowErrors] = useState(false);
+  const [withdrawRequest, setWithdrawRequest] = useState({} as WithdrawRequest);
 
   const onSubmit = (withdraw: WithdrawRequest) => {
+    console.log(withdraw);
```
jessethach, line 52: Console log here.

jessethach (approved):
Just nits!

bamader:
The console logs were left in deliberately in case anyone wanted to compare the log during runs with and without my fix, because there's that forsaken issue where the `exitField` data is vanishing after clicking `Submit` but before the API endpoint is ever reached, which makes no sense. However, if we're satisfied with this as a fix for now I can just go ahead and delete them / clean up. I'll see what other reviewers say / want to do and see if we can get a consensus.

gordonfarrell (approved):
This all looks good to me, and nice job with the extra little bugs picked up along the way. Definitely a tough situation to solve elegantly with the driller and all so I'd definitely say it working properly > giant overhaul to make it perfect. Just the logs and the couple lgtm things and then I'm happy to give it a validation whirl. 🌪

gordonfarrell:
Validated this, works great! Just those lgtm warnings and then its off to the races 🏁  🏇🏾🏇🏾🏇🏾🏇🏾

jessethach:
@bamader Looks like this is ready for merge.

bamader:
Yup yup! Just gotta peel out some console logs and delete whatever those new LGTM warnings are and I'll merge in. I'm closing in on the upload bug so gonna try to open a PR for that first.

PR #1598, opened 2021-12-21
Add user roles column to admin user view

nickbristow:
## Background
Adds user roles column to admin user table
## GitHub Issue
Replace this text with the GitHub issue link.
**NOTE: Please don't mark the issue as closed by this pull request!  Issues are to be closed manually within Zenhub upon merge & prod deploy.**

## Associated PRs

## Validation Plan

## Tech Specs

## Automated Testing

## Additional Context


client/src/containers/Users/Users.tsx
```diff
@@ -16,11 +16,22 @@ const Users: React.FC = () => {
 
   const [data, setData] = useState<UserSummary[]>([]);
 
+  const formatRes = (res: UserSummary[]) => {
+    return res.map((u) => {
+      if (u.roles)
```
jessethach, line 21: We have a pattern of defaulting an empty array for any falsey values. Maybe we should add one here:
```
return (res || []).map(u => {...})
```

client/src/shared/payloads/UsersResponse.ts
```diff
@@ -3,4 +3,5 @@ export interface UserSummary {
   name: string;
   email?: number;
   organizations?: string;
+  roles?: string;
```
jessethach, line 6: This typing seems to be unhelpful here. Is it necessary that the db return a comma string? What if it returned an array? This way we can type the roles stronger. Like:

```ts
export interface UserSummary {
  id: number;
  name: string;
  email?: number;
  organizations?: string;
  roles?: RoleType[];
}
```

Example of what the RoleType array would look like:
```ts
const roles: RoleType[] = [RoleType.Provider, RoleType.SchoolReadinessLiaison];
```

bamader:
@nickbristow Given what we know about roles (specifically school readiness liaisons) changing (and I think not being present more in the app), is this PR still something we need?

bamader:
@nickbristow is this PR ready for review again? If so, can you re-mark Jesse as a reviewer?

jessethach:
@bamader @nickbristow I recall having a question about this https://github.com/ctoec/data-collection/pull/1598/files#r773469813

Not sure what decision was made in terms of typing role type.

client/src/shared/payloads/UsersResponse.ts
nickbristow: @jessethach it's always going to return a comma string, its really just a matter of parsing it on the backend or front end.

nickbristow:
@jessethach @bamader this is ready for review again

client/src/shared/payloads/UsersResponse.ts
jessethach: So the this is a specific type of the db? It looks like it's falling under payloads dir, which I assume is an API response type. Will there be a type when it's parsed at some point?

jessethach (approved):
It looks good and I trust your judgement on the RoleType question. I just like having hard types for props that have only have a limited amount of available values.

lukenewman (approved):

PR #1620, opened 2022-01-21
Add "Record Status" filter options to roster redesign

lukenewman:
## Background
Add in "Complete" and "Incomplete" options to the "Record Status" filter. "Complete" will show only records with no validation errors, while "Incomplete" will show only records _with_ validation errors.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1470

## Associated PRs

## Validation Plan
- [ ] Have a roster with children in it
- [ ] Filter by "Complete" -- you should see only children with no validation errors
- [ ] Filter by "Incomplete" -- you should see only children _with_ validation errors. If you have none, you can edit a record's enrollment to get some validation errors.

## Tech Specs

## Automated Testing

## Additional Context


lukenewman:
One product question I have regarding this and other filters, should we be considering these to be "subtractive" filters or "additive" filters? For example, if you go select all the funding spaces, no children will show up because no record is a member of every funding space. I consider this the "subtractive" filter model. In an "additive" model, this situation would show all children because any record with a funding space will be included in the selected options.

Regarding this particular filter, it almost seems like a toggle may be an option, i.e. "Show Incomplete Records Only". If you go select both "Complete" and "Incomplete" nothing will show up, which I feel like is an unnecessary user path. Happy to talk about this "irl" cause it's not ideal to describe in text.

jessethach (approved):

gordonfarrell (approved):
Yay filters! Looks good

jessethach:
I can validate this.

src/controllers/children/get.ts
```diff
@@ -224,6 +224,10 @@ export const getChildren = async (
   const [children, totalItems] = await qb.getManyAndCount();
   let validatedChildren = await Promise.all(children.map(validateChild));
   let { ...filterSet } = query ?? {};
+  /// we can't filter by validation status in `getChildrenQuery`
+  /// so we keep track of a separate total count to return
+  /// the appropriate total pages and count
+  var totalChildren = totalItems;
```
jessethach, line 230: This should be a `let` probably.

lukenewman: fixed (my swift is showing)

lukenewman:
@jessethach this is ready for another validation when you have a moment

jessethach:
This is validated with most recent changes! @nickbristow Good to merge.

Issue #1262, opened 2021-03-31
Write enrollment uploads directly to S3, not to the EC2 file system


JordanGuinn:
## Background
In order to better triage potential upload failures and the underlying reasons, we should start writing all enrollment file uploads directly to S3, as opposed to the EC2 instance directly.  Given that our EC2s in prod-like environments are not in any way accessible outside of through HTTPS from the load balancer, gaining access to those files for debugging purposes in a pretty big pain.  This is also the exact kinda thing that S3 is _made for_.

## Acceptance Criteria
- File uploads are written to a new `uploads` directory in the already existing `ece-fawkes-<stage>-store` S3 bucket on upload
   - If possible, contents of this directory have a 2 week expiration set on them
- Corresponding file uploads are removed after the enrollment data from said file has been successfully added to the roster

kmariepat-skylight:
@jessethach I accidentally moved this card, it may be in the incorrect status, please confirm and adjust as needed

PR #1626, opened 2022-01-28
Add upload to S3 bucket logic

jessethach:
# Currently in Draft Mode
**I need to sort out role keys and environment variables from Circle CI before this can be official.**

---

- Added aws client-s3 module package.
- Added env variablest o configure aws sdk methods.
- Added middleware logic to send files to S3 bucket folder.

## Background
In order to better triage potential upload failures and the underlying reasons, we should start writing all enrollment file uploads directly to S3, as opposed to the EC2 instance directly. Given that our EC2s in prod-like environments are not in any way accessible outside of through HTTPS from the load balancer, gaining access to those files for debugging purposes in a pretty big pain. This is also the exact kinda thing that S3 is made for.

## GitHub Issue
#1262

## Associated PRs

## Validation Plan

**Log in as provider user**
- [ ] Upload csv file and confirm success of upload end to end.
- [ ] Upload escel file and confirm success of upload end to end.
- [ ] Confirm that there are no network failures when uploading a file.
- [ ] Log into AWS and search for QA ece environment for storing files. Confirm that uploads directory exists.
- [ ] In AWS, confirm that the files that were uploaded through data collection appears to in the QA store bucket.

## Automated Testing

### You may need to add env variables in `.env` or `.env.local`.

```
AWS_ACCESS_KEY_ID=**********
AWS_SECRET_ACCESS_KEY=********
BUILD_ENV=dev*****
AWS_REGION=us-east-*
```
This will be sorted out further once I obtain IAM role keys from AWS admin. For now, these can be filled with personal AWS tokens. DM me for specific values and how to obtain them.


src/controllers/fileUploadS3.ts
```diff
@@ -0,0 +1,9 @@
+import { S3Client } from '@aws-sdk/client-s3';
+
+export const s3instance = new S3Client({
+  region: `${process.env.AWS_REGION}`,
+  credentials: {
+    accessKeyId: `${process.env.AWS_ACCESS_KEY_ID}`,
```
jessethach, line 6: @jdorothy This leg will need to be up to an aws admin or aws user with ability to create IAM roles for EC2 instances to gain read/write access to specific S3 Buckets. It seems that each EC2 instance that coincides with an environment (dev, qa, etc) will need a role that points to a matching S3 bucket environment.

Found an article that summarizes our similar use case https://cloud-gc.readthedocs.io/en/latest/chapter03_advanced-tutorial/iam-role.html

jdorothy (approved):
LGTM, but let's also make an issue to address the IAM profiles. Feel free to assign it to me.

.circleci/config.yml
```diff
@@ -254,7 +254,7 @@ jobs:
       - aws-cli/setup
       - run:
           name: Set server-side environment variables
-          command: ./scripts/set-env-vars.sh << parameters.env >>
+          command: ./scripts/set-env-vars.sh << parameters.env  >> $AWS_REGION $AWS_ACCESS_KEY_ID $AWS_SECRET_ACCESS_KEY
```
lukenewman, line 257: extra space before the `>>`?

lukenewman (approved):
dope

lukenewman:
one question i have is the potential performance hit from reading the file twice. we're reading it once here with the new functionality, but then again in `parseUploadedTemplate`. i'm wondering if there's a way to parse once, if there is a performance concern.

jessethach:
@lukenewman That's a valid concern. In this case, the file from the tmp directory is being read in the async function before the parsing logic is called. It's not actually doing any parsing during the function I created, just sending the file buffer to S3. 

I did actually try to pass the file from the `Multer` middleware into the function I added. However, the sdk api method did not seem to like the `Multer` file, so I used asynchronous `readFile` instead. That said, it shouldn't be too much of a performance hit on a server. `fs.readFile` is a NodeJS method, whereas the `readFile` method used in the `parseUploadedTemplate` function is imported from `xlsx`.

src/controllers/fileUploadS3.ts
jessethach: @jdorothy Should I change this to use AWS secret manager instead?

PR #1642, opened 2022-02-10
Changed how batch size is calculated for batch save and batch soft delete

nickbristow:
## Background
Large files of more than 400 rows were causing a 413 error. This fixes it by better estimating the parameter size of the entity being batch saved. 

## GitHub Issue
Replace this text with the GitHub issue link.
**NOTE: Please don't mark the issue as closed by this pull request!  Issues are to be closed manually within Zenhub upon merge & prod deploy.**

## Associated PRs

## Validation Plan
- [ ] Upload a file of at least 400 rows

## Tech Specs

## Automated Testing

## Additional Context


src/controllers/enrollmentReports/map/batchUpsert.ts
```diff
@@ -9,7 +9,13 @@ import {
   IncomeDetermination,
 } from '../../../entity';
 
-const maxParametersPerQuery = 2000;
+const maxParametersPerQuery = 1900;
```
lukenewman, line 12: why was this lessened?

lukenewman (approved):

src/controllers/enrollmentReports/map/batchUpsert.ts
lukenewman: oh ok so that's not the "limit" in question, and lessening that will guard us from hitting said other "limit"

nickbristow: Yes, exactly

Issue #1617, opened 2022-01-18
Refactor alerts across app to not use location state


nickbristow:
Alerts are currently displayed inconsistently and this story is to research and implement a new way of handling alerts across the app. 

Note: useAlerts file is where most of the work lies. Move away from location state. 

Acceptance criteria
- Use some sort of global state ie context API or something similar

PR #1656, opened 2022-02-21
Refactor alert management

lukenewman:
## Background
This PR refactors the way alerts were managed in the app. I removed the custom alert management and switched over to using [react-hot-toast](https://react-hot-toast.com/). Rather than sticky alerts with cleared-on-refresh behavior, the new success and failure alerts should dismiss after 5 seconds, while the info and warning alerts should stick around.

## GitHub Issue
https://github.com/ctoec/data-collection/issues/1617

## Associated PRs
N/A

## Validation Plan
The entirety of the app will need to be smoke-tested to ensure alerts are displaying properly.

**There may still be some visual bugs.**

## Tech Specs

## Automated Testing

## Additional Context


client/src/containers/Organizations/Organizations.tsx
```diff
@@ -17,18 +16,18 @@ const Organizations: React.FC = () => {
...
-      alerts.find((a) =>
-        [
-          'Organization not found',
-          'Organization updated',
-          'Organization created',
-        ].includes(a?.heading ?? '')
-      ),
-    ]);
-  }, [setAlerts]);
+  // const [alertElements, setAlerts] = useAlerts();
```
jessethach, line 19: Need this?

client/src/containers/Roster/Roster.tsx
```diff
@@ -172,6 +164,7 @@ const Roster: React.FC = () => {
     currentPage: 0,
     totalItems: 0,
   });
+  const [showPastEnrollments, setShowPastEnrollments] = useState<boolean>(true);
```
jessethach, line 167: Don't need to define the type in the generic argument if setting a value, unless you are defining something more complicated like a union type (e.g. `string | number`).

It would be more helpful when there isn't a value set for whatever reason, like:

```
const [ _ , _ ] = useState<boolean>();
```

jessethach (approved):
Great work!

client/src/containers/Roster/Roster.tsx
lukenewman: gotcha. this is from merging in the epic brarnch

client/src/containers/Organizations/Organizations.tsx
lukenewman: honestly not sure. seems like custom logic from the past implementation that might be obsolete now? need to test it out

client/src/containers/Help/SupportRequest.tsx
```diff
@@ -14,16 +14,21 @@ const SupportRequest: React.FC = () => {
...
+                </h2>
+                <p className="usa-alert__text">
+                  <span>
+                    Don't share any personally identifying child information
+                    over the support form or email.
+                  </span>
+                </p>
+              </div>
+            </div>
+          </div>
```
SMakaiTakori, line 31: Is there a reason you didn't want to use the `createAlert` util here? Just curious.

client/src/utils/createAlert.tsx
```diff
@@ -0,0 +1,40 @@
...
+          )}
+          <p className="usa-alert__text">{alert.text}</p>
+        </div>
+        {alert.actionItem && (
+          <div className="usa-alert__body flex-auto display-flex flex-align-center">
+            {alert.actionItem}
+          </div>
+        )}
+      </div>
+    </div>,
```
SMakaiTakori, line 32: You grabbed this section from the original `<Alert/> component, right? I should be able to just update this section to incorporate the new design once this PR is merged.

SMakaiTakori (approved):
This all looks good, great work!

client/src/utils/createAlert.tsx
lukenewman: yes! we can incorporate the new design here.

client/src/containers/Help/SupportRequest.tsx
lukenewman: yeah this one is sticky/persistent, so i just left it here

client/src/containers/Organizations/Organizations.tsx
lukenewman: org creation/editing success alerts still work fine, and i'm not a huge fan of the manual filtering here, so i'm gonna get rid of this

SMakaiTakori:
Validating currently!

lukenewman:
ok so after reviewing makai's validation notes, i believe i'm going to make the following changes:

* static alerts (info, warning) will be placed inside the component and will not use `createAlert` / `react-hot-toast` at all
* success and error alerts will be shown as-is but i'm gonna change the position of the `Toaster` to top-right so that they don't overlap with the UI too much

